PR #104, opened 2024-05-22
Run Integration Tests in CI

jkmassel:


oguzkocer (approved):
There are still a few things that could be improved, but I think this is a good start as the integration tests can be run both locally and from CI.

We can keep improving it in subsequent PRs.

PR #207, opened 2024-07-26
Site settings

oguzkocer:
Implements `/settings` endpoint. It follows the patterns from existing endpoints with a few minor differences: 

* Its mutable integration tests uses `wp_cli` instead of `wp_db` to validate the updates. The output of `wp_cli` commands are much more similar to the API responses, so it's much easier to work with.
* By using `generate_test` macro, I was able to fully cover all possible single field updates in update site settings tests

---

**To Test**
* `make test-server && make dump-mysql && make backup-wp-content-plugins`
* `cargo test`
* `cd native/kotlin && ./gradlew :api:kotlin:integrationTest`

### DIFF
--- a/native/swift/Example/Example/ExampleApp.swift
+++ b/native/swift/Example/Example/ExampleApp.swift
@@ -37,6 +37,9 @@ struct ExampleApp: App {
             ]
 
             return items.map { $0.asListViewData }
+        }),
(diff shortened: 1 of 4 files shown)


jkmassel (approved):
We'll need to come back to arbitrary values, but we should land this as-is for now

PR #230, opened 2024-07-31
Condense error variants into `WpError`

oguzkocer:
* Condenses `WpRestErrorWrapper::Recognized` & `WpRestErrorWrapper::UnrecognizedWpRestError` into `WpApiError:WpError`
* Moves `WpInternalErrorCode` variants into `WpRestErrorCode` and adds `WpCore` prefix to them
* Renames the `WpRestErrorCode` as `WpErrorCode`
* Adds `WpApiError::InvalidStatusCode`
* Adds `WpErrorCode::CustomError(String)` as a fallback
* Updates Kotlin's `WpRequestResult` to include all `WpApiError` types directly to make it easier to work with
* Adds helper extension functions to `WpRequestResult` in Kotlin integration tests
* Updates all integration tests for the above changes

---

**To Test**
* `make test-server && make dump-mysql && make backup-wp-content-plugins`
* `cargo test`
* `cd native/kotlin && ./gradlew :api:kotlin:integrationTest`

### DIFF
--- a/wp_api/src/api_error.rs
+++ b/wp_api/src/api_error.rs
@@ -24,9 +24,10 @@ pub enum WpApiError {
         status_code: Option<u16>,
         reason: String,
     },
-    #[error("Rest error '{:?}' with Status Code '{}'", rest_error, status_code)]
-    RestError {
-        rest_error: WpRestError,
+    #[error("Rest error '{:?}' with Status Code '{}'", error_code, status_code)]
+    WpError {
+        error_code: WpRestErrorCode,
+        error_message: String,
         status_code: u16,
         response: String,
     },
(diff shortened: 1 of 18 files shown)


wp_api/src/api_error.rs
```diff
@@ -15,6 +15,8 @@ pub enum RequestExecutionError {
 
 #[derive(Debug, PartialEq, Eq, thiserror::Error, uniffi::Error)]
 pub enum WpApiError {
+    #[error("Status code ({}) is not valid", status_code)]
+    InvalidStatusCode { status_code: u16 },
```
crazytonyli, line 19: Nitpick: This error is for HTTP status code like `1000`, right? What do you think making it more explicit, like `InvalidHTTPStatusCode`?

crazytonyli (approved):

wp_api/src/api_error.rs
oguzkocer: Sure - I'll make that change in a subsequent PR since this got auto-merged. I think it needs to be `InvalidHttpStatusCode` though to follow naming guidelines.

PR #235, opened 2024-08-07
Publishing Kotlin artifacts no longer depends on Detekt

oguzkocer:
There is no reason to prevent publishing Kotlin/Android artifacts from publishing if Detekt fails. So, this should speed up the time it takes for a PR to be ready to be merged.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -144,8 +144,6 @@ steps:
       - label: ":kotlin: Publish `rs.wordpress.api:kotlin`"
         key: "publish-wp-api-kotlin"
         plugins: *common_plugins
-        depends_on:
-          - "kotlin-detekt"
         command: |
           echo "--- :rust: Installing Rust"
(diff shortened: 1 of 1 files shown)


jkmassel:
Is this because `detekt` is just code style, so we're ok if it's not quite right?

oguzkocer:
> Is this because detekt is just code style, so we're ok if it's not quite right?

Right, but more importantly, that sort of thing should only prevent us from merging a PR, not from testing it.

Let's say I open a PR to test its integration in WPAndroid and take a break while waiting for it to publish. I don't want to come back and see that it didn't even publish because there is an unused import. If we publish it, we can still test the artifact and fix the styling issue afterwards.

I think this should be the case for lint as well, even though lint covers more than code style. As long as we can't merge a faulty PR, I think we can publish the artifacts as soon as possible.

jkmassel:
Works for me, thanks for the explanation!

jkmassel (approved):

PR #236, opened 2024-08-07
Remove unnecessary JNI configuration

oguzkocer:
It looks like all of this configuration is already handled by `org.mozilla.rust-android-gradle.rust-android` plugin.

**To Test**

If publishing is successful, that should be enough. However, double checking that the JNI artifacts are included can be done by:

1. Running the `UsersEndpointAndroidTest` which runs on Android device/emulator and will require android JNI artifacts to be there.
2. By updating the dependency version for [WPAndroid integration PR](https://github.com/wordpress-mobile/WordPress-Android/pull/21108) and by testing the integration.

---

I've tested this in https://github.com/wordpress-mobile/WordPress-Android/pull/21108/commits/8899eb39d3800f36bba3aff179e41dfe0cd79344 and validated that it works correctly.


jkmassel (approved):
I'm confident it's working if the build succeeds

PR #238, opened 2024-08-07
Remove native cargo target from `:api:android` Kotlin module

oguzkocer:
We run our native tests from `:api:kotlin` module and use `:api:android` only as a wrapper to package Android artifacts. So, we don't need to build the Rust project for the native architecture as part of `:api:android` Kotlin module.

### DIFF
--- a/native/kotlin/api/android/build.gradle.kts
+++ b/native/kotlin/api/android/build.gradle.kts
@@ -100,7 +100,7 @@ cargo {
     module = "$cargoProjectRoot/$moduleName/"
     libname = moduleName
     profile = "release"
-    targets = listOf("arm", "arm64", "x86", "x86_64", nativeRustTarget())
+    targets = listOf("arm", "arm64", "x86", "x86_64")
     targetDirectory = "$cargoProjectRoot/target"
     exec = { spec: ExecSpec, _: com.nishtahir.Toolchain ->
         // https://doc.rust-lang.org/rustc/command-line-arguments.html#-g-include-debug-information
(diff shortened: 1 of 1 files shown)


jkmassel:
> So, we don't need to build the Rust project for the native architecture as part of :api:android Kotlin module.

This is because `:api:android` wraps `:api:kotlin` which already has it? So for tests and for app integrations we're compiling the Rust code then just throwing it away currently?

oguzkocer:
@jkmassel The platform native Rust binary - such as for Mac OS or Linux - is only necessary if we are running the code on that platform. For example, integration tests from `:api:kotlin` are run directly on the platform they are called from, so we need the Rust binary to be available to it. This is also the case for the multi platform example app, because there is a Desktop version.

In `:api:android` module's case, the module is only supposed to be run from an Android target. It technically shouldn't even load in any other target because the `jna` dependency we use for this module uses the `aar` artifact which similarly is only available for Android targets. (`:api:kotlin` includes the non-Android `jna` dependency)

`:api:android` is a wrapper around `:api:kotlin` module, and its only job is to package the Rust binary for the Android architectures in its `jniLibs` folder.

Note that `:api:kotlin` module doesn't package _any_ Rust binaries. I don't believe there is a built-in way to do that. The loading `jniLibs` folder is something that's provided by Android Gradle Plugin. So, any client that adds `:api:kotlin` as a dependency also has to load the Rust binary. This is something we already do for Desktop version in the multi platform example app.

> This is because `:api:android` wraps `:api:kotlin` which already has it?

I've explained this in more detail above, but I'll reply specifically to make sure. Published artifacts for `:api:kotlin` doesn't include any binaries as there isn't a way to do it - at least as far as I know. We might be able to write our own Gradle plugin that does it in the future, but it wouldn't be particularly helpful for our current projects. So, we should only look into it when we intend to use the project on a non-Android target - such as Pocket Casts backend.

The reason we don't need the Rust native binary for `:api:android` is because we don't use it. If we wanted to run unit tests that loads the Rust binary from this module, we'd need it. That's actually why it's there, because until `:api:kotlin` was separated, we were doing everything from a single module and had to built the native target. It's just something I forgot to remove during the module re-work.

> So for tests and for app integrations we're compiling the Rust code then just throwing it away currently?

I am not sure if I completely understand this question. We run the integration tests in `:api:kotlin` for which we do compile the Rust code and then use it while running the test.

In `:api:android`, we only have a single Android integration test, which runs on Android platform. So, we don't need the native binary for it.

---

Does that answer your questions?

jkmassel (approved):
That answers my question, yep – thanks!!

PR #246, opened 2024-08-08
Kotlin integration tests in Docker & CI

oguzkocer:
This PR adds a `wordpress.Dockerfile` that handles some of the setup we have been doing in our scripts, so it's done once and properly cached. It's especially useful to do this now that we will be running most of our integration tests from Docker, even when we need to run them locally. It installs Rust, Kotlin, Android SDK and Gradle on top of the WordPress image.

It also adds a new Make task, `test-kotlin-integration` so we can easily run the Kotlin integration tests from Docker. A new CI step runs Kotlin integration tests and runs the aforementioned Make task.

**To Test**

* `make test-kotlin-integration` to verify that Kotlin integration tests are working locally

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,7 +2,20 @@
 common_params:
   plugins: &common_plugins
     - automattic/a8c-ci-toolkit#3.4.1
-  # Common environment values to use with the `env` key.
+  matrix: &wordpress_version_matrix
+    # All versions except for the latest have been disabled until we have version specific 
+    # integration tests
+    # - '5.6' # First version to introduce appliation passwords
+    # - '5.7'
+    # - '5.8'
+    # - '5.9'
+    # - '6.0'
+    # - '6.1'
+    # - '6.2'
+    # - '6.3'
+    # - '6.4'
+    # - '6.5'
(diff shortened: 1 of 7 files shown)


jkmassel:
@oguzkocer WDYT about pushing the KT test results over to Buildkite Test Analytics? Separate PR? 

I see all the tests passing, but I guess I'd get more output if one was failing?

The log looks _too_ clean 😂

oguzkocer:
> @oguzkocer WDYT about pushing the KT test results over to Buildkite Test Analytics? Separate PR?

@jkmassel I am not familiar with that. Let's do it in a separate PR at a later time.

> I see all the tests passing, but I guess I'd get more output if one was failing?

That's correct. It's the same for local runs. Having said that, it's been throwing me off as well, so let me see if I can figure out a way to print out the tests being run. I'll open a separate PR if I can figure it out.

---

I'd like to merge this as soon as possible to update the branch protection rules. Since there is a name change, I had to disable one of the requirements for now.

jkmassel (approved):

oguzkocer:
@jkmassel Your concerns have been addressed in https://github.com/Automattic/wordpress-rs/pull/249.

PR #251, opened 2024-08-10
Add CodeQL

jkmassel:
This PR adds `CodeQL` to see if it can surface issues for us. We get it for free, so might as well.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,93 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "trunk" ]
+  pull_request:
+    branches: [ "*" ]
+  schedule:
+    - cron: '33 23 * * 4'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,93 @@
...
+      with:
+        key: "${{ matrix.language }}"
+
+    - name: Prepare Kotlin Project
+      if: matrix.language == 'java-kotlin'
+      shell: bash
+      run: |
+        make setup-rust-android-targets
+        cd native/kotlin
+        ./gradlew cargoBuild --info
```
oguzkocer: What's the intention here? `./gradlew cargoBuild` is basically the same as running `cargo build --release -p wp_api`. (not exactly, but very minor differences) So, it doesn't really build anything Kotlin related.

You probably want something like `./gradlew :api:android assemble` instead, but hard to say without knowing the intention.

jkmassel: You're right about wanting `./gradlew :api:android assemble`, but I'm getting this error:

> Cannot locate tasks that match ':api:android' as task 'android' not found in project ':api'.

Am I holding it wrong?

oguzkocer: Oh, sorry, it looks like I put a space where I should have put a `:` 🤦 It should be `./gradlew :api:android:assemble`. Sorry about that!

Having said that, maybe do `./gradlew :api:android:assembleRelease` instead. We don't have any debug/release specific config, and running `assemble` would do both. So, `assembleRelease` should be faster without any downside.

oguzkocer (approved):

PR #259, opened 2024-08-13
Integration tests backend support

oguzkocer:
This PR adds a tiny Rust server that runs on the `wordpress` Docker container and exposes endpoints such as `/restore?wp_db=true&plugins=true`, `wp-cli/users` etc.

---

**Motivation**

Integration tests that mutate the server state need to be able to restore the server state, so that the other tests don't get affected by the changes. This requires access to the DB and filesystem, which are only available from the Docker container that runs the WordPress server. Furthermore, integration tests need to validate the changes without using the WordPress API - as using the same API implementation may not properly surface issues - for which we need to either run DB queries or `wp_cli` commands.

Up until now, we side stepped the restrictions as much as possible. However, the end result was that we are not able to run the integration tests everywhere. Rust mutable integration tests either only work in the Docker container or our local computer. We could get them to work on both, but it'd be somewhat messy involving environment variables. We don't even have any Kotlin/Swift integration tests that mutate state, because it was simply not possible to restore the state.

By exposing these necessary actions as HTTP endpoints, we remove any external restrictions. That means we can run the tests anywhere. For example, now `make test-server` and then `cargo test` works as expected, whereas previously we had to run the integration tests with `make test-rust-integration` which of course still works.

---

Making this work required a lot of changes across the board, but I think they are all very simple and straightforward. I don't think listing all of them in the PR description is going to be helpful, so I am going to avoid it. However, please feel free to raise any questions and I'll document them in more detail. Here are a few things worth knowing:

* I've removed the unused `dev-server` & `prepare-dev-server` tasks, renamed `test-server` as `start-test-server` and then added a new `test-server` task that calls both `start-test-server` and the new `integration-test-backend` task - which starts the newly added server
* Any changes to the Rocket server requires re-running `make integration-test-backend`, but it's safe to run this without restarting the whole server which speeds up the development cycle.
* `make print-log-integration-test-server` will print the logs of the Rocket server to the terminal. It's a bit rudimentary, but it should be good enough for now.
* `serde_helper` module was moved to its own crate so that we can re-use it across other crates.
* Mutable integration tests no longer require a wrapper function such as `wp_db::run_and_restore(|| {})` instead we can restore the db with `RestoreServer::db()`. I think this may be a little too easy to forget, so I am considering adding a function attribute for it.

_Note that this PR includes a lot of whitespace changes. I recommend taking advantage of the `Hide whitespace` feature on GitHub's diff view to make it easier to spot the actual changes._

**To Test**
* `make test-server`
* `cargo test`
* `make test-rust-integration`


wp_api_integration_tests/tests/test_plugins_mut.rs
```diff
@@ -1,32 +1,29 @@
...
+    };
+    let created_plugin = api_client()
+        .plugins()
+        .create(&params)
+        .await
+        .assert_response();
+    assert_eq!(created_plugin.status, status);
+    println!("Created Plugin: {:?}", created_plugin);
+
+    RestoreServer::all().await;
```
crazytonyli, line 26: Does this line run if any assertion fails? If it doesn't, is there any way to make it always run even if this test function fails, so that its failure won't affect other tests?

oguzkocer: It doesn't at the moment. It's possible to always restore, but I am not sure about the best approach for that within the current implementation. So, I'll leave that to a future improvement.

I _think_ I might be able to address this as part of my work for adding function attributes macros such as `#[mutable_test]`. I intend to handle the restoration within that macro, so it's a good chance to address this as well.

Thank you for bringing it up!

crazytonyli (approved):

wp_cli/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+pub use wp_cli_settings::*;
+pub use wp_cli_users::*;
+
+fn run_wp_cli_command<I, S>(args: I) -> std::process::Output
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    let mut c = Command::new("wp");
+    c.arg("--allow-root")
```
jkmassel, line 15: If you need this, the process is running as the wrong user – I'd strongly suggest modifying this to run as `www-data`

PR #260, opened 2024-08-14
Run integration test server as `www-data` user

oguzkocer:


oguzkocer:
I couldn't get this to work in the limited time I spent on it and I don't think it's particularly important, so I'll close this PR and if necessary open a new one.

Issue #262, opened 2024-08-14
Date type


oguzkocer:
We should use a strongly typed date wherever we can. For example, the `/posts` [endpoint](https://developer.wordpress.org/rest-api/reference/posts/) uses several parameters as such:

> [List Posts](https://developer.wordpress.org/rest-api/reference/posts/#list-posts)
> [Arguments](https://developer.wordpress.org/rest-api/reference/posts/#arguments)
> before | Limit response to posts published before a given ISO8601 compliant date.

PR #264, opened 2024-08-14
List `/posts` endpoint

oguzkocer:
This PR handles the initial setup for the `/posts` endpoint and implements the `List` variant and its integration tests.

For `/posts`, I'd like to try opening smaller work-in-progress type PRs. These won't contain any major issues, but it also won't be as polished as a fully completed endpoint. Completely implementing an endpoint takes time and makes it harder to work on anything else until that work is completed. It also results in a harder to review PR. So, I want to give this approach a shot for bigger endpoints.

There are several improvements to this implementation that I'll tackle in the upcoming PRs:
* `PostListParams::query_pairs` contains _a lot_ of code duplication and the same is true for some other params types. I think there is an easy way to improve this and also make it less error prone. I currently copy/paste one of the field implementations and then update the field name and anything else necessary. However, I am always worried that I'll update the name of the parameter name and not the field name and then end up including the incorrect data. Unit tests address this to a degree, but it's possible to make a similar mistake in unit test implementation, so I want our implementation to help us avoid these issues.
* Add unit tests for `list_posts` endpoint. These are currently missing because I want to use the integration tests to verify that we are making the correct requests before implementing the unit tests. If I do it now and we are not building our query correctly, the unit tests will also need to be updated.
* Go through each integration test and verify that each parameter actually works. The integration tests validate that we are sending the correct parameters, but it can't validate that we are getting back the expected data. That's something we _could_ implement since we control the test server. However, that'd require us to maintain a data set to compare against - which might be worth it specifically for `/posts` as it's arguably the most important endpoint.
* The `date` fields use `String` at the moment and should be replaced with a strong type #262

### DIFF
--- a/wp_api/src/posts.rs
+++ b/wp_api/src/posts.rs
@@ -1,7 +1,28 @@
 use serde::{Deserialize, Serialize};
 use wp_contextual::WpContextual;
 
-use crate::UserId;
+use crate::{UserId, WpApiParamOrder};
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, uniffi::Enum)]
+pub enum WpApiParamPostsOrderBy {
+    Author,
+    #[default]
+    Date,
+    Id,
+    Include,
+    IncludeSlugs,
+    Modified,
+    Parent,
+    Relevance,
+    Slug,
+    Title,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, uniffi::Enum)]
+pub enum WpApiParamPostsTaxRelation {
+    And,
+    Or,
+}
 
 #[derive(Debug, Default, uniffi::Record)]
 pub struct PostListParams {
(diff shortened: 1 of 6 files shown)


wp_api/src/posts.rs
```diff
@@ -0,0 +1,468 @@
...
+    pub tags: Option<Vec<TagId>>,
+}
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct PostGuid {
+    pub rendered: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct PostTitle {
```
jkmassel, line 370: These (ie – all fields related to posts) will probably need to be specialized for the context – the title object has different fields depending on the context.

It _may_ make sense to have this be an enum with each case based on the context, then those cases define the available params? Just a suggestion – feel free to disregard

jkmassel (approved):
Looks like a good start

wp_api/src/posts.rs
oguzkocer: We have `#[WpContext({available_contexts})]` attributes for this, so an `enum` shouldn't be necessary. I'll see what the differences are, but for now integration tests are able to parse it for all 3 contexts 🤔

PR #265, opened 2024-08-15
Add failing test cases for autodiscovery

jkmassel:
Adds some test cases for issues I've seen on real sites:

- Sometimes caching is so aggressive that there's no Link header – we could probably fix that by speculatively loading `/wp-json` and seeing if it works. Worst-case, we'd need an entire HTML parser to read the site and see if it's there.
- IDN URLs can be a bit weird, we should make sure that's covered (it seems to work fine!)
- Sites will use compression in their responses, and the test runner needs to handle it (so I've included all of `reqwest`'s options related to that)

### DIFF
--- a/wp_api_integration_tests/tests/test_login_immut.rs
+++ b/wp_api_integration_tests/tests/test_login_immut.rs
@@ -50,6 +50,10 @@ const VANILLA_WP_SITE_URL: &str = "https://vanilla.wpmt.co/wp-admin/authorize-ap
     "https://jetpack.wpmt.co",
     "https://jetpack.wpmt.co/wp-admin/authorize-application.php"
 )]
+#[case(
+    "https://aggressive-caching.wpmt.co",
+    "https://jetpack.wpmt.co/wp-admin/authorize-application.php"
+)] // Returns gzip responses, may not always include Link header
 #[tokio::test]
 #[parallel]
 async fn test_login_flow(#[case] site_url: &str, #[case] expected_auth_url: &str) {


wp_api_integration_tests/tests/test_login_immut.rs
```diff
@@ -39,6 +39,7 @@ const VANILLA_WP_SITE_URL: &str = "https://vanilla.wpmt.co/wp-admin/authorize-ap
 #[case("vanilla.wpmt.co", VANILLA_WP_SITE_URL)]
 #[case("http://vanilla.wpmt.co", VANILLA_WP_SITE_URL)]
 #[case("https://optional-https.wpmt.co", "https://optional-https.wpmt.co/wp-admin/authorize-application.php")]
+#[case("https://gzip.wpmt.co", "https://gzip.wpmt.co/wp-admin/authorize-application.php")] // Returns a gzip-encoded response that the library should know how to deal with
```
jkmassel: I could see an argument for not having this case (or not having it here) – I left it in there for now for discussion (since we should probably have it _somewhere_)

Ideally though, I don't think I want a dedicated test site just for this – it seems like a one-off that we should test for, but all our test servers should just support it (since I think it's handled invisibly on other platforms)

wp_api_integration_tests/src/lib.rs
```diff
@@ -165,3 +165,17 @@ impl<T: std::fmt::Debug, E: std::error::Error> AssertResponse for Result<T, E> {
...
+}
+
+impl<T: std::fmt::Debug, E: std::error::Error> AssertError for Result<T, E> {
+    type Item = E;
+
+    fn assert_error(self) -> E {
+        assert!(self.is_err(), "Request was successful");
+        self.unwrap_err()
+    }
+}
```
oguzkocer: This is probably not going to be common enough to have a `trait` extension. Also, the `Request was successful` might throw us of in the test logs, because it's not clear that it was expected to fail.

In most cases, we need to validate that the request failed with a certain error for which we already have `AssertWpError` trait. So, at least for now, I think we should individually implement this assertion and pass in a message that'll point out the issue. If there is enough common usage, then we can implement a specialized one like `AssertWpError`.

I realize that we discussed me taking over these PRs and implement them as I see fit, but I just wanted to drop my thoughts here, so if I end up changing it, it doesn't come out of nowhere. 😅

oguzkocer:
> Sometimes caching is so aggressive that there's no Link header – we could probably fix that by speculatively loading `/wp-json` and seeing if it works. Worst-case, we'd need an entire HTML parser to read the site and see if it's there.

We should probably disable caching for these requests with [CACHE_CONTROL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) & [CDN_CACHE_CONTROL](https://www.rfc-editor.org/rfc/rfc9213) headers. What do you think @jkmassel?

jkmassel:
> > Sometimes caching is so aggressive that there's no Link header – we could probably fix that by speculatively loading `/wp-json` and seeing if it works. Worst-case, we'd need an entire HTML parser to read the site and see if it's there.
> 
> We should probably disable caching for these requests with [CACHE_CONTROL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) & [CDN_CACHE_CONTROL](https://www.rfc-editor.org/rfc/rfc9213) headers. What do you think @jkmassel?

This doesn't work :( 

I can't get anything like `curl "https://aggressive-caching.wpmt.co/"  -H 'Cache-Control: private, no-store, max-age=0'` to give the `Link` header – I suspect we'll need an HTML parser

oguzkocer:
We addressed most of the issues from the original PR and added the test cases to `trunk`. So, after discussing with @jkmassel, I've rebased the PR to add the only remaining failing case: `https://aggressive-caching.wpmt.co/`. We plan to address this case in #514.

oguzkocer:
Last remaining case has been addressed in https://github.com/Automattic/wordpress-rs/pull/538.

PR #266, opened 2024-08-16
`AppendUrlQueryPairs` & `QueryPairsExtension` traits

oguzkocer:
This PR implements a more structured way to generate the url query for our `Params` types. This helps avoid duplicating the same logic for these types and makes it less error prone. It's both easier to work with and easier to read.

For the `AppendUrlQueryPairs` implementation, I've opted for the `fn append_query_pairs(&self, query_pairs_mut: &mut QueryPairs) {}` function signature. In general, I'd prefer us to return data as oppose to mutating given data. However, in this case, that meant that we had to clone some of the fields. That's _probably_ OK for the gained ergonomics, but I don't feel comfortable with just _probably_. There is also no practical downside in this case and it sort of works like a unit function because we start with a completely empty query. Another option could be to return a `form_urlencoded::Serializer` type, then turn that into `&str` and pass it into [`url::Url::set_query`](https://docs.rs/url/latest/url/struct.Url.html#method.set_query), but we would still pay an extra cost.

It's also possible I might have missed an approach that can return an `IntoIterator`, but I don't think it's worth spending too much time on it when we have an acceptable implementation.

### DIFF
--- a/wp_api/src/lib.rs
+++ b/wp_api/src/lib.rs
@@ -4,6 +4,7 @@ pub use api_client::{WpApiClient, WpApiRequestBuilder};
 pub use api_error::{RequestExecutionError, WpApiError, WpErrorCode};
 pub use parsed_url::{ParseUrlError, ParsedUrl};
 use plugins::*;
+use url_query::AsQueryValue;
 use users::*;
 pub use uuid::{WpUuid, WpUuidParseError};
 
@@ -19,6 +20,7 @@ pub mod post_types;
 pub mod posts;
 pub mod request;
 pub mod site_settings;
+pub mod url_query;
 pub mod users;
 pub mod wp_site_health_tests;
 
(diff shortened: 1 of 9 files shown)


wp_api/src/lib.rs
```diff
@@ -73,6 +75,8 @@ pub enum WpApiParamOrder {
     Desc,
 }
 
+impl_as_query_value_from_as_str!(WpApiParamOrder);
```
crazytonyli, line 78: Is it possible to implement a `#[derive(AsQueryValue)]` to do the same? This is not a request for changes though.

crazytonyli (approved):

wp_api/src/lib.rs
oguzkocer: It's, but the implementation would be more complicated because if we are going to have a single derive such as `AsQueryValue`, then we'd need to have options for it, such as `from=as_str`, `from=to_string` etc. Handling that is more complicated, so it add maintenance burden. We could add different derives such as `AsQueryValueFromAsStr`, `AsQueryValueFromToString`, but that's not any better unfortunately.

We are trying to implement derive macros when there is significant benefit to them compared to other solutions. This is not such a case, because the simple `macro_rules!` does a good job and it is easier to understand.

Fwiw, for a personal project I'd probably go with the derive macro as I am used to working with them. However, I don't think it's the right choice for this project - at least not at this time.

crazytonyli: Thanks for the explanation!

PR #267, opened 2024-08-19
Make `ApiBaseUrl` support WPCom API endpoint

crazytonyli:
This PR makes minimal changes to `ApiBaseUrl`, to make it be able to return `https://public-api.wordpress.com/wp/v2` URLs (see [this blog](https://wordpress.com/blog/2016/11/11/wordpress-rest-api-on-wordpress-com/) for details about the URL structure comparing to .org REST API endpoints).

Let me know what you think. I'll add more changes to make API client types support WP.com site, if you think this direction is worth perusing.

### DIFF
--- a/wp_api/src/request/endpoint.rs
+++ b/wp_api/src/request/endpoint.rs
@@ -80,13 +80,6 @@ impl ApiBaseUrl {
         site_base_url.try_into()
     }
 
-    fn by_appending(&self, segment: &str) -> Url {
-        self.url
-            .clone()
-            .append(segment)
-            .expect("ApiBaseUrl is already parsed, so this can't result in an error")
-    }
-
     pub fn by_extending_and_splitting_by_forward_slash<I>(&self, segments: I) -> Url
     where
         I: IntoIterator,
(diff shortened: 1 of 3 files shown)


jkmassel:
We're avoiding adding any WordPress.com-specific code to this repo – we can do the same thing by storing the API root for the site (and allowing injection of tokens).

Probably best to mock this up on the Swift side then lower it into this repo?

crazytonyli:
Do you mind explaining what you mean by "storing the API root for the site"? Like using `https://public-api.wordpress.com/wp/v2/sites/<host>/` as `ApiBaseUrl` to create `WpApiClient` instances?

jkmassel:
> Do you mind explaining what you mean by "storing the API root for the site"? Like using https://public-api.wordpress.com/wp/v2/sites/<host>/ as ApiBaseUrl to create WpApiClient instances?

Yeah, exactly – we'd pass `https://public-api.wordpress.com/wp/v2/sites/123456` as the `ApiBaseUrl`, and we'd get that by running the autodiscovery process on the site URL (which would give us that root).

PR #294, opened 2024-09-09
Implement create `/posts` endpoint and its integration tests

oguzkocer:
Implement create `/posts` endpoint and integration tests that validate a post can be created from content, title or excerpt. If all these fields are missing, the server responds with `empty_content` error which is validated through an integration test.

The integration test backend now uses fields to retrieve all relevant fields from `wp_cli` and exposes it from `wp-cli/post` and `wp-cli/posts` endpoints.

Implements some of the missing contextual fields for `SparsePost` and its sub types.

### DIFF
--- a/wp_cli/src/wp_cli_posts.rs
+++ b/wp_cli/src/wp_cli_posts.rs
@@ -4,6 +4,8 @@ use wp_serde_helper::deserialize_i64_or_string;
 
 use crate::run_wp_cli_command;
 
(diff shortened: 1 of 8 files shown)


wp_cli/src/wp_cli_posts.rs
```diff
@@ -4,6 +4,8 @@ use wp_serde_helper::deserialize_i64_or_string;
 
 use crate::run_wp_cli_command;
 
+const POST_FIELDS_ARG: &str = "--fields=ID, post_title, post_date, post_status, post_author, post_date_gmt, post_content, post_excerpt, comment_status, ping_status, post_password, post_modified, post_modified_gmt, guid, post_type";
```
oguzkocer, line 7: I missed the spaces in this argument. Having them doesn't impact this PR and I am removing them in the next PR, so it can be safely ignored.

crazytonyli (approved):

PR #304, opened 2024-09-11
Implements update `/posts` endpoint and its integration tests

oguzkocer:
Implements update `/posts` endpoint and its integration tests following established patterns. It provides 100% integration test coverage - as far as I am aware - for individual field updates. I believe `/posts` is one of the most important endpoints for us, and the extra coverage will provide helpful. I am even thinking about going back to create `/posts` endpoint and adding more tests for it.

I've changed the timezone of the test site, so that the `date` & `date_gmt` values would be different which should help us avoid related false positive tests.

The `meta.footnotes` return a JSON string that contains a JSON array. I've added a custom parser and serializer for this, so that we can work with strong types. It seems to be working well at the moment, but it's entirely possible that I am missing some conditions which hopefully will pop-up as we add more WordPress versions and test sites.

### DIFF
--- a/scripts/setup-test-site.sh
+++ b/scripts/setup-test-site.sh
@@ -108,6 +108,10 @@ wp plugin install classic-editor
 # Used in `test_posts_immut`. If the resulting ID changes, `PASSWORD_PROTECTED_POST_ID` needs to be updated
 wp post create --post_type=post --post_password=INTEGRATION_TEST --post_title=Password_Protected
 
+# Update the timezone, so that the `date` & `date_gmt` values will be different
+# Otherwise, the integration tests might result in false positives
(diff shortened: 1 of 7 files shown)


jkmassel:
Anything in `meta` should probably be left as a `HashMap<String, String>` – I don't think this library will have enough context to parse what's in there (unless it's generated as part of something in core?)

In that case, I wonder if a separate crate would be the way to go – it could contain a second-level set of parsers for common core-level data stored there?

oguzkocer:
We briefly discussed `post.meta` with @jkmassel on Slack and decided to tackle it as its own thing at a later time and it is now being tracked in https://github.com/Automattic/wordpress-rs/issues/306.

jkmassel (approved):

PR #305, opened 2024-09-11
Skip rust-cache step from GitHub Action Swift pipeline

crazytonyli:
The Swift part of the GitHub actions fails due to [no disk space](https://github.com/Automattic/wordpress-rs/actions/runs/10821373034/job/30023240356#step:8:634). The issue went away after I removed [the rust-cache step](https://github.com/Automattic/wordpress-rs/actions/runs/10821672827/job/30024168374?pr=305). It seems rust-cache is not optimized for taking less disk space: https://github.com/Swatinem/rust-cache/issues/158.

The Swift pipeline with and without rust-cache both take about 22m to complete. The kotline pipeline with rust-cache runs faster though.
- One of the commits in trunk: https://github.com/Automattic/wordpress-rs/actions/runs/10817741523/job/30011786468
- This PR: https://github.com/Automattic/wordpress-rs/actions/runs/10821672827/job/30024168374

Based on that, this PR disables rust-cache only for the Swift pipeline.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -69,11 +69,6 @@ jobs:
         # For more details on CodeQL's query packs, refer to: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
         # queries: security-extended,security-and-quality
 
-    - name: Restore Rust Cache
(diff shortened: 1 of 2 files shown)


oguzkocer (approved):

jkmassel (approved):
I'm not sure why this would result in such a huge cache, but sure – seems like a good fix

PR #308, opened 2024-09-13
Generate integration test credentials as JSON

oguzkocer:
I am going to need to add even more data to the test credentials as I work on `/posts` error handling and I think we are at a point where we can't rely on the line numbers and positioning anymore. By generating a JSON file, we can keep things maintainable.

This also simplifies the Rust build script that generates the test credentials. To make that one even simpler, I opted to switch to a `struct` instance.

I've updated the Kotlin integration test credentials, but decided to remove the Android one. The Android integration tests are only there to validate that the library is loading, so we don't really need authentication or any specific test credentials. Also note that the Android integration tests and the example app is currently broken. I'll address that in a separate PR.


jkmassel (approved):
This is a good plan – I didn't review the implementation too much, because if you got something wrong it'll break pretty clearly :)

PR #316, opened 2024-09-19
Add `/posts` related errors and their integration tests

oguzkocer:
This PR also includes a follow up to #308 by moving the password protected post information to `test_credentials.json`. I was adding a trashed post to test credentials and didn't want to leave that one as is.

Unfortunately a lot of error types had to go into the untested section. I believe at least some of these errors are impossible to get with a default WordPress installation. That's because the necessary conditions to retrieve these errors will trigger an error in a prior check. Having said that, it's possible that some of these errors are possible to test and I just couldn't figure out how.

I am hoping that when we get logging and tracking, we can figure out how some of these untested errors can be triggered and add them to our test suite.


native/kotlin/api/kotlin/src/integrationTest/kotlin/IntegrationTestHelpers.kt
```diff
@@ -8,7 +8,7 @@ const val SECOND_USER_ID: UserId = 2
 const val FIRST_USER_EMAIL = "test@example.com"
 const val SECOND_USER_EMAIL = "themeshaperwp+demos@gmail.com"
 const val SECOND_USER_SLUG = "themedemos"
-const val NUMBER_OF_USERS = 3
+const val NUMBER_OF_USERS = 4
```
jkmassel, line 11: Weird nit – I wonder if this should be `EXPECTED_NUMBER_OF_USERS`?

wp_api_integration_tests/src/lib.rs
```diff
@@ -14,13 +14,19 @@ include!(concat!(env!("OUT_DIR"), "/generated_test_credentials.rs"));
...
+    pub subscriber_username: &'static str,
+    pub subscriber_password: &'static str,
+    pub subscriber_password_uuid: &'static str,
+    pub author_username: &'static str,
+    pub author_password: &'static str,
+    pub password_protected_post_id: i32,
+    pub password_protected_post_password: &'static str,
+    pub password_protected_post_title: &'static str,
+    pub trashed_post_id: i32,
 }
```
jkmassel, line 30: This seems more performant, but less ergonomic since we have to do `.to_string()` all the time – WDYT?

jkmassel (approved):
Looks as expected, left some thoughts for your consideration

wp_api_integration_tests/src/lib.rs
oguzkocer: This change is due to the addition of a `i32` field, because I can no longer do the `.to_string()` part in `build.rs`. It may be possible to still use `String` if we feel strongly about avoiding `to_string()` everywhere. Let me know if you feel that way and I'll look into it.

native/kotlin/api/kotlin/src/integrationTest/kotlin/IntegrationTestHelpers.kt
oguzkocer: That makes sense, although I want to move all of these into the `test_credentials.json` file and in that file number of users will make more sense.

wp_api_integration_tests/src/lib.rs
jkmassel: Don't feel strongly about it 🤷

PR #344, opened 2024-10-16
Separate Kotlin bindings into its own module

oguzkocer:


oguzkocer:
This doesn't seem to be necessary at the moment, so I am closing the PR. We can re-open it if needed.

PR #353, opened 2024-10-22
Fix parsing url with trailing slash in ApiBaseUrl

crazytonyli:
### Issue

`ApiBaseUrl` has issue parsing url with trailing slash.

```rust
let api_base_url: ApiBaseUrl = "https://example.com/f/".try_into().unwrap();
assert_eq!(api_base_url.as_str(), "https://example.com/f/wp-json");
```

Result:
```
assertion `left == right` failed
  left: "https://example.com/f//wp-json"
 right: "https://example.com/f/wp-json"
```

### DIFF
--- a/wp_api/src/request/endpoint.rs
+++ b/wp_api/src/request/endpoint.rs
@@ -121,9 +121,8 @@ trait UrlExtension {
 }
 
 impl UrlExtension for Url {
-    fn append(mut self, segment: &str) -> Result<Url, ()> {
-        self.path_segments_mut()?.push(segment);
-        Ok(self)
+    fn append(self, segment: &str) -> Result<Url, ()> {
+        self.extend([segment])
     }
 
     fn extend<I>(mut self, segments: I) -> Result<Url, ()>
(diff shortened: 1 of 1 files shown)


wp_api/src/request/endpoint.rs
```diff
@@ -121,16 +121,22 @@ trait UrlExtension {
...
 
 impl UrlExtension for Url {
-    fn append(mut self, segment: &str) -> Result<Url, ()> {
-        self.path_segments_mut()?.push(segment);
-        Ok(self)
+    fn append(self, segment: &str) -> Result<Url, ()> {
+        self.extend([segment])
     }
 
     fn extend<I>(mut self, segments: I) -> Result<Url, ()>
```
crazytonyli, line 128: BTW, is there any reason why `UrlExtension` functions take `self` instead of `&self`?

oguzkocer: We can't mutate the `url` if we take `&self`. We can either take `&mut self` or `self`.

If we take `&mut self`, we won't be able to chain functions, so we take `self`. Also, `url` already works with `&mut self`, so we don't need extensions for it.

Does that make sense?

oguzkocer (approved):

wp_api/src/request/endpoint.rs
crazytonyli: Since the function return `Result<Url>`, does that mean the function can be declared as `fn append(&self, ...) -> Result<Url, ()>`, which looks more similar to other Rust code?

oguzkocer: If you do that, you'll have to clone the value because you can't modify the existing one. When we use `fn append(self)`, the original value is moved to the function so we are free to modify it and then we return the original value.

Keep in mind that these are compile time semantics for Rust. In practice, we are modifying the object in place just like we'd do with `&mut self` so the memory address doesn't change.

PR #355, opened 2024-10-23
Update `:api:android` configuration to correctly include the native binaries

oguzkocer:
In #236 we removed some configuration that was being handled by `org.mozilla.rust-android-gradle.rust-android` plugin. However, it turns out we still need this configuration. I am not sure exactly why and I don't think spending time investigating it would be helpful.

Reverting the changes from #236 fixes the example compose app as well as the Android integration test. It's very likely that WordPress-Android integration was broken as well, but I didn't verify this.

The PR also reverts compose plugin to `1.6.11` as the newer version was having a dependency resolution problem when I ran the compose app.

**To Test**
* Run the `UsersEndpointAndroidTest`
* Run the example compose Android app (don't forget to update the password in `AppModule.kt`)


jkmassel (approved):
I tested this with content-heavy.wpmt.co in the example app and it worked great!

PR #376, opened 2024-11-06
Pagination support for `/posts`

oguzkocer:
Follows up #375 with pagination support for `/posts`. It also adds extra Kotlin integration tests for `/posts`.

Please see #375 for details as this PR follows the patterns established in that PR.


jkmassel (approved):
Tested this with Swift and it worked. Everything is very similar to #375 so approving as well.

PR #380, opened 2024-11-12
Add Swift Pagination

jkmassel:
Adds basic Swift auto-pagination based on #375 and #376.

To Test:
- Launch the example app
- Sign into https://content-heavy.wpmt.co
- Note that > 100 users and posts are loaded

### DIFF
--- a/native/swift/Sources/wordpress-api/Pagination.swift
+++ b/native/swift/Sources/wordpress-api/Pagination.swift
@@ -42,13 +42,19 @@ extension PaginationAwareExecutor {
     public func paginatedWithEditContext(
         params: EditContextResponseType.ParamsType
     ) async throws -> [EditContextResponseType.DataType] {
-        var workingResponse = try await self.listWithEditContext(params: params)
-        var allObjects: [EditContextResponseType.DataType] = workingResponse.data
+        var allObjects: [EditContextResponseType.DataType] = []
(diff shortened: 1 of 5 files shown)


native/swift/Sources/wordpress-api/Pagination.swift
```diff
@@ -0,0 +1,135 @@
...
+    public func paginatedWithEditContext(
+        params: EditContextResponseType.ParamsType
+    ) async throws -> [EditContextResponseType.DataType] {
+        var workingResponse = try await self.listWithEditContext(params: params)
+        var allObjects: [EditContextResponseType.DataType] = workingResponse.data
+
+        while let nextPageParams = workingResponse.nextPageParams {
+            workingResponse = try await self.listWithEditContext(params: nextPageParams)
+            allObjects.append(contentsOf: workingResponse.data)
+        }
```
crazytonyli: Nitpick: using `do-while` to avoid duplicating `listWithEditContext` and `appending(contentsOf:...)` calls.

crazytonyli (approved):

native/swift/Sources/wordpress-api/Pagination.swift
jkmassel: Ah, that is nicer. Done in 098c1e8.

PR #382, opened 2024-11-13
Add async Swift Pagination

jkmassel:
Builds on #380 by adding ~~Combine and~~ `AsyncSequence`-based pagination.

You can try it out by running the sample app and loading posts for https://content-heavy.wpmt.co – you'll see the posts slowly load in as they're paginated (the code has the page size set to 5 so it's easier to see).

~~I put the `Combine` stuff in its own module so that it doesn't have to be loaded (for instance, it'd be annoying for Linux environments).~~

~~Depends on #386 to build properly on Linux because Swift 6 brings unified `Foundation`.~~


native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -68,11 +68,8 @@ public struct WordPressAPI {
     }
 
     package func perform(request: WpNetworkRequest) async throws -> WpNetworkResponse {
-        try await withCheckedThrowingContinuation { continuation in
-            self.perform(request: request) { result in
-                continuation.resume(with: result)
-            }
-        }
+        let (data, response) = try await self.urlSession.data(for: request.asURLRequest())
```
jkmassel, line 71: Swift Foundation has feature parity across platforms as of Swift 6, so we should just be able to use the async/await stuff everywhere now?

native/swift/Sources/wordpress-api/LoginAPI.swift
```diff
@@ -8,21 +8,84 @@ import WordPressAPIInternal
...
+public final class WordPressLoginClient: Sendable {
 
     public protocol Authenticator {
         func authenticate(url: URL, callbackURL: URL) async -> Result<URL, Error>
     }
 
     private static let callbackURL = URL(string: "x-wordpress-app://login-callback")!
 
-    public enum Error: Swift.Error {
+    public enum Error: Swift.Error, Equatable {
```
jkmassel, line 19: This `Equatable` conformance works for the test suite, but probably doesn't work _super_ well in production.

This shouldn't really matter because nobody should need that, but in a future PR I'd like to make this way more robust – the trick is that that a lot of the underlying types have different information, so comparing them isn't straightforward. I'd want to capture these from raw HTTP responses then write a whole bunch of tests. 

I don't think it's worth it at this point though, because nobody really needs that – it's just for the test suite.

native/swift/Sources/wordpress-api/Pagination.swift
```diff
@@ -106,20 +120,80 @@ extension PaginationAwareExecutor {
...
+public struct PaginationSequence<ResponseType: PaginatableResponse>: AsyncSequence {
+    public typealias Transformer = (ResponseType.ParamsType) async throws -> ResponseType
+
+    private let params: ResponseType.ParamsType
+    private let transform: Transformer
+
+    public init(params: ResponseType.ParamsType, transform: @escaping Transformer) {
+        self.params = params
+        self.transform = transform
+    }
```
crazytonyli, line 158: It's probably better to not make this initializer (and the `Transformer`) public, since it makes sense that only this library can create `PaginationSequence` instances.

native/swift/Sources/wordpress-api-combine/Stream.swift
```diff
@@ -0,0 +1,58 @@
...
+        do {
+            for try await objects in sequence {
+                accumulatedObjects.append(contentsOf: objects)
+                publisher.send(accumulatedObjects)
+            }
+        } catch {
+            publisher.send(completion: .failure(error))
+        }
+    }
+}
```
crazytonyli: The fact that the publisher doesn't emit elements unless `fetch` is called doesn't follow combine's programming model.

I don't think we really need this combine library/target, because it doesn't really relate to WordPressAPI. Its core is more like a generic solution of wrapping `AsyncSequence` in `Publisher`.

jkmassel: Yeah, agreed – we'll need some support for this on the Rust side, so I've removed it for now in [ba6ba90](https://github.com/Automattic/wordpress-rs/pull/382/commits/ba6ba903bc897dac6b89ed3e353bce2ee94960b9)

native/swift/Sources/wordpress-api/Pagination.swift
jkmassel: Addressed in [d597868](https://github.com/Automattic/wordpress-rs/pull/382/commits/d597868ae19fdffc4ba6dc280189c49ca906a4ee)

native/swift/Sources/wordpress-api/Pagination.swift
```diff
@@ -106,20 +120,80 @@ extension PaginationAwareExecutor {
...
+        params: EmbedContextResponseType.ParamsType
+    ) -> PaginationSequence<EmbedContextResponseType> {
+        PaginationSequence(params: params) { params in
+            try await self.listWithEmbedContext(params: params)
+        }
+    }
+}
+
+public struct PaginationSequence<ResponseType: PaginatableResponse>: AsyncSequence {
+    public typealias Transformer = (ResponseType.ParamsType) async throws -> ResponseType
```
crazytonyli, line 150: Remove this `public` too?

native/swift/Example/Example/ListViewModel.swift
```diff
@@ -60,3 +97,34 @@ struct MyError: LocalizedError {
         underlyingError.localizedDescription
     }
 }
+
+struct ListViewSequence: AsyncSequence {
```
crazytonyli, line 101: Is this struct implementation equavilant of `asyncSequence.compactMap({ ($0 as? [ListViewDataConvertable]).asListViewData() })`?

crazytonyli (approved):

PR #287, opened 2024-08-30
Add initial OAuth 2 support

jkmassel:
- Work in Progress -

PR #383, opened 2024-11-14
List `/media` endpoint

oguzkocer:
This PR implements list [`/media` endpoint](https://developer.wordpress.org/rest-api/reference/media/#list-media) by mostly following the established patterns. Here are a few things worth noting:

* 0b8b06c874953dfeb17d9b2c54be91e6fa05f10d implements `FromStr` for `PluginStatus`. I was planning to work on `/plugins` pagination and later realized that the endpoint doesn't support pagination. Since this is a useful addition, I didn't want to throw it away and decided to include in this PR.
* `MediaId` was temporarily implemented in `wp_api/src/posts.rs` but now moved to `wp_api/src/media.rs`.
* A new `JsonValue` type is introduced. The `media_details` field in `/media` has a dynamic type without a clear set of fields. This new `JsonValue` type will be used in such cases. @jkmassel mentioned that he had to implement something similar while working in #287 and we agreed that this generic version is a reasonable addition.
* The PR implements both `MediaType` & `MediaTypeParam` types. This is because the server accepts values such as `image, video, text, application, audio` to filter the media list, but it returns `image` or  `file`.
* There is likely an error in the `status` of [`/media` documentation](https://developer.wordpress.org/rest-api/reference/media/). The `Schema` section indicates `status` is `One of: publish, future, draft, pending, private`, however using `status=publish` will result in `InvalidParam` error saying `status[0] is not one of inherit, private, and trash.`. Our test site also returns `inherit` for every media object. That's why `MediaStatus` is only implemented to have `inherit, private, trash` variants - with the addition of `Custom(String)` variant for fallback.

### DIFF
--- a/native/kotlin/api/kotlin/src/integrationTest/kotlin/MediaEndpointTest.kt
+++ b/native/kotlin/api/kotlin/src/integrationTest/kotlin/MediaEndpointTest.kt
@@ -2,7 +2,6 @@ package rs.wordpress.api.kotlin
 
 import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Test
-import uniffi.wp_api.JsonValue
 import uniffi.wp_api.MediaListParams
 import uniffi.wp_api.wpAuthenticationFromUsernameAndPassword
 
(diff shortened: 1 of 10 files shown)


jkmassel (approved):
Code looks pretty straightforward – I tried it out from Swift (as usual) and it worked great as usual.

PR #386, opened 2024-11-14
Upgrade project to Swift 6

jkmassel:
Updates the project to Swift 6, which brings a few benefits:

1. We can remove our Linux shims – `Foundation` is now the same across platforms.
2. We can use `swift-testing`, which makes our test suite a lot simpler


native/swift/Tests/wordpress-api/WordPressAPITests.swift
```diff
@@ -47,33 +45,6 @@ final class WordPressAPITests: XCTestCase {
             executor: stubs
         )
         let user = try await api.users.retrieveWithViewContext(userId: 1)
-        XCTAssertEqual(user.data.name, "User Name")
-    }
-
-#if !os(Linux)
-    // Skip on Linux, because `XCTExpectFailure` is unavailable on Linux
-    func testTimeout() async throws {
```
jkmassel, line 55: This test was in the project twice, so I removed it here

native/swift/Tests/wordpress-api/Support/HTTPStubs.swift
```diff
@@ -1,15 +1,17 @@
...
 
-class HTTPStubs: SafeRequestExecutor {
+    typealias Stub = (condition: @Sendable (WpNetworkRequest) -> Bool, response: WpNetworkResponse)
 
-    var stubs: [(condition: (WpNetworkRequest) -> Bool, response: WpNetworkResponse)] = []
+    private let stubs: [Stub]
+    private let missingStub: Result<WpNetworkResponse, Error>?
 
-    var missingStub: Result<WpNetworkResponse, Error>?
+    init(stubs: [Stub] = [], missingStub: Result<WpNetworkResponse, Error>? = nil) {
```
jkmassel, line 11: Because `SafeRequestExecutor` needs to be `Sendable`, we need our stubs to be too.

The easiest way to do that is to just make them immutable, which doesn't change the code all that much

native/swift/Tests/wordpress-api/Support/Extensions.swift
```diff
@@ -0,0 +1,25 @@
...
+        try! WpNetworkHeaderMap.fromMap(hashMap: [:])
+    }
+
+    static func withLinkHeader(_ value: String) -> WpNetworkHeaderMap {
+        // swiftlint:disable:next force_try
+        try! WpNetworkHeaderMap.fromMap(hashMap: ["Link": value])
+    }
+}
+
+// This might not be safe – we should figure that out
```
jkmassel, line 16: These are children of the `HttpStubs`, so they need to be `Sendable`. They're not safe to be marked as such though because they're `uniffi::Object`-attributed, so we'll just do it in the test suite for now because we own that and can guarantee how it'll be used.

crazytonyli (approved):
🥳

PR #393, opened 2024-11-14
[Tooling] Call `use-bot-for-git` in the release script 

iangmaia:
I've noticed we're using a SSH deploy key with write permission so that we can [`git push` during the release process](https://github.com/Automattic/wordpress-rs/blob/b54a394bb4cc2155f9d136f6658a35252589585d/fastlane/Fastfile#L77).

This PR adds a call to `use-bot-for-git` in the beginning of the release script so that we can perform the `git push` operation in our trusted environment and revert this repo's key to read-only.
While at it, I've also moved the Xcode version to `shared-pipeline-vars` (I considered using `16.1` already but I'll let that to the team).

See also:
- 554-gh-Automattic/buildkite-ci (⚠️ **needs to be merged + deployed before merging the present PR**)
- 39-gh-Automattic/apps-infra-plans
- paaHJt-6EP-p2

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -68,13 +68,13 @@ steps:
           - target/libwordpressFFI.xcframework.zip
           - native/swift/Sources/wordpress-api-wrapper/wp_api.swift
         env:
-          IMAGE_ID: xcode-16.0-v7
+          IMAGE_ID: $IMAGE_ID
         agents:
           queue: mac
       - label: ":swift: :darwin: Build + Test on Simulators"
         command: .buildkite/swift-test.sh run_tests
         env:
-          IMAGE_ID: xcode-16.0-v7
+          IMAGE_ID: $IMAGE_ID
         depends_on: xcframework
         plugins: [$CI_TOOLKIT]
         agents:
@@ -82,7 +82,7 @@ steps:
       - label: ":swift: :darwin: Build for Real Devices"
         command: .buildkite/swift-test.sh build_for_real_device
         env:
-          IMAGE_ID: xcode-16.0-v7
+          IMAGE_ID: $IMAGE_ID
         depends_on: xcframework
         plugins: [$CI_TOOLKIT]
         agents:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -68,29 +68,29 @@ steps:
           - target/libwordpressFFI.xcframework.zip
           - native/swift/Sources/wordpress-api-wrapper/wp_api.swift
         env:
-          IMAGE_ID: xcode-16.0-v7
+          IMAGE_ID: $IMAGE_ID
```
AliSoftware: Nice DRYing!

💄 Nowadays we usually also move the `env:` section for `IMAGE_ID: $IMAGE_ID` at the top of the `pipeline.yml` files so that it's common to all steps of that `.yml` file.
(sure, that means that the env var is _also_ passed to steps that use agents _other_ than `mac`, and for which defining this `IMAGE_ID` is useless… but it's also harmless and allows some more DRYing, so…)

.buildkite/shared-pipeline-vars
```diff
@@ -1,7 +1,12 @@
...
 
- export CI_TOOLKIT="automattic/a8c-ci-toolkit#3.5.0"
- export TEST_COLLECTOR="test-collector#v1.10.2"
+XCODE_VERSION="16.0-v7"
+CI_TOOLKIT_PLUGIN_VERSION="3.7.1"
+TEST_COLLECTOR_VERSION="v1.10.2"
+
+export IMAGE_ID="xcode-$XCODE_VERSION"
+export CI_TOOLKIT="automattic/a8c-ci-toolkit#$CI_TOOLKIT_PLUGIN_VERSION"
+export TEST_COLLECTOR="test-collector#$TEST_COLLECTOR_VERSION"
```
AliSoftware, line 12: 👌

AliSoftware (approved):
Left a small nitpick, otherwise LGTM 👍

.buildkite/pipeline.yml
iangmaia: > sure that means that env var is also passed to steps that use agents other than mac and for which defining this IMAGE_ID is useless… but it's also harmless and allows some more DRYing, so…

Yeah, that's what I considered in principle (passing IMAGE_ID to other agents that don't use it, in principle with no harm), but indeed it looks a lot cleaner. Updated on e8936045c164074c6015160827bb3caed5830cd4

iangmaia:
554-gh-Automattic/buildkite-ci has been merged and deployed, so I'm merging this PR.

PR #395, opened 2024-11-14
Export `parse_as_response_type` functions

oguzkocer:
We allow two different ways to work with this library.

The main way is to use the `WpApiClient` type which exposes endpoint specific request executors which executes the request and parses the response as such:
```
let api_client = WpApiClient::new(/*_arguments_*/);
let result: Result<UsersRequestListWithEditContextResponse, WpApiError> = api_client
    .users()
    .list_with_edit_context(&UserListParams::default())
    .await;
```

We also provide an alternative way where building & executing the request as well as parsing it has to be done in individual steps. We always allowed building the request with the `WpApiRequestBuilder`, however one had to manually parse it. This is relatively easy to do in Rust by using the `WpNetworkResponse::parse` function, but native wrappers had to implement a parser of their own.

This PR generates these individual parsers, so native wrappers can use the Rust response parser. Here is how that looks in `Kotlin`:
```
val requestBuilder = UniffiWpApiRequestBuilder(siteUrl, authentication)
val userListRequest = requestBuilder.users().listWithEditContext(UserListParams())
val userListResponse = requestExecutor.execute(userListRequest)
val userList = parseAsUsersRequestListWithEditContextResponse(userListResponse).data
```

This looks very similar to the `WpApiClient` approach, but here we execute the given request ourselves and then pick the correct parser. Note that this approach isn't recommended unless you need the extra flexibility because it's both easier and safer to use the `WpApiClient` since it'll choose the correct parser for you.

### DIFF
--- a/native/kotlin/api/kotlin/build.gradle.kts
+++ b/native/kotlin/api/kotlin/build.gradle.kts
@@ -39,6 +39,7 @@ testing {
             dependencies {
                 implementation(project())
 
+                implementation(libs.okhttp)
                 implementation(rootProject.libs.kotlin.test)
                 implementation(rootProject.libs.kotlinx.coroutines.test)
                 implementation(libs.kotlinx.serialization)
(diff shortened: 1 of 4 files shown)


jkmassel (approved):
This works great (as tested by #396)

PR #396, opened 2024-11-15
Try adding Combine again

jkmassel:


native/swift/Sources/wordpress-api/Protocols/PublisherAwarePerformer.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        return paramsPublisher                                    // Following Combine chains can be tricky so:
+            .flatMap { perform(request: requestTransformer($0)) } // 1. `dataTaskPublisher` replaces `paramsPublisher`
+            .tryMap { try responseTransformer($0) }               // 2. Map the raw response back to the expected type
+            .handleEvents(receiveOutput: { output in              // 3. When a new event comes in...
+                if let nextPageParams = output.nextPageParams {   //    If there's another page...
+                    paramsPublisher.send(nextPageParams)          //       Kick off the paramsPublisher again
+                } else {                                          //    Otherwise we're done
+                    paramsPublisher.send(completion: .finished)   //       And paramsPublisher is complete
+                }
```
crazytonyli, line 90: Publishers are supposed to be "pure", without side-effect. I think that's broken by this `handleEvents` code here.

For example, both sink calls in the code below should return the exact same result:

```swift
let publisher = [1, 2, 3].publisher
publisher.sink { ... } // Output: 1, 2, 3

DispatchQueue.global.asyncAfter(.now + .senconds(3)) {
  publisher.sink { ... } // Output: 1, 2, 3
}
```

If you swap the `publisher` to be `let publisher = recursivePublisher(...)`, the `sink` calls would behave differently. (I didn't run the code, please correct me if I'm wrong).

```swift
let publisher = recursivePublisher(...)
publisher.sink { ... } // Output: API call result.

// 3 Seconds later, after the above API call finishes.
DispatchQueue.global.asyncAfter(.now + .senconds(3)) {
    publisher.sink { ... } // Unexpected: No element will be emitted here.
}
```

oguzkocer:
Swift wrapper is using async stream API, so this PR is no longer needed.

PR #397, opened 2024-11-16
Update `/media` endpoint

oguzkocer:
Implements update `/media` endpoint following the established patterns. There are a few things worth noting:

* [`wp media`](https://developer.wordpress.org/cli/commands/media/) doesn't support retrieving or listing the media, so the integration tests don't validate whether the value was actually updated. It'll only ensure that the request was successful.
* Switches the test site theme to `twentytwentyfour` which should fix the failing integration test that updates the post `template` field
* Adds `--no-format` argument to Kotlin bindings generation Gradle task. This will significantly speed up the Kotlin bindings generation when `ktlint` is installed - somewhere from ~10 minutes to under a second.

### DIFF
--- a/wp_api_integration_tests/tests/test_media_mut.rs
+++ b/wp_api_integration_tests/tests/test_media_mut.rs
@@ -1,0 +1,97 @@
+use macro_helper::generate_update_test;
+use serial_test::serial;
+use wp_api::{
+    media::MediaUpdateParams,
+    posts::{PostCommentStatus, PostPingStatus, PostStatus},
+};
+use wp_api_integration_tests::{
+    api_client, backend::RestoreServer, AssertResponse, FIRST_POST_ID, MEDIA_ID_611,
+};
+
+generate_update_test!(update_date, date, "2024-09-09T12:00:00".to_string());
+
+generate_update_test!(update_date_gmt, date_gmt, "2024-09-09T12:00:00".to_string());
+
+generate_update_test!(update_slug, slug, "new_slug".to_string());
+
+generate_update_test!(update_status_to_draft, status, PostStatus::Draft);
+generate_update_test!(update_status_to_future, status, PostStatus::Future);
+generate_update_test!(update_status_to_pending, status, PostStatus::Pending);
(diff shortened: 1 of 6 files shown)


wp_api_integration_tests/tests/test_media_mut.rs
```diff
@@ -0,0 +1,97 @@
...
+    ping_status,
+    PostPingStatus::Closed
+);
+
+// TODO: `POST_TEMPLATE_SINGLE_WITH_SIDEBAR` doesn't work for `/media`.
+//generate_update_test!(
+//    update_template,
+//    template,
+//    POST_TEMPLATE_SINGLE_WITH_SIDEBAR.to_string()
+//);
```
oguzkocer, line 54: I can't figure out what would be a valid value for the `media.template` field. I could remove this test entirely, but I'd love to have a test confirming that the field is working as expected. Having said that, the error we get back from the request confirms that we are using the correct field name, so I don't want to invest too much time into it.

Here is a curl command that reproduces this issue:
```
curl --user test@example.com:$(jq .admin_password test_credentials.json) -H "Content-Type: application/json" -d '{"template":"single-with-sidebar"}' http://localhost/wp-json/wp/v2/media/611
```

And here is the response:
```
{
  "code": "rest_invalid_param",
  "message": "Invalid parameter(s): template",
  "data": {
    "status": 400,
    "params": {
      "template": "template is not one of ."
    },
    "details": {
      "template": {
        "code": "rest_invalid_param",
        "message": "template is not one of .",
        "data": null
      }
    }
  }
}
```

jkmassel: I can't imagine a valid template for the `media` post type tbh

I'd be fine to ignore this parameter until/unless we find out about one

jkmassel (approved):

wp_api_integration_tests/tests/test_media_mut.rs
oguzkocer: Sounds good. I'll delete the `TODO` then as I only left it there to make sure I didn't forget to bring it up.

Thanks for the review!

PR #410, opened 2024-11-27
Create `/media` endpoint

oguzkocer:
Implements create `/media` endpoint. Due to its custom nature, it does NOT follow the established patterns, but instead requires some new patterns:

* Adds a custom `MediaUploadRequest` that's similar to `WpNetworkRequest`, but it includes information such as `file_path` & `file_content_type`.
* Adds a new error type `MediaUploadRequestExecutionError` which is similar to `RequestExecutionError` but it has an extra variant `MediaFileNotFound`. We can add more error types specific to media uploads here.
* Adds a new method to `RequestExecutor` to upload media that uses the new `MediaUploadRequest` and `MediaUploadRequestExecutionError`.
* `MediaUploadRequest` uses `multipart/form-data` content type header value and adds the parameters as form values. The header value is included in its `header_map` and the key value pairs for `MediaCreateParams` can be accessed from its `media_params()` function.
* Adds Kotlin & Rust integration tests to verify that media upload is working correctly.

I am a little concerned about the `clone`s in `WpNetworkRequest` & `MediaUploadRequest` and not entirely sure about the correct way to do this. From pure Rust perspective, this is definitely wrong. However, these types have to be cloned for FFI - or shared as a pointer. So, maybe we need to separate how Rust uses it and how it's exposed through FFI. I don't want to hold up this PR for this, so I opened a separate issue for it: https://github.com/Automattic/wordpress-rs/issues/416


native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -29,6 +35,44 @@ class WpRequestExecutor(
...
+    override suspend fun uploadMedia(mediaUploadRequest: MediaUploadRequest): WpNetworkResponse =
+        withContext(dispatcher) {
+            val requestBuilder = Request.Builder().url(mediaUploadRequest.url())
+            val multipartBodyBuilder = MultipartBody.Builder()
+                .setType(MultipartBody.FORM)
+            mediaUploadRequest.mediaParams().forEach { (k, v) ->
+                multipartBodyBuilder.addFormDataPart(k, v)
+            }
+            val file =
+                WpRequestExecutor::class.java.classLoader?.getResource(mediaUploadRequest.filePath())?.file?.let {
```
oguzkocer: I am not sure if this is the correct way to get the file in Android. I _think_ it's correct, but we might need to pass a function in the initializer so it can be configured.

I am leaving it as is for now and we can address it if we find that it's not working when we integrate it to WordPress-Android.

jkmassel (approved):
This works for me over in #412.

The code looks fine, the Android stuff included. That might be worth adding to the example app to be sure, but that could be a subsequent PR.

I took the liberty of pushing a commit to put Swift stubs in place to unblock CI.

PR #411, opened 2024-11-27
Add explicit Sendable conformance to WordPressAPI

crazytonyli:
To address warnings like

> Non-sendable type 'WordPressAPI' in implicitly asynchronous access to actor-isolated property 'api' cannot cross actor boundary; this is an error in the Swift 6 language mode

### DIFF
--- a/native/swift/Sources/wordpress-api/WordPressAPI.swift
+++ b/native/swift/Sources/wordpress-api/WordPressAPI.swift
@@ -1,13 +1,13 @@
 import Foundation
 #if canImport(WordPressAPIInternal)
-import WordPressAPIInternal
+@preconcurrency import WordPressAPIInternal
 #endif
 
 #if os(Linux)
 import FoundationNetworking
 #endif
 
-public struct WordPressAPI {
+public struct WordPressAPI: Sendable {
 
     enum Errors: Error {
         case unableToParseResponse
--- a/native/swift/Sources/wordpress-api/WordPressAPI.swift
+++ b/native/swift/Sources/wordpress-api/WordPressAPI.swift
@@ -7,7 +7,7 @@ import Foundation
 import FoundationNetworking
 #endif
 
-public struct WordPressAPI: Sendable {
+public actor WordPressAPI {
 
     enum Errors: Error {
         case unableToParseResponse
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,4 +1,4 @@
-swiftlint_version: 0.55.1
+swiftlint_version: 0.57.1
 strict: true
 included:
   - native/swift
--- a/Package.resolved
+++ b/Package.resolved
@@ -40,10 +40,10 @@
     {
       "identity" : "swift-syntax",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-syntax.git",
+      "location" : "https://github.com/swiftlang/swift-syntax.git",
       "state" : {
-        "revision" : "303e5c5c36d6a558407d364878df131c3546fad8",
-        "version" : "510.0.2"
+        "revision" : "cb53fa1bd3219b0b23ded7dfdd3b2baff266fd25",
+        "version" : "600.0.0"
       }
     },
     {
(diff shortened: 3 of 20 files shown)


native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -1,13 +1,13 @@
...
-import WordPressAPIInternal
+@preconcurrency import WordPressAPIInternal
 #endif
 
 #if os(Linux)
 import FoundationNetworking
 #endif
 
-public struct WordPressAPI {
+public struct WordPressAPI: Sendable {
```
jkmassel: I'm not sure if we can guarantee that this is actually `Sendable` because of the underlying pointer to a Rust object.

Would making this an actor address the issue? That way it'd never need to cross thread boundaries.

WDYT?

crazytonyli: The generated Swift code won't conform to `Sendable`, but they should be safe to use in a multi-thread environment. @oguzkocer Please correct me if I'm wrong, does most, if not all, structs in the library automatically implements `Send + Sync`? On top of that, we use things like Arc and Mutex for sharing internvally. Does that mean most structs are safe to use in a multi-thread environment?

crazytonyli: > I'm not sure if we can guarantee that this is actually Sendable because of the underlying pointer to a Rust object.

Declaring `WordPressAPI` as as an actor (which implies it's also Sendable) won't address the issue of whether Rust types are actually Sendable/thread-safe.

But using an actor means we'll need to call `await api.users()`, instead of `api.users()`. Of course,  that's not a big deal, but the `await` feels unnecessary.

crazytonyli: I don't feel strongly about either ways, but I'm more leaning towards declaring it as `Sendable`, instead of an actor. Because its implementation does not match the semantics of an actor, considering it's like a proxy to the `WpApiClient` Rust type.

jkmassel: My thinking about the actor approach is that an Actor would never be copied – instead, objects get data from it in a way that's guaranteed to be thread-safe.

crazytonyli: Hmm, I don't see any difference in terms of thread-safety, though. `struct WordPressAPI` would be thread-safe too.

If you'd like to avoid copying, then we have to go with `actor`. But I'm not sure how much we need to care about Swift copying structs when they are passed around.

oguzkocer: > The generated Swift code won't conform to Sendable, but they should be safe to use in a multi-thread environment. @oguzkocer Please correct me if I'm wrong, does most, if not all, structs in the library automatically implements Send + Sync? On top of that, we use things like Arc and Mutex for sharing internvally. Does that mean most structs are safe to use in a multi-thread environment?

This is correct and as far as I know `uniffi` [doesn't allow](https://mozilla.github.io/uniffi-rs/latest/internals/design_principles.html) `!Send` or `!Sync` to be used:

> Example: We insist that all object instances exposed to foreign-language code be Sync and Send, so that they're safe to access regardless of the threading model of the calling code. We do not allow thread-safety guarantees to be deferred to assumptions about how the code is called.
>
> Example: We do not allow returning any borrowed data from function calls, because we can't make any guarantees about when or how the foreign-language could access it.

For example:

```
#[derive(uniffi::Object)]
pub struct Bar {
    counter: PhantomData<std::rc::Rc<u32>>,
}
```

results in an error such as:
```
139 | #[derive(uniffi::Object)]
    |          ^^^^^^^^^^^^^^ `Rc<u32>` cannot be shared between threads safely
    |
    = help: within `Bar`, the trait `Sync` is not implemented for `Rc<u32>`, which is required by `Bar: Sync`
```

---

Having said that I am not sure if we can interpret that all binding code is safe to share between threads. [This bit](https://mozilla.github.io/uniffi-rs/latest/internals/design_principles.html) says:

> The generated bindings need to be safe by default. It should be impossible for foreign-language code to trigger undefined behaviour in Rust by calling the public API of the generated bindings, even if it is called in egregiously wrong or malicious ways. We will accept reduced performance in the interests of ensuring this safety.

which sounds like we can. I think we have 2 choices and I am fine with either one:

1. We can take this at face value and assume it's fine to use from multiple threads and then keep an eye on Sentry. If we see any crashes related to it, we can quickly roll a hotfix and act accordingly.
2. We can try and test it ourselves. Ideally, we'd have a unit test validating that this is working as expected, but I am not sure if that's possible or not.

jkmassel (approved):
I think we've sorted out all the issues here – anything left we can address in a future PR

PR #412, opened 2024-11-27
Swift Create Media Support

jkmassel:


native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -37,47 +37,58 @@ public final class WpRequestExecutor: SafeRequestExecutor {
...
 
-            return .failure(.RequestExecutionFailed(
-                statusCode: nil,
-                redirects: nil,
-                reason: .genericError(errorMessage: error.localizedDescription)
-            ))
-       }
+    public func uploadMedia(mediaUploadRequest: MediaUploadRequest) async throws -> WpNetworkResponse {
+        let urlrequest = try await mediaUploadRequest.asUrlRequest()
+        let (_, response) = try await self.fetch(request: self.preflight(urlrequest))
```
crazytonyli: [`URLSession.uploadTask`](https://developer.apple.com/documentation/foundation/urlsession/uploadtask(with:from:completionhandler:)) is probably more appropriate for uploading media. We can choose to stream from a file or memory. Things may go wrong when users choose to upload a video if we always read the file content into memory.

jkmassel: We're not currently reading the file into memory – instead, we make a file that contains the entire request body then provide that as an `InputStream` on the `URLRequest`. 

It might make sense to separate binary upload from the file metadata though, in which case we could definitely use an upload task

jkmassel:
Closing in favour of #717

PR #414, opened 2024-11-29
Rework API and app for Sendable and remove Result

jkmassel:
Some suggestions for #413:

1. I realized that the `ListViewDataConvertable` `Identifiable` conformance doesn't make sense – `ListViewData` should be `Identifiable` but  `ListViewDataConvertable` shouldn't. That's my bad in the original design – this PR corrects it.
2. I wasn't sure about having `globalInstance` be `@MainActor` because there's no real reason to synchronize with the main thread for access – this PR removes that.
3. There was a bunch of API that used `Result` types, which is weird in an async/await context, so I've adjusted it. The mock object retains the `Result` type so that its initializer can be expressed as a single argument.

### DIFF
--- a/native/swift/Sources/wordpress-api/LoginAPI.swift
+++ b/native/swift/Sources/wordpress-api/LoginAPI.swift
@@ -11,78 +11,10 @@ import FoundationNetworking
 public final class WordPressLoginClient {
 
     public protocol AuthenticatorProtocol {
-        func authenticate(url: URL, callbackURL: URL) async throws -> URL
+        func authenticate(url: URL, callbackURL: URL) async throws(WordPressLoginClientError) -> URL
     }
 
     private static let callbackURL = URL(string: "x-wordpress-app://login-callback")!
-
-    public enum Error: Swift.Error {
-        case invalidSiteAddress(UrlDiscoveryError)
-        case missingLoginUrl
-        case authenticationError(OAuthResponseUrlError)
-        case invalidApplicationPasswordCallback
-        case cancelled
-        case unknown(Swift.Error)
-
-        func isAutodiscoveryError() -> Bool {
-            guard case let .invalidSiteAddress(urlDiscoveryError) = self else {
-                return false
-            }
-
-            guard case .UrlDiscoveryFailed = urlDiscoveryError else {
-                return false
-            }
-
-            return true
-        }
-
-        var isFailedToFetchApiDetails: Bool {
-            guard
-                case let .invalidSiteAddress(urlDiscoveryError) = self,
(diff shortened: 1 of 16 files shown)


native/swift/Sources/wordpress-api/LoginAPI.swift
```diff
@@ -10,8 +10,8 @@ import FoundationNetworking
 
 public final class WordPressLoginClient {
 
-    public protocol Authenticator {
-        func authenticate(url: URL, callbackURL: URL) async -> Result<URL, Error>
+    public protocol AuthenticatorProtocol {
+        func authenticate(url: URL, callbackURL: URL) async throws -> URL
```
crazytonyli: I used `Result` so that we can easily change the API to `throws(Error) -> URL`, but I forgot to do it...

Now that you are changing this, do you mind adding the error type here: `throws(Error)`? You'll probably need to undo your changes first because it should be more straightforward to update based on the original code, like changing `return result` to `return try result.get()` should be sufficient.

native/swift/Tests/wordpress-api/LoginTests.swift
```diff
@@ -42,15 +43,21 @@ class LoginTests {
...
-            guard let loginError = error as? WordPressLoginClient.Error, case .invalidSiteAddress = loginError else {
+                authenticator: MockAuthenticator
+                    .returning(
+                        .failure(WordPressLoginClient.Error.invalidSiteAddress(.UrlDiscoveryFailed(attempts: .init())))
+                    )
+            )
+        },
+ throws: { error in
+            guard let loginError = error as? WordPressLoginClient.Error,
+ case .invalidSiteAddress = loginError else {
```
crazytonyli: Fix indentations (performing, throws, case) here?

jkmassel: Should be sorted!

native/swift/Sources/wordpress-api/LoginAPI.swift
jkmassel: Yep, I ran around and adjusted it – I also broke the error out into its own file since it's pretty big (and I suspect may get bigger)

jkmassel: There was also some weird SwiftLint behaviour when I made the `throws` typed – it was yelling about nested type depth. 

Updating `SwiftLint` fixed it

jkmassel:
@crazytonyli I think this is ready for another look? Feel free to merge yourself if you're happy with it

crazytonyli (approved):

PR #423, opened 2024-12-03
List `/comments` endpoint

oguzkocer:
Implements list `/comments` endpoint following the established patterns.

### DIFF
--- a/wp_api/src/comments.rs
+++ b/wp_api/src/comments.rs
@@ -434,10 +434,12 @@ mod tests {
     #[case(generate!(CommentListParams, (page, Some(2))), "page=2")]
     #[case(generate!(CommentListParams, (per_page, Some(2))), "per_page=2")]
     #[case(generate!(CommentListParams, (search, Some("foo".to_string()))), "search=foo")]
+    #[case(generate!(CommentListParams, (search, Some("foo + bar".to_string()))), "search=foo+%2B+bar")]
     #[case(generate!(CommentListParams, (after, Some("2023-08-14 17:00:00.000".to_string()))), "after=2023-08-14+17%3A00%3A00.000")]
     #[case(generate!(CommentListParams, (author, vec![UserId(1), UserId(2)])), "author=1%2C2")]
     #[case(generate!(CommentListParams, (author_exclude, vec![UserId(1), UserId(2)])), "author_exclude=1%2C2")]
     #[case(generate!(CommentListParams, (author_email, Some("foo".to_string()))), "author_email=foo")]
+    #[case(generate!(CommentListParams, (author_email, Some("foo+bar@example.com".to_string()))), "author_email=foo%2Bbar%40example.com")]
     #[case(generate!(CommentListParams, (before, Some("2023-08-14 17:00:00.000".to_string()))), "before=2023-08-14+17%3A00%3A00.000")]
     #[case(generate!(CommentListParams, (exclude, vec![CommentId(1), CommentId(2)])), "exclude=1%2C2")]
     #[case(generate!(CommentListParams, (include, vec![CommentId(1), CommentId(2)])), "include=1%2C2")]
@@ -471,7 +473,7 @@ mod tests {
             after: Some("d_a".to_string()),
             author: vec![UserId(111), UserId(112)],
             author_exclude: vec![UserId(211), UserId(212)],
-            author_email: Some("a_email".to_string()),
+            author_email: Some("a_email@example.com".to_string()),
             before: Some("d_b".to_string()),
             exclude: vec![CommentId(1111), CommentId(1112)],
             include: vec![CommentId(2111), CommentId(2112)],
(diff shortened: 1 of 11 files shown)


wp_api/src/comments.rs
```diff
@@ -0,0 +1,497 @@
...
+    pub parent: Option<CommentId>,
+    #[WpContext(edit, view)]
+    pub post: Option<PostId>,
+    #[WpContext(edit, view)]
+    pub status: Option<String>,
+    #[serde(rename = "type")]
+    #[WpContext(edit, embed, view)]
+    pub comment_type: Option<CommentType>,
+    #[WpContext(edit, embed, view)]
+    pub author_avatar_urls: Option<CommentAuthorAvatarUrls>,
```
crazytonyli, line 342: The type here should be the same as the one used in user

https://github.com/Automattic/wordpress-rs/blob/b18099c95cc970cb4a93a79669cb68fbafec0e31/wp_api/src/users.rs#L474

wp_api/src/comments.rs
```diff
@@ -0,0 +1,497 @@
...
+    #[case(generate!(CommentListParams, (comment_type, Some(CommentType::Custom("foo".to_string())))), "type=foo")]
+    #[case(generate!(CommentListParams, (password, Some("foo".to_string()))), "password=foo")]
+    #[case(CommentListParams {
+            page: Some(11),
+            per_page: Some(22),
+            search: Some("s_q".to_string()),
+            after: Some("d_a".to_string()),
+            author: vec![UserId(111), UserId(112)],
+            author_exclude: vec![UserId(211), UserId(212)],
+            author_email: Some("a_email".to_string()),
```
crazytonyli: Question: how should an email like `foo+bar@example.com` be encoded as a URL query? As it is, or `foo%2B@example.com`? Does WordPress accept both?

Similarly, if the user types "Comments about + sign" to search, what's the expected content of the `search` query?

wp_api/src/comments.rs
oguzkocer: That's a good point.

I am not sure which type we should prefer. `HashMap<String, String>` would be the safer choice, but we'll get strong types with `UserAvatarUrls`. We can also switch to `HashMap<UserAvatarUrlSize, String>` where `UserAvatarUrlSize` is an `enum`, so we can support custom sizes and still have some strong type information.

@jkmassel Do you know how customizable this field is? Do you have a preference?

wp_api/src/comments.rs
oguzkocer: > Question: how should an email like foo+bar@example.com be encoded as a URL query? As it is, or foo%2B@example.com?

It'll be encoded as `foo%2Bbar%40example.com`

> Does WordPress accept both?

Although I don't know the answer to this question, I think our current implementation works as expected. For example, if I update the [`email` field used in create user integration test](https://github.com/Automattic/wordpress-rs/blob/2f2cf1d22c196623ec5fcfa8db6c608134ada9a0/wp_api_integration_tests/tests/test_users_mut.rs#L14) to `t_email+foo@example.com`, the created user has the correct email address:

Screenshot 2024-12-09 at 12 09 01

> Similarly, if the user types "Comments about + sign" to search, what's the expected content of the search query?

Pretty much every field will get url encoded by [the `url` crate](https://docs.rs/url/latest/url/) which follows the [URL standard.](https://url.spec.whatwg.org/)

I've added a few more tests about these in 6c3d86c61018733c3b7a18993f43afb6438e34db.

wp_api/src/comments.rs
crazytonyli: It's not a surprise that plugins can customize avatar sizes (thus changing the response schema). https://github.com/WordPress/wordpress-develop/blob/faff3a2085bda2413547813172e5902c93389830/src/wp-includes/rest-api.php#L1298

oguzkocer: I see. In that case, I think the `enum` solution would work best. It's an approach we use in several other places where we want stronger types, but we want to ensure we support customizing the fields. I'll make that change.

wp_api/src/comments.rs
crazytonyli: > Pretty much every field will get url encoded by [the url crate](https://docs.rs/url/latest/url/) which follows the [URL standard.](https://url.spec.whatwg.org/)

The reason I asked was different web apps treat `+` differently and we need to make sure we are encoding URL queries the way WordPress expects. For example, Google encodes whitespaces as '+' ([example](https://www.google.com/search?q=foo+bar)), instead of its percentage encode value '%20', which is against "the standard". That means a URL like `https://www.google.com/search?q=foo+bar%40email.com` makes Google think the search is `foo bar@email.com`

> The query percent-encode set is the [C0 control percent-encode set](https://url.spec.whatwg.org/#c0-control-percent-encode-set) and U+0020 SPACE, U+0022 ("), U+0023 (#), U+003C (<), and U+003E (>).

oguzkocer: As shown in the screenshot above, I think we are doing it correctly.

wp_api/src/comments.rs
oguzkocer: @crazytonyli I am working on this [as a separate PR](https://github.com/Automattic/wordpress-rs/pull/440). Do you mind if we land this PR as is?

I am building the retrieve `/comments` and other comment actions on top of these, so it'd be great to have this merged to make progress with the PR reviews.

crazytonyli (approved):

PR #425, opened 2024-12-04
Add `Client` for sending requests to plugin directory

crazytonyli:
> [!Note]
> This PR is built on top of https://github.com/Automattic/wordpress-rs/pull/424.

This PR adds an API client type for sending requests to api.wordpress.org. The new `Client` and other related types are exported as uniffi types. But they are not included in the native uniffi wrapper yet, which I will do in a separate PR.


crazytonyli:
The API client will be moved to wp_api

oguzkocer:
@crazytonyli I think this one is outdated now and can be closed?

PR #429, opened 2024-12-08
Refactor generating Swift binding files

crazytonyli:
It's more technically correct to generate uniffi binding files with Rust library built for each platform. Because the Rust crate that exposes API via uniffi may have different APIs on different platforms (for example, it could have some API for iOS but not for tvOS), and uniffi-rs itself could have different implementations for each platform (for example, there could be subtle differences in the generated code for certain CPU arch).

However, in the reality of this project and the current state of uniffi-rs, there is no difference in the generated bindings on each Apple platform.

This PR updates the build pipeline to use a more "technically correct" approach, even though the result is no different than what we currently have.

### Details

At the moment, `make bindings` builds the wp_api library for macOS(*1), generates bindings using the built library, and uses them on all Apple platforms.

This PR removes the `make bindings` and uses a shell script to generate bindings. Please note the `cargo build` run in the step marked as *1 is skipped because it's unnecessary. We already built libraries for each Apple platform. We should use them to generate bindings.

This PR does run the `wp_uniffi_bindgen` multiple times, one for each platform. Their result will be used in the creating xcframework command: `xcodebuild -create-xcframework -library library-for-ios -headers headers-for-ios --library library-for-macos -headers headers-for-macos ...`. Again, even though we are now passing different header directories, they don't really make any practical difference from what we have right now because the header directories' content is exactly the same.

### DIFF
--- a/xcframework/src/main.rs
+++ b/xcframework/src/main.rs
@@ -6,7 +6,6 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 const XCFRAMEWORK_OUTPUT_PATH: &str = "target/libwordpressFFI.xcframework";
-const SWIFT_BINDINGS_HEADER_DIR: &str = "target/swift-bindings/headers";
 const LIBRARY_FILENAME: &str = "libwordpress.a";
 
 fn main() -> Result<()> {
@@ -50,7 +49,6 @@ impl CreateXCFramework {
 // work together to make it easier to create a xcframework.
 struct XCFramework {
     libraries: Vec<LibraryGroup>,
-    headers: PathBuf,
 }
 
 // Represent a group of static libraries that are built for the same platform.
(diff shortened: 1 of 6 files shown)


xcframework/src/main.rs
```diff
@@ -208,6 +197,18 @@ impl LibraryGroup {
 
         Ok(dest)
     }
+
+    fn headers_dir(&self) -> Result<PathBuf> {
+        let slice = self
+            .slices
+            .first()
```
crazytonyli, line 204: Use the first one here, because bindings in each slice (a.k.a target triple in cargo, `x86_64-apple-darwin`, `aarch64-apple-darwin`, etc.) should have the exact same content.

We could add a check here to ensure they are the same. Maybe it's an overkill? But I'm open to that.

jkmassel (approved):

PR #440, opened 2024-12-11
Use `HashMap<UserAvatarSize, WpResponseString>` for user avatar urls

oguzkocer:


wp_api/src/users.rs
```diff
@@ -131,6 +131,21 @@ impl Display for WpApiParamUsersHasPublishedPosts {
...
+    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, uniffi::Enum,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum UserAvatarSize {
+    #[serde(rename = "24")]
+    Size24,
+    #[serde(rename = "48")]
+    Size48,
+    #[serde(rename = "96")]
+    Size96,
```
jkmassel, line 144: Are these the only out-of-the-box sizes?

jkmassel (approved):
LGTM pending one question

wp_api/src/users.rs
oguzkocer: I _think_ so: https://github.com/WordPress/WordPress/blob/caaf759d21dabffa84ab1e47d1a40798f2693fa0/wp-includes/rest-api.php#L1279-L1299

wp_api/src/lib.rs
```diff
@@ -137,6 +137,27 @@ pub enum JsonValue {
...
 
+uniffi::custom_newtype!(WpResponseString, Option<String>);
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(from = "BoolOrString")]
+pub struct WpResponseString(pub Option<String>);
+
+impl From<BoolOrString> for WpResponseString {
+    fn from(value: BoolOrString) -> Self {
+        match value {
+            BoolOrString::Bool(_) => Self(None),
```
crazytonyli: Return error if a `true` value is returned?

crazytonyli (approved):

wp_api/src/lib.rs
oguzkocer: We could do that, but if this happens, it'll be due to a bug in the API and we probably don't need to propagate that. 🤔

oguzkocer: What do you think @crazytonyli @jkmassel?

crazytonyli: Yes it'd be a bug in the API. It'd be an API bug too if we got `"avatar_urls": ["...", "..."]`, which would cause a parsing error. I think for consistency, we should throw a parsing error here, too, when receiving a `true` value, because it's an unexpected value, in the same way a list is an unexpected value for `avatar_urls`.

I don't mind it too much, though. That's why I approved the PR...

oguzkocer: I could be wrong, but my thinking is that the issue this PR fixes, an object returning as `[]` etc are all expected behavior. I agree with you that they should be considered bugs, but they are not. So, we have to handle them - otherwise we can't use the endpoints at all.

On the other hand, returning `true` when a `String` is expected doesn't make any sense. `false` does in a weird way - although it really should be `null`. That's why I thought it's better not to cause any client side errors for this.

I don't feel strongly about it, so happy to go with whichever.

oguzkocer: @crazytonyli I wanted to get this merged, so I implemented the `try_from` approach as well in 395c8e25377e65fd49480c36fb05fc441a4e1256. Let me know what you think!

wp_api/src/lib.rs
```diff
@@ -137,6 +137,54 @@ pub enum JsonValue {
...
+//
+// During this conversion, there may be some values that are not clear how they should be mapped.
+// For example, when we are expecting a `String` field, if we get a `false` value, we can assume
+// that it's `null`, but there isn't a clear conversion for the `true` value, so we return an
+// error.
+#[derive(Debug, PartialEq, Eq, thiserror::Error, uniffi::Error)]
+pub enum WpApiNewtypeParsingError {
+    #[error("Expecting a `String` value for this field, but received the boolean `true` instead")]
+    BooleanTrueIsReturnedWhenStringIsExpected,
+}
```
crazytonyli, line 160: I'm not sure if a dedicated error type is necessary. For reference, [I used serde's error in the api.wordpress.org implementation](https://github.com/Automattic/wordpress-rs/blob/a452263910aed900276b6e44e2f30f1c433a4e14/wp_api/src/wordpress_org/de.rs#L33-L36).

oguzkocer: If we are going to throw an error, I like that being one of our own to differentiate from the parsing errors we get from `serde`.

I saw that in your PR, but didn't want to say anything yet, because I figured we'll be working on this a bit more and this error handling might not even remain like this 🤷‍♂️

crazytonyli: The type would be lost, right? Because serde would convert the error instance to one of the cases in [`serde::de::Error`](https://rust-lang.github.io/hashbrown/serde/de/trait.Error.html), maybe (`custom`)?

I don't mind either way. My approval still stands. Please don't let this comment be the blocker for merging this PR.

oguzkocer: Yes, when it's used to parse by serde, but the error type is still there when the `TryFrom` is directly used. Again, these are mostly personal preferences to be honest, I don't think there is a right answer 🤷 

In that case, I'll go ahead and merge to keep things moving.

PR #441, opened 2024-12-11
Move wordpress_org_api code into wp_api

crazytonyli:
What says in the title.


oguzkocer (approved):
I think this is a good first step. We might consider moving `wordpress_org/de` module into `wp_serde_helper`, but before we do that, we might want to settle the de-serialization of the weird API responses, such as a `String` field being returned as a `bool`. For example, https://github.com/Automattic/wordpress-rs/pull/440 tries a different approach to this.

This is mainly because I am still a little concerned about the catch all nature of the `deserialize_default_values` implementation. On one side, it makes it easy to drop in everywhere, but on the other hand, that fact might end up doing something unexpected in an edge case. It also doesn't work as is for what https://github.com/Automattic/wordpress-rs/pull/440 is trying to handle and would require additional logic added to `AlternativeValues`.

In any case, this is currently a internal module for `wordpress_org`, so it should be safe to land for now.

PR #445, opened 2024-12-13
Add WordPressOrgAPIClient to list plugins and get a specific plugin

crazytonyli:
Since there are only a few API request functions (get plugin info, browse plugins, search plugins) we are going to implement, I chose to build HTTP requests on each function instead of having some generic request builder.

The response parsing part is even simpler: we expect a 200 status code and JSON body response. Anything else is an error.

### DIFF
--- a/wp_api/src/wordpress_org/client.rs
+++ b/wp_api/src/wordpress_org/client.rs
@@ -24,16 +24,37 @@ impl WordPressOrgApiClient {
         &self,
         slug: &str,
     ) -> Result<PluginInformation, WordPressOrgApiClientError> {
-        self.execute(self.plugin_information_request(slug)).await
+        self.execute(Self::plugin_information_request(slug)).await
     }
 
-    pub fn browse_plugins_request(
+    pub async fn browse_plugins(
         &self,
         category: Option<WordPressOrgApiPluginDirectoryCategory>,
         page: u64,
         page_size: u64,
+    ) -> Result<QueryPluginResponse, WordPressOrgApiClientError> {
+        let request = Self::browse_plugins_request(category, page, page_size);
+        self.execute(request).await
+    }
+
+    pub async fn search_plugins(
+        &self,
+        search: String,
+        page: u64,
+        page_size: u64,
+    ) -> Result<QueryPluginResponse, WordPressOrgApiClientError> {
+        let request = Self::search_plugins_request(search, page, page_size);
+        self.execute(request).await
+    }
+}
+
+impl WordPressOrgApiClient {
+    pub fn browse_plugins_request(
+        category: Option<WordPressOrgApiPluginDirectoryCategory>,
+        page: u64,
+        page_size: u64,
     ) -> WpNetworkRequest {
-        self.query_plugins_request(page, page_size, |url| match category {
+        Self::query_plugins_request(page, page_size, |url| match category {
             Some(category) => {
                 let mut url = url;
                 url.query_pairs_mut()
@@ -44,42 +65,15 @@ impl WordPressOrgApiClient {
         })
     }
 
-    pub async fn browse_plugins(
-        &self,
-        category: Option<WordPressOrgApiPluginDirectoryCategory>,
-        page: u64,
-        page_size: u64,
-    ) -> Result<QueryPluginResponse, WordPressOrgApiClientError> {
-        let request = self.browse_plugins_request(category, page, page_size);
-        self.execute(request).await
-    }
-
-    pub fn search_plugins_request(
-        &self,
-        search: String,
-        page: u64,
-        page_size: u64,
-    ) -> WpNetworkRequest {
-        self.query_plugins_request(page, page_size, |url| {
+    pub fn search_plugins_request(search: String, page: u64, page_size: u64) -> WpNetworkRequest {
+        Self::query_plugins_request(page, page_size, |url| {
             let mut url = url;
             url.query_pairs_mut().append_pair("search", &search);
             url
         })
     }
 
-    pub async fn search_plugins(
-        &self,
-        search: String,
-        page: u64,
-        page_size: u64,
-    ) -> Result<QueryPluginResponse, WordPressOrgApiClientError> {
-        let request = self.search_plugins_request(search, page, page_size);
-        self.execute(request).await
-    }
-}
-
-impl WordPressOrgApiClient {
-    fn plugin_information_request(&self, slug: &str) -> WpNetworkRequest {
+    fn plugin_information_request(slug: &str) -> WpNetworkRequest {
         let mut url = Self::plugin_info_api_url();
         url.query_pairs_mut()
             .append_pair("action", "plugin_information")
@@ -88,12 +82,7 @@ impl WordPressOrgApiClient {
         WpNetworkRequest::get(WpEndpointUrl(url.to_string()))
     }
 
-    fn query_plugins_request<F>(
-        &self,
-        page: u64,
-        page_size: u64,
-        url_builder: F,
-    ) -> WpNetworkRequest
+    fn query_plugins_request<F>(page: u64, page_size: u64, url_builder: F) -> WpNetworkRequest
     where
         F: FnOnce(Url) -> Url,
     {
@@ -106,18 +95,16 @@ impl WordPressOrgApiClient {
         WpNetworkRequest::get(WpEndpointUrl(url.to_string()))
     }
 
+    fn plugin_info_api_url() -> Url {
+        Url::parse("https://api.wordpress.org/plugins/info/1.2/").expect("The URL is valid")
+    }
+
     async fn execute<T>(&self, request: WpNetworkRequest) -> Result<T, WordPressOrgApiClientError>
     where
         T: DeserializeOwned,
     {
         let response = self.request_executor.execute(Arc::new(request)).await?;
         Self::parse(response)
-    }
-}
-
-impl WordPressOrgApiClient {
-    fn plugin_info_api_url() -> Url {
-        Url::parse("https://api.wordpress.org/plugins/info/1.2/").expect("The URL is valid")
     }
 
     fn parse<T>(response: WpNetworkResponse) -> Result<T, WordPressOrgApiClientError>
@@ -195,52 +182,17 @@ impl From<RequestExecutionError> for WordPressOrgApiClientError {
 
 #[cfg(test)]
 mod tests {
-
     use super::*;
-    use crate::{
-        request::{endpoint::media_endpoint::MediaUploadRequest, RequestExecutor},
-        MediaUploadRequestExecutionError,
-    };
-    use std::sync::Arc;
-
-    #[derive(Debug)]
-    struct MockRequestExecutor;
-
-    #[async_trait::async_trait]
-    impl RequestExecutor for MockRequestExecutor {
-        async fn execute(
-            &self,
-            _wp_request: Arc<WpNetworkRequest>,
-        ) -> std::result::Result<WpNetworkResponse, RequestExecutionError> {
-            Err(RequestExecutionError::RequestExecutionFailed {
-                status_code: None,
-                reason: "Mocked request executor".to_string(),
-            })
-        }
-
-        async fn upload_media(
-            &self,
-            media_upload_request: Arc<MediaUploadRequest>,
-        ) -> std::result::Result<WpNetworkResponse, MediaUploadRequestExecutionError> {
-            unimplemented!(
-                "upload_media is not implemented for sending requests to api.wordpress.org"
-            )
-        }
-    }
 
     #[test]
     fn test_plugin_info_requests_include_icons() {
-        let request_executor = Arc::new(MockRequestExecutor {});
-        let client = WordPressOrgApiClient::new(request_executor.clone());
-        let request = client.plugin_information_request("akismet");
+        let request = WordPressOrgApiClient::plugin_information_request("akismet");
         assert!(request.url.0.contains("fields=icons"));
     }
 
     #[test]
     fn test_search_does_not_include_pagination() {
-        let request_executor = Arc::new(MockRequestExecutor {});
-        let client = WordPressOrgApiClient::new(request_executor.clone());
-        let request = client.search_plugins_request("akismet".to_string(), 3, 24);
+        let request = WordPressOrgApiClient::search_plugins_request("akismet".to_string(), 3, 24);
 
         // The 'request[x]' parameters do not work for the search endpoint.
         // The 'page' and 'per_page' parameters do.
--- a/wp_api/src/wordpress_org/client.rs
+++ b/wp_api/src/wordpress_org/client.rs
@@ -33,12 +33,55 @@ impl WordPressOrgApiClient {
         Self::parse(response)
     }
 
-    pub async fn list_plugins(&self, page: u64, page_size: u64) -> Result<QueryPluginResponse> {
+    pub async fn browse_plugins(
+        &self,
+        category: Option<WordPressOrgApiPluginDirectoryCategory>,
+        page: u64,
+        page_size: u64,
+    ) -> Result<QueryPluginResponse> {
+        self.query_plugins(page, page_size, |url| match category {
+            Some(category) => {
+                let mut url = url;
+                url.query_pairs_mut()
+                    .append_pair("browse", category.as_str());
+                url
+            }
+            None => url,
+        })
+        .await
+    }
+
+    pub async fn search_plugins(
+        &self,
+        search: String,
+        page: u64,
+        page_size: u64,
+    ) -> Result<QueryPluginResponse> {
+        self.query_plugins(page, page_size, |url| {
+            let mut url = url;
+            url.query_pairs_mut().append_pair("search", &search);
+            url
+        })
+        .await
+    }
+}
+
+impl WordPressOrgApiClient {
+    async fn query_plugins<F>(
+        &self,
+        page: u64,
+        page_size: u64,
+        url_builder: F,
+    ) -> Result<QueryPluginResponse>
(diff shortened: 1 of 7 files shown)


wp_api/src/wordpress_org/client.rs
```diff
@@ -0,0 +1,249 @@
...
+    request::{endpoint::WpEndpointUrl, RequestExecutor, WpNetworkRequest, WpNetworkResponse},
+    RequestExecutionError,
+};
+use serde::de::DeserializeOwned;
+use std::sync::Arc;
+use url::Url;
+
+use super::plugin_directory::{PluginInformation, QueryPluginResponse};
+
+pub type Result<T> = std::result::Result<T, self::WordPressOrgApiClientError>;
```
oguzkocer: I prefer to have the error type in the function definition. I know this is something a lot of Rust libraries do, but most of them have a single error type, so it's somewhat understandable. However, in this case, the error type is not clear unless we navigate to it - either from source code or from the code.

I feel like that extra step is unnecessary as the new type doesn't add that much value here.

Having said that, if you feel strongly about having the new type, we can keep it as is.

wp_api/src/wordpress_org/client.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    use super::*;
+    use crate::{
+        request::{endpoint::media_endpoint::MediaUploadRequest, RequestExecutor},
+        MediaUploadRequestExecutionError,
+    };
+    use std::sync::Arc;
+
+    #[derive(Debug)]
+    struct MockRequestExecutor {
```
oguzkocer: In my opinion, dealing with the request executor in the unit test level is not necessary. The added complexity doesn't bring much to the table when all we really need to test is the url building. I would also prefer if we don't add `tokio` as a dependency if we can avoid it.

If the url building happens on its own, then it'd be easy to unit test that and I'd really prefer if we did that.

It's worth noting that we are trying to design the API to work without `async`, which is why we have the extra step of building a `WpNetworkRequest` from a request builder in all other clients. This not only avoids the [function coloring issue](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/), but we might end up using the manual request building ourselves to provide a more complex solution. An example of this happened recently when @jkmassel tackled pagination in Swift.

If we did add manual request building, and I think we should, then we could also validate the `WpNetworkRequest` instances as the request execution part is always the same across all clients. This would be an alternative to testing the urls as mentioned above - not in addition to it. I don't think we need to do both.

wp_api_integration_tests/tests/test_plugin_directory.rs
```diff
@@ -1,46 +1,95 @@
...
+    ) -> Result<WpNetworkResponse, MediaUploadRequestExecutionError> {
+        unimplemented!("upload_media is not implemented for sending requests to api.wordpress.org")
+    }
+}
+
+fn wordpress_org_api_client() -> WordPressOrgApiClient {
+    let client = ReqwestClient::new();
+    let executor = ReqwestExecutor { client };
+    WordPressOrgApiClient::new(Arc::new(executor))
 }
```
oguzkocer:
```suggestion
fn wordpress_org_api_client() -> WordPressOrgApiClient {
    WordPressOrgApiClient::new(Arc::new(
        wp_api_integration_tests::AsyncWpNetworking::default(),
    ))
}
```

I don't think we need a new executor. You can remove everything  related to `ReqwestExecutor` and replace the `wordpress_org_api_client` function as above and it'll work as expected.

wp_api_integration_tests/tests/test_plugin_directory.rs
```diff
@@ -101,3 +149,34 @@ async fn test_parsing_full_plugin_directory() {
...
+#[rstest::rstest]
+#[case(WordPressOrgApiPluginDirectoryCategory::New)]
+#[case(WordPressOrgApiPluginDirectoryCategory::Popular)]
+#[case(WordPressOrgApiPluginDirectoryCategory::Updated)]
+#[case(WordPressOrgApiPluginDirectoryCategory::TopRated)]
+async fn test_browse_plugins(#[case] category: WordPressOrgApiPluginDirectoryCategory) {
+    let client = wordpress_org_api_client();
+    let response = client.browse_plugins(Some(category), 1, 30).await;
+    assert!(response.is_ok());
+    assert!(!response.unwrap().plugins.is_empty());
```
oguzkocer:
```suggestion
    use wp_api_integration_tests::AssertResponse;
    let response = wordpress_org_api_client()
        .browse_plugins(Some(category), 1, 30)
        .await
        .assert_response();
    assert!(!response.plugins.is_empty());
```

wp_api_integration_tests/tests/test_plugin_directory.rs
```diff
@@ -101,3 +149,34 @@ async fn test_parsing_full_plugin_directory() {
...
+    assert!(response.is_ok());
+
+    let slugs = response
+        .unwrap()
+        .plugins
+        .into_iter()
+        .map(|plugin| plugin.slug)
+        .collect::<Vec<_>>();
+    println!("{:?}", slugs);
+    assert!(slugs.contains(&"jetpack-social".to_string()));
```
oguzkocer:
```suggestion
    use wp_api_integration_tests::AssertResponse;
    let plugins = wordpress_org_api_client()
        .search_plugins("jetpack-social".to_string(), 1, 30)
        .await
        .assert_response()
        .plugins;
    assert!(
        plugins.iter().any(|p| p.slug == "jetpack-social"),
        "Plugins search result doesn't contain 'jetpack-social': {:#?}",
        plugins
    );
```

oguzkocer (changes requested):
@crazytonyli I'm done with my initial review and left some suggestions. It looks like there is also a binding generation issue on the Kotlin side, but I haven't investigated it. Let me know if you need any help with that.

wp_api/src/wordpress_org/client.rs
crazytonyli: I copied this code from my original implementation in the separate wordpress_org_api crate. It's deleted from this PR now.

wp_api/src/wordpress_org/client.rs
crazytonyli: > In my opinion, dealing with the request executor in the unit test level is not necessary. The added complexity doesn't bring much to the table when all we really need to test is the url building.

The two tests I added are not so much about URL/request building, though. Those tests are about ensuring `fields=icons` is included in API requests, and the right pagination parameters are used (i.e., `page` instead of `request[page]`).

Also, I want to write tests that check application behaviors rather than their implementation details. By "application behaviors", I mean assertions like a GET request whose URL includes query `search=xxx` is sent out when the `search_plugin()` function is called. By "implementation details", I mean assertions like `PluginSearchRequest(search: "xxx").build_request().url.contains("search=xxx")`. I'm not saying testing implementation details is not important, though. It's just not what the tests in this PR are about.

> If the url building happens on its own, then it'd be easy to unit test that and I'd really prefer if we did that.

Sure. I can look into that.

crazytonyli: > I would also prefer if we don't add tokio as a dependency if we can avoid it.

Can you talk a bit more about this?

oguzkocer: > Also, I want to write tests that check application behaviors rather than their implementation details.

Wouldn't these be integration tests though?

---

I am not sure I understand the benefit of dealing with a mock executor when we can check the details of the `WpNetworkRequest` that will be executed. Although I don't think even _that_ is necessary because testing the `url` should be more than enough. Since it's a GET request, everything about page, fields, search etc will be part of the url query pairs.

---

> Can you talk a bit more about this?

We want to remain async executor agnostic within `wp_api`, as we don't want to ship an executor as part of the library. This was something @jkmassel was specifically worried about when we first worked on the prototype.

Adding `tokio` as a dev dependency does not break that, so it's not an issue. However, I feel like if it's there, it's just one step away from making it as a regular dependency - which when I say it out loud, I didn't find myself convincing 😅 I guess in this case, it's easy to not require it, so I guess I'd rather not - until we really need it.

wp_api/src/wordpress_org/client.rs
oguzkocer: I see, sorry for missing it in the original review. Also, as I mentioned in my comment, there is some personal preference involved, but hopefully I was able to explain why I prefer this.

With these types of comments, if you don't find my argument convincing, please feel free to say so and keep it as is.

wp_api/src/wordpress_org/client.rs
crazytonyli: > Wouldn't these be integration tests though?

No, I would not consider them as integration tests. I thought it's pretty common to use test doubles to take dependencies (HTTP client in this particular case) out of the test env in unit tests.

oguzkocer: Although that's definitely common, it's generally because that's the only way to test something - mainly because of poor design either in the language/framework, or the implementation itself.

Otherwise, unit testing requires testing the smallest unit possible - hence the name _unit_ testing. What you were doing in the tests - note that I haven't re-reviewed, was use a mock executor just to get the url information out of it and then check the details of it. Doesn't that sound unnecessary when you could just test the url itself?

Testing the smallest possible unit not only makes it more useful, but it's also a lot easier to understand what the test is trying to achieve. For example, in the implementation, you had an assertion to check only one request was being made - and you also implemented the logic for tracking how many requests are made within a mock executor. If I didn't have any context about the code base, looking at that test I'd think that it's possible and worrisome that there might be multiple requests. That's not a possibility, but even if it was, it'd be the wrong place to test it, simply because we'd have to do it for every single endpoint we implement.

wp_api/src/wordpress_org/client.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    pub async fn plugin_information(
+        &self,
+        slug: &str,
+    ) -> Result<PluginInformation, WordPressOrgApiClientError> {
+        self.execute(self.plugin_information_request(slug)).await
+    }
+
+    pub fn browse_plugins_request(
+        &self,
```
oguzkocer, line 31: We don't need to take `&self` for building requests. Once it's removed, we also won't require a mock executor in the unit tests.

Uniffi won't allow us to have associated functions, so they'd need to be moved to the unexported `impl WordPressOrgApiClient` block which we have 2 of at the moment - and I think we should stick to just one in this case. This means native wrappers won't be able to access the build request functions, but if that ends up being necessary, we can extract them into a `RequestBuilder` type which is why we have that abstraction for all other clients. For now, we can skip this step.

This suggestion requires moving a few pieces around which I already did to make sure it all works. If you like, you can apply the following diff:

```
diff --git a/wp_api/src/wordpress_org/client.rs b/wp_api/src/wordpress_org/client.rs
index fc8f419..1796bc9 100644
--- a/wp_api/src/wordpress_org/client.rs
+++ b/wp_api/src/wordpress_org/client.rs
@@ -24,62 +24,56 @@ impl WordPressOrgApiClient {
         &self,
         slug: &str,
     ) -> Result<PluginInformation, WordPressOrgApiClientError> {
-        self.execute(self.plugin_information_request(slug)).await
+        self.execute(Self::plugin_information_request(slug)).await
     }
 
-    pub fn browse_plugins_request(
+    pub async fn browse_plugins(
         &self,
         category: Option<WordPressOrgApiPluginDirectoryCategory>,
         page: u64,
         page_size: u64,
-    ) -> WpNetworkRequest {
-        self.query_plugins_request(page, page_size, |url| match category {
-            Some(category) => {
-                let mut url = url;
-                url.query_pairs_mut()
-                    .append_pair("browse", category.as_str());
-                url
-            }
-            None => url,
-        })
+    ) -> Result<QueryPluginResponse, WordPressOrgApiClientError> {
+        let request = Self::browse_plugins_request(category, page, page_size);
+        self.execute(request).await
     }
 
-    pub async fn browse_plugins(
+    pub async fn search_plugins(
         &self,
-        category: Option<WordPressOrgApiPluginDirectoryCategory>,
+        search: String,
         page: u64,
         page_size: u64,
     ) -> Result<QueryPluginResponse, WordPressOrgApiClientError> {
-        let request = self.browse_plugins_request(category, page, page_size);
+        let request = Self::search_plugins_request(search, page, page_size);
         self.execute(request).await
     }
+}
 
-    pub fn search_plugins_request(
-        &self,
-        search: String,
+impl WordPressOrgApiClient {
+    pub fn browse_plugins_request(
+        category: Option<WordPressOrgApiPluginDirectoryCategory>,
         page: u64,
         page_size: u64,
     ) -> WpNetworkRequest {
-        self.query_plugins_request(page, page_size, |url| {
+        Self::query_plugins_request(page, page_size, |url| match category {
+            Some(category) => {
+                let mut url = url;
+                url.query_pairs_mut()
+                    .append_pair("browse", category.as_str());
+                url
+            }
+            None => url,
+        })
+    }
+
+    pub fn search_plugins_request(search: String, page: u64, page_size: u64) -> WpNetworkRequest {
+        Self::query_plugins_request(page, page_size, |url| {
             let mut url = url;
             url.query_pairs_mut().append_pair("search", &search);
             url
         })
     }
 
-    pub async fn search_plugins(
-        &self,
-        search: String,
-        page: u64,
-        page_size: u64,
-    ) -> Result<QueryPluginResponse, WordPressOrgApiClientError> {
-        let request = self.search_plugins_request(search, page, page_size);
-        self.execute(request).await
-    }
-}
-
-impl WordPressOrgApiClient {
-    fn plugin_information_request(&self, slug: &str) -> WpNetworkRequest {
+    fn plugin_information_request(slug: &str) -> WpNetworkRequest {
         let mut url = Self::plugin_info_api_url();
         url.query_pairs_mut()
             .append_pair("action", "plugin_information")
@@ -88,12 +82,7 @@ impl WordPressOrgApiClient {
         WpNetworkRequest::get(WpEndpointUrl(url.to_string()))
     }
 
-    fn query_plugins_request<F>(
-        &self,
-        page: u64,
-        page_size: u64,
-        url_builder: F,
-    ) -> WpNetworkRequest
+    fn query_plugins_request<F>(page: u64, page_size: u64, url_builder: F) -> WpNetworkRequest
     where
         F: FnOnce(Url) -> Url,
     {
@@ -106,6 +95,10 @@ impl WordPressOrgApiClient {
         WpNetworkRequest::get(WpEndpointUrl(url.to_string()))
     }
 
+    fn plugin_info_api_url() -> Url {
+        Url::parse("https://api.wordpress.org/plugins/info/1.2/").expect("The URL is valid")
+    }
+
     async fn execute<T>(&self, request: WpNetworkRequest) -> Result<T, WordPressOrgApiClientError>
     where
         T: DeserializeOwned,
@@ -113,12 +106,6 @@ impl WordPressOrgApiClient {
         let response = self.request_executor.execute(Arc::new(request)).await?;
         Self::parse(response)
     }
-}
-
-impl WordPressOrgApiClient {
-    fn plugin_info_api_url() -> Url {
-        Url::parse("https://api.wordpress.org/plugins/info/1.2/").expect("The URL is valid")
-    }
 
     fn parse<T>(response: WpNetworkResponse) -> Result<T, WordPressOrgApiClientError>
     where
@@ -195,52 +182,17 @@ impl From<RequestExecutionError> for WordPressOrgApiClientError {
 
 #[cfg(test)]
 mod tests {
-
     use super::*;
-    use crate::{
-        request::{endpoint::media_endpoint::MediaUploadRequest, RequestExecutor},
-        MediaUploadRequestExecutionError,
-    };
-    use std::sync::Arc;
-
-    #[derive(Debug)]
-    struct MockRequestExecutor;
-
-    #[async_trait::async_trait]
-    impl RequestExecutor for MockRequestExecutor {
-        async fn execute(
-            &self,
-            _wp_request: Arc<WpNetworkRequest>,
-        ) -> std::result::Result<WpNetworkResponse, RequestExecutionError> {
-            Err(RequestExecutionError::RequestExecutionFailed {
-                status_code: None,
-                reason: "Mocked request executor".to_string(),
-            })
-        }
-
-        async fn upload_media(
-            &self,
-            media_upload_request: Arc<MediaUploadRequest>,
-        ) -> std::result::Result<WpNetworkResponse, MediaUploadRequestExecutionError> {
-            unimplemented!(
-                "upload_media is not implemented for sending requests to api.wordpress.org"
-            )
-        }
-    }
 
     #[test]
     fn test_plugin_info_requests_include_icons() {
-        let request_executor = Arc::new(MockRequestExecutor {});
-        let client = WordPressOrgApiClient::new(request_executor.clone());
-        let request = client.plugin_information_request("akismet");
+        let request = WordPressOrgApiClient::plugin_information_request("akismet");
         assert!(request.url.0.contains("fields=icons"));
     }
 
     #[test]
     fn test_search_does_not_include_pagination() {
-        let request_executor = Arc::new(MockRequestExecutor {});
-        let client = WordPressOrgApiClient::new(request_executor.clone());
-        let request = client.search_plugins_request("akismet".to_string(), 3, 24);
+        let request = WordPressOrgApiClient::search_plugins_request("akismet".to_string(), 3, 24);
 
         // The 'request[x]' parameters do not work for the search endpoint.
         // The 'page' and 'per_page' parameters do.
```

oguzkocer (changes requested):
@crazytonyli I just have one suggestion for removing the `&self` parameter from the building request functions which should simplify the code quite a bit. Otherwise, I think we are good to go. Thanks for addressing my feedback 🙇‍♂️

wp_api/src/wordpress_org/client.rs
crazytonyli: ⬆️ Diff is taken in https://github.com/Automattic/wordpress-rs/pull/445/commits/d2ebf6e0aa75278b2e43c0ff73da0cc5b8ebe310

oguzkocer (approved):

PR #452, opened 2024-12-18
Add `/comments` related errors and their integration tests

oguzkocer:
Follows established patterns to implement error types for `/comments` and their integration tests. It also introduces `CommentCreateParamsBuilder` type to make it easier to build a `CommentCreateParams` type in Rust.

There is some [discussion](https://internals.rust-lang.org/t/default-for-a-subset-of-fields/20013/22) and pre-RFCs for using something like the `..Default::default()` syntax for all the remaining fields when the type itself doesn't implement `Default`, but the remaining fields do. If such a feature is ever implemented, we could drop these builders.

There are also proc macro crates that implement these, but the ones I tried before didn't work out the way I'd like. For example, most `build` functions will return a `Result<T, E>`, but that only makes sense when the build might fail. In this builder type, I make the mandatory fields for the params type also mandatory for the builder, so we don't require any error handling.

In any case, I think this setup will serve us well on the Rust side for now. In native wrappers, this problem doesn't exist because we can use the `#[uniffi(default = None)]` macro attribute and both Kotlin and Swift can take arbitrary number of arguments in their constructor.

### DIFF
--- a/wp_api/src/api_error.rs
+++ b/wp_api/src/api_error.rs
@@ -143,16 +143,52 @@ pub enum WpErrorCode {
     CannotManagePlugins,
     #[serde(rename = "rest_cannot_read_application_password")]
     CannotReadApplicationPassword,
+    #[serde(rename = "rest_cannot_read")]
+    CannotRead,
+    #[serde(rename = "rest_cannot_read_post")]
+    CannotReadPost,
     #[serde(rename = "rest_cannot_view")]
     CannotView,
     #[serde(rename = "rest_cannot_view_plugin")]
     CannotViewPlugin,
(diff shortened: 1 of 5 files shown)


crazytonyli (approved):

crazytonyli:
@oguzkocer This PR's integration tests failed, and it caused the trunk branch and a few other PRs to fail.

oguzkocer:
@crazytonyli Thank you for letting me know. I added the `buildkite/wordpress-rs/wordpress-rust-wordpress-6-dot-6` task to the required checks, so that this won't happen again. I'll open a PR fixing the issue.

PR #453, opened 2024-12-19
Speed up nightly job by running in a different agent

crazytonyli:


oguzkocer (commented):
I don't think we care how long this job takes as it's never a blocker. So, it makes sense to focus on the cost and availability. Since I am not sure how busy the `tumblr-metal` agent is and what the cost looks like, I'll leave this decision to @jkmassel.

oguzkocer:
@crazytonyli Since #530 significantly sped up the tests, do you think this PR is still relevant? I don't have a strong opinion on it either way, just wanted to check in with you to avoid the PR going stale.

crazytonyli:
Nope. We don't need to change the agent now.

PR #454, opened 2024-12-20
Add `WpDateTime` to parse UTC/GMT time

crazytonyli:
Relate to #262.

The new `WpDateTime` type is used on posts. If the solution looks good, I can make further changes to use it on other endpoints.

### DIFF
--- a/wp_api/src/date.rs
+++ b/wp_api/src/date.rs
@@ -1,0 +1,111 @@
+use std::fmt::Display;
+use std::str::FromStr;
+use std::sync::Arc;
+
+use chrono::DateTime;
+use chrono::NaiveDateTime;
+use chrono::TimeZone;
+use chrono::Utc;
+use serde::Deserializer;
+use serde::Serializer;
+use serde::{Deserialize, Serialize};
+
+pub const REST_API_DATE_FORMAT: &str = "%Y-%m-%dT%H:%M:%S";
+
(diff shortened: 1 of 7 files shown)


wp_api/src/date.rs
```diff
@@ -0,0 +1,111 @@
...
+impl WpDateTime {
+    pub fn date_time(&self) -> &DateTime<Utc> {
+        &self.gmt.date_time
+    }
+}
+
+// This is an intermediate struct to allow the public `WpDateTime` to be exported as a
+// uniffi Record type. This struct itself is exported as a uniffi Object type.
+#[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+struct GMTDateTime {
```
oguzkocer, line 29: I don't think we need both `WpDateTime` & `GMTDateTime` at the same time since one is just a wrapper for the other. We can use one exported `uniffi::Object` with an inner type and that should be enough.

oguzkocer (commented):
I had already worked [on a similar solution in another branch](https://github.com/Automattic/wordpress-rs/commit/09c280bdabc9af95c19438e9a9eb13daee678783#diff-34896c4af5b013f64231ce1a9cb00e556fecceda5880d744173206817d2bb42dL399). I think that solution might be slightly more flexible to cover multiple date formats, but curious to hear what you think.

PR #456, opened 2024-12-20
List & retrieve `/taxonomies` endpoint

oguzkocer:
Follows established patterns to implement [`/taxonomies`](https://developer.wordpress.org/rest-api/reference/taxonomies/) endpoint.

[List `/taxonomies` endpoint](https://developer.wordpress.org/rest-api/reference/taxonomies/#retrieve-a-taxonomy) has the following argument:

> `type` 	Limit results to taxonomies associated with a specific post type.

However, I couldn't get this to work for any post types except for `post`. I believe this corresponds to the `$object_type` in [this source function](https://github.com/WordPress/WordPress/blob/1747403b13a6d20f52082237ab8b8b28d605a4a0/wp-includes/taxonomy.php#L311), but it's not clear to me what other values would be appropriate here.

This might mean that using `PostType` here might be incorrect, but for now I left the other test cases as a `TODO`.

@jkmassel @crazytonyli I'd appreciate your help with this one.

### DIFF
--- a/wp_api/src/post_types.rs
+++ b/wp_api/src/post_types.rs
@@ -1,7 +1,8 @@
+use std::convert::Infallible;
 use std::{collections::HashMap, fmt::Display, str::FromStr};
 
+use crate::impl_as_query_value_from_as_str;
 use crate::url_query::AsQueryValue;
-use crate::{impl_as_query_value_from_as_str, EnumFromStrParsingError};
 use serde::{Deserialize, Serialize};
 use wp_contextual::WpContextual;
 
@@ -43,7 +44,7 @@ impl PostType {
 }
 
 impl FromStr for PostType {
-    type Err = EnumFromStrParsingError;
+    type Err = Infallible;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
--- a/wp_api/src/post_types.rs
+++ b/wp_api/src/post_types.rs
@@ -57,9 +57,7 @@ impl FromStr for PostType {
             "wp_navigation" => Ok(Self::WpNavigation),
             "wp_font_family" => Ok(Self::WpFontFamily),
             "wp_font_face" => Ok(Self::WpFontFace),
-            value => Err(EnumFromStrParsingError::UnknownVariant {
-                value: value.to_string(),
-            }),
+            value => Ok(Self::Custom(value.to_string())),
         }
     }
 }
--- a/wp_api/src/post_types.rs
+++ b/wp_api/src/post_types.rs
@@ -1,5 +1,7 @@
-use std::{collections::HashMap, fmt::Display};
+use std::{collections::HashMap, fmt::Display, str::FromStr};
 
+use crate::url_query::AsQueryValue;
+use crate::{impl_as_query_value_from_as_str, EnumFromStrParsingError};
 use serde::{Deserialize, Serialize};
 use wp_contextual::WpContextual;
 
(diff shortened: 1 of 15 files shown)


wp_api_integration_tests/tests/test_taxonomies_immut.rs
```diff
@@ -0,0 +1,203 @@
...
+        .assert_response();
+}
+
+#[template]
+#[rstest]
+#[case::default(TaxonomyListParams::default())]
+#[case::post(TaxonomyListParams { post_type: Some(PostType::Post) })]
+// TODO: These post types return:
+// ```
+// {"code":"rest_cannot_view","message":"Sorry, you are not allowed to manage terms in this taxonomy.","data":{"status":403}}
```
crazytonyli: I used a jurrasic ninja site (with pre-populated content) for testing. And I only see this "Sorry, you are not allowed to manage terms in this taxonomy." error when using `context=edit` and a type that does not have taxonomies.

For example, `/wp/v2/taxonomies` in my test site returns "Pattern Categories", which has `wp_block` in its `types`, and `/wp/v2/taxonomies?type=wp_block&context=edit` returns a valid response for me. But there is no `wp_template` results in `/wp/v2/taxonomies`, and `wp/v2/taxonomies?type=wp_template&context=edit` returns the "Sorry, you are not allowed to manage terms in this taxonomy" error.

oguzkocer: @crazytonyli Thanks a lot for looking into this. Your explanation clarified how it works and I found that we have `post`, `nav_menu_item` & `wp_block` post types in our test site, so I've added tests for them in d3736919fe7523c171b10e5b731bd6fff8ba111c.

wp_api/src/post_types.rs
```diff
@@ -36,11 +38,40 @@ impl Display for PostType {
...
+            "nav_menu_item" => Ok(Self::NavMenuItem),
+            "wp_block" => Ok(Self::WpBlock),
+            "wp_template" => Ok(Self::WpTemplate),
+            "wp_template_part" => Ok(Self::WpTemplatePart),
+            "wp_navigation" => Ok(Self::WpNavigation),
+            "wp_font_family" => Ok(Self::WpFontFamily),
+            "wp_font_face" => Ok(Self::WpFontFace),
+            value => Err(EnumFromStrParsingError::UnknownVariant {
+                value: value.to_string(),
+            }),
```
crazytonyli: Should this case return `PostType::Custom(value)`? IMHO, it's more ideal if this conversion between concrete type and string works both ways. For example:

```rust
let type: PostType = PostType::Custom("foo")
assert_eq!(type.as_str(), "foo")

let converted: PostType = type.as_str().into()
assert_eq!(type, converted)
```

oguzkocer: It definitely should be, thanks for catching that! Addressed in 0afb76c1ff8486a0d3a10133f45d1c5d01fc08d3.

Hopefully I didn't make the same mistake in the other PRs that build on this 😓

crazytonyli: Now that the parsing does not return error, should we implement `From<&str>` instead of `FromStr`?

oguzkocer: We need to implement `FromStr` because of [this trait](https://github.com/Automattic/wordpress-rs/blob/trunk/wp_api/src/url_query.rs#L108-L110) bound. Some of the types can't be converted from `String`, so we have to be able to throw an error.

For this type, we could implement `From<&str>` and then call that in the `FromStr` implementation, but that's currently not necessary. If it becomes necessary, we can do that for all the available types in a single PR, rather than break the established pattern in this one.

crazytonyli (approved):

wp_api/src/post_types.rs
crazytonyli: Maybe changing the error type to `std::convert::Infallible`?

oguzkocer: TIL!

I've made the change in 42eedcbddda31ef6763e5f736c098892d7adb0cd - including for some other types that would benefit from this.

PR #471, opened 2024-12-27
`/categories` endpoint

oguzkocer:
Follows established patterns to implement the entirety of [`/categories` endpoint.](https://developer.wordpress.org/rest-api/reference/categories/)

`/categories` and `/tags` are very similar, the difference being `/categories` can have a `parent` whereas `/tags` can't. In order to save us time, I implemented the entirety of `/categories` in a single PR. The reviewer can mainly focus on the `parent` related changes and that will probably be enough.

---

I've thought a little about whether we want to combine `/tags` and `/categories` somehow. However, I believe that'll not only make the code more complicated and harder to maintain, it'll also make it more likely we introduce bugs due to the extra complication.

If we need to treat both taxonomy types in a similar manner, we can introduce new traits to establish "is a" relationship. That way we can still avoid duplicating code that includes logic. Whereas the code that's "duplicated" right now is the type information and wouldn't benefit from a DRY approach.

I'd be happy to discuss this further as I'd like to establish this as a pattern going forward where we'll implement other endpoints that are similar to existing ones - such as `/pages` being similar to `/posts`.

### DIFF
--- a/native/kotlin/api/kotlin/src/integrationTest/kotlin/CategoriesEndpointTest.kt
+++ b/native/kotlin/api/kotlin/src/integrationTest/kotlin/CategoriesEndpointTest.kt
@@ -1,0 +1,131 @@
+package rs.wordpress.api.kotlin
+
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+import uniffi.wp_api.PostListParams
+import uniffi.wp_api.SparseCategoryFieldWithEditContext
+import uniffi.wp_api.CategoryCreateParams
+import uniffi.wp_api.CategoryListParams
+import uniffi.wp_api.CategoryUpdateParams
+import uniffi.wp_api.WpErrorCode
+import uniffi.wp_api.wpAuthenticationFromUsernameAndPassword
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+private const val CATEGORY_ID_48: Long = 48
+private const val CATEGORY_ID_59: Long = 59
+
+class CategoriesEndpointTest {
+    private val testCredentials = TestCredentials.INSTANCE
+    private val siteUrl = testCredentials.parsedSiteUrl
+    private val authentication = wpAuthenticationFromUsernameAndPassword(
+        username = testCredentials.adminUsername, password = testCredentials.adminPassword
+    )
+    private val client = WpApiClient(siteUrl, authentication)
+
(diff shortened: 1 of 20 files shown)


native/kotlin/api/kotlin/src/integrationTest/kotlin/CategoriesEndpointTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        assertNull(categoryList.first().description)
+    }
+
+    @Test
+    fun testRetrieveMediaRequest() = runTest {
+        val category = client.request { requestBuilder ->
+            requestBuilder.categories().retrieveWithEditContext(CATEGORY_ID_59)
+        }.assertSuccessAndRetrieveData().data
+        assertNotNull(category)
+    }
```
oguzkocer, line 56: Not sure how I sneaked this one in here. I am removing it in the `/themes` PR I am working on right now.

oguzkocer: Correction: This is a valid test, the only issue is the name of the function which is corrected in the `/themes` PR.

wp_api/src/categories.rs
```diff
@@ -0,0 +1,273 @@
...
+        AppendUrlQueryPairs, AsQueryValue, FromUrlQueryPairs, QueryPairs, QueryPairsExtension,
+        UrlQueryPairsMap,
+    },
+    EnumFromStrParsingError, WpApiParamOrder,
+};
+
+impl_as_query_value_for_new_type!(CategoryId);
+uniffi::custom_newtype!(CategoryId, i64);
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CategoryId(pub i64);
```
crazytonyli, line 21: From a quick look, it appears WordPress uses [one table for all taxonomies](https://codex.wordpress.org/Database_Description#Table:_wp_term_taxonomy). So, maybe we can use a struct `TaxonomyId` for tag, category, etc.

crazytonyli (approved):

wp_api/src/categories.rs
oguzkocer: We can just use a `i64` as well, but the extra types are used to differentiate each type of id and ensure that we always use the correct type for operations on the Rust side. When it's exposed to Kotlin, it's just a typealias, so unfortunately it doesn't make any difference there.

PR #478, opened 2025-01-02
`/search` endpoint

oguzkocer:
Follows established patterns to implement the entirety of [`/search` endpoint.](https://developer.wordpress.org/rest-api/reference/search-results/)

Up until now every endpoint we implemented supported all 3 contexts, so we have been generating functions for all of them. However, `/search` only supports `view` & `embed` contexts. I've considered manually implementing this endpoint instead of using `#[derive(WpDerivedRequest)]`, but I figured there might be other endpoints that has a similar limitation. That's why this PR implements an optional `available_contexts` parameter, where we can provide which contexts we want to support as a comma separated list. Omitting this parameter will result in the previous behavior where all 3 contexts are supported.

### DIFF
--- a/wp_api/src/lib.rs
+++ b/wp_api/src/lib.rs
@@ -27,6 +27,7 @@ pub mod plugins;
 pub mod post_types;
 pub mod posts;
 pub mod request;
+pub mod search_results;
 pub mod site_settings;
 pub mod tags;
 pub mod taxonomies;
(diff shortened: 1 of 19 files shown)


wp_api/src/lib.rs
```diff
@@ -183,20 +184,27 @@ impl TryFrom<BoolOrString> for WpResponseString {
...
     Bool(bool),
     VecString(Vec<String>),
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, uniffi::Enum)]
+#[serde(untagged)]
+pub enum IntegerOrString {
+    Integer(i64),
+    String(String),
+}
```
crazytonyli, line 206: Any reason not using the existing `DeserializeI64OrStringVisitor`?

oguzkocer: `DeserializeI64OrStringVisitor` is used to parse the data into `i64` whereas here we want to be able to support both `i64` and an arbitrary string (not just a number). Furthermore, this type is directly exposed to the clients, it's not an intermediate type that's only used for parsing.

crazytonyli: I wonder if it'd be valuable to introduce a `Either<A, B>` type, which can represent all the `FooOrBar` types above.

oguzkocer: `uniffi` does not support generics, so that doesn't help us in this case.

crazytonyli (approved):

PR #487, opened 2025-01-15
Use strum to replace manual implementation of `as_str` and `FromStr`

crazytonyli:
Here are the two core changes:
1. Use `strum_macros::EnumString` derive macro to replace our own `FromStr` implementation.
2. Use `strum_macros::Display` derive macro to replace our own `as_str` implementation. The macro adds `std::fmt::Display` implementation to the attached enums.

### Breaking change

The enums now no longer have the `as_str` function. Instead, they have `to_string`.

### Do not use `strum_macros::AsRefStr`

strum also has a derive macro `AsRefStr` to add `fn as_ref() -> AsRef<str>` to the attached enums. However, the function does not match the `to_string` implementation generated by `strum_macros::Display` macro. The `Custom(String)` variant returns `"Custom"`, instead of the `String` value in the tuple.

### DIFF
--- a/wp_api/src/lib.rs
+++ b/wp_api/src/lib.rs
@@ -8,7 +8,7 @@ pub use api_error::{
 pub use parsed_url::{ParseUrlError, ParsedUrl};
 use plugins::*;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, str::FromStr};
+use std::collections::HashMap;
 use url_query::AsQueryValue;
 use users::*;
 pub use uuid::{WpUuid, WpUuidParseError};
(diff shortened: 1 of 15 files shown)


wp_api/src/lib.rs
```diff
@@ -218,3 +206,31 @@ macro_rules! generate {
 }
 
 uniffi::setup_scaffolding!();
+
+#[cfg(test)]
+mod lib_tests {
```
crazytonyli: When making this mod as `tests` like we usually do in other source files, running this mod in VS code would run the tests in the whole library. Maybe the "tests" mod at the root level is reserved by cargo or something.

wp_api/src/search_results.rs
```diff
@@ -22,88 +21,54 @@ use wp_contextual::WpContextual;
...
 pub enum SearchResultType {
     #[default]
     #[serde(rename = "post")]
+    #[strum(serialize = "post")]
     Post,
     #[serde(rename = "term")]
+    #[strum(serialize = "term")]
     Term,
     #[serde(rename = "post-format")]
+    #[strum(serialize = "post-format")]
```
oguzkocer, line 36: It looks like we might also need the `to_string` [strum attribute](https://docs.rs/strum/latest/strum/additional_attributes/index.html#attributes-on-variants), at least according to the documentation. We should probably add some unit tests for at least one of the `enum`s that uses `#[strum(serialize = "{value}")]` attribute. We can also use that test to confirm whether we need to `to_string` attribute or not.

wp_api/src/lib.rs
oguzkocer: I run all the commands from the command line, so I am not 100% sure, but I think this is the expected behavior. I assume VS code triggers the tests for the mod you're running and since this is the root mod, everything else falls under it. If my assumption is correct, than running tests in `request` module should also run all the tests under `endpoint` module. I unfortunately can't test this right now because I don't have VS code setup.

oguzkocer: Also, I think I'd prefer if we used the same name for the mod for all unit tests. It makes it easier to run them from the command line as we don't need to remember what the mod name is. It may also effect some partial matches.

```suggestion
mod tests {
```

oguzkocer (changes requested):
@crazytonyli I left one important comment about `#[strum(serialize = "{value}")]`, but otherwise this looks good to me.

wp_api/src/search_results.rs
crazytonyli: I added a unit test in https://github.com/Automattic/wordpress-rs/pull/487/commits/231fa098edb8bf3891cb03e1659638099111aecd. I think `strum_macros::Display` does the trick without the `to_string` attribute.

However, this part of the documentation explains the issue I described in the PR description where I can't get `as_ref` to return the same result as `to_string`.
> More importantly, this specifies what text to use when calling variant.to_string() with the Display derivation, or when calling variant.as_ref() with AsRefStr.

wp_api/src/lib.rs
crazytonyli: So, I took a look what VS code does. Below is the command VS code runs when click "Run tests" button under a `mod test_mod_name`:

```
cargo test --package wp_api --lib --all-features -- <test_mod_name> --show-output 
```

When naming this module as `tests`, VS code would run `cargo test [...] -- tests`. Since pretty much every test function would have `tests` in their qualified name, all tests are executed.

oguzkocer (approved):

PR #490, opened 2025-01-20
Let uniffi add `Sendable` conformation to endpoint response types

crazytonyli:
See internal discussion in p1737332574514769-slack-C06S1QS55K9.

PR #489, opened 2025-01-17
Implement dotorg plugin update check

crazytonyli:
The HTTP request encoding and parsing is based on [the update check implementation in core](https://github.com/WordPress/wordpress-develop/blob/6.7.1/src/wp-includes/update.php#L417-L426).

Here is an example request

```
curl -sX POST https://api.wordpress.org/plugins/update-check/1.1/ \
-H "User-Agent: WordPress/6.8; https://example.com" -d all=true \
-d plugins='{
  "plugins": {
    "jetpack/jetpack": {
      "Name": "Jetpack",
      "PluginURI": "https://jetpack.com/",
      "Version": "14.1.0",
      "Author": "Automattic",
      "AuthorURI": "https://jetpack.com",
      "TextDomain": "jetpack",
      "Network": false,
      "RequiresWP": "6.6",
      "RequiresPHP": "7.2",
      "RequiresPlugins": []
    }
  },
  "active": ["jetpack/jetpack"]
}' | jq
```

PR #500, opened 2025-01-21
Change `WpNetworkHeaderMap` to be a newtype for `HashMap<String, Vec<String>>`

oguzkocer:
This is a follow up to #490. As described in that PR, we are changing the exposed header map type to be `HashMap<String, Vec<String>>`. However, #490 doesn't change the inner type that we work with, leaving it as `http::HeaderMap`. Although there is some merit in using `http::HeaderMap` as our inner type, I don't think it outweighs the extra complexity of using 2 different types for the same thing. That's why this PR proposes converting `WpNetworkHeaderMap` to a new type for `HashMap<String, Vec<String>>`.

We still expose a function that converts this new type to `http::HeaderMap`, ensuring that Rust clients, including our integration test suite, can easily work with it.

The PR removes some unit tests that no longer make sense and adds a new one to validate that we are correctly parsing the number of pages. It also exposes additional fields for `WpNetworkRequest` so that Rust clients can use them without cloning the value - which wasn't an issue when using `Arc<T>`.

I believe there's still room for improvement in the request module, but as far as I can tell, this PR addresses our immediate needs. Any further improvements can be made in subsequent PRs.

### DIFF
--- a/wp_api/src/request.rs
+++ b/wp_api/src/request.rs
@@ -28,7 +28,7 @@ const HEADER_KEY_WP_TOTAL_PAGES: &str = "X-WP-TotalPages";
 pub struct ParsedResponse<DataType, ParamsType> {
     pub data: DataType,
     #[serde(skip)]
-    pub header_map: Arc<WpNetworkHeaderMap>,
+    pub header_map: WpNetworkHeaderMap,
     #[serde(skip)]
     pub next_page_params: Option<ParamsType>,
     #[serde(skip)]
@@ -64,7 +64,7 @@ impl InnerRequestBuilder {
         WpNetworkRequest {
             method: RequestMethod::GET,
             url: url.into(),
-            header_map: self.header_map().into(),
+            header_map: self.header_map(),
             body: None,
         }
     }
@@ -76,7 +76,7 @@ impl InnerRequestBuilder {
         WpNetworkRequest {
             method: RequestMethod::POST,
             url: url.into(),
-            header_map: self.header_map_for_post_request().into(),
+            header_map: self.header_map_for_post_request(),
             body: serde_json::to_vec(json_body)
                 .ok()
                 .map(|b| Arc::new(WpNetworkRequestBody::new(b))),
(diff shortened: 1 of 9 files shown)


wp_api/src/request.rs
```diff
@@ -244,10 +240,29 @@ impl WpNetworkHeaderMap {
...
 
     pub fn header_value_as_u32(&self, header_name: &str) -> Option<u32> {
-        self.inner
-            .get(header_name)
-            .and_then(|h_v| h_v.to_str().ok())
-            .and_then(|h| h.parse().ok())
+        [
+            self.0.get(header_name),
+            self.0.get(&header_name.to_lowercase()),
+        ]
```
crazytonyli, line 246: We should make an insensitive case comparison here, right? But this is not quite that? For example, `header_value_as_u32("x-wp-total")` returns none if the server uses HTTP1.

crazytonyli: I prefer the current implementation of using `HeaderMap`, since we no longer need to think about header names being case insensitive whenever reading from the "header map" type.

My thinking is we should stick with the `HeaderMap` _internally in the library_, but export the "header map" type as `HashMap` in the uniffi interface.

crazytonyli: BTW, _I think_ we should be able to implement [Lift](https://docs.rs/uniffi/latest/uniffi/trait.Lift.html) and [Lower](https://docs.rs/uniffi/latest/uniffi/trait.Lower.html) ourselves, instead of using the `uniffi::Record` macros, to expose `HeaderMap` directly as uniffi API. But that's probably a whole other topic to explore.

oguzkocer: Using a separate type just to be able to handle things in a case insensitive way doesn't make sense to me. When we use two different types, we have to consider them both and just its extra mental load is not really worth it. Furthermore, converting `http::HeaderMap` is not a free operation and we don't get many benefits out of that cost anymore as we have to keep converting back and forth between requests and responses.

I'd have preferred if we could use `http::HeaderMap` everywhere, but if we can't do that, I'd prefer we use `HashMap<String, Vec<String>>` everywhere.

oguzkocer: > For example, header_value_as_u32("x-wp-total") returns none if the server uses HTTP1.

Would you mind adding a failing test for this? I am probably missing something here because when I test it, it does seem to work with lowercase.

crazytonyli: Did you try it in unit tests or integration tests?

crazytonyli: I have added some test in 7317f99. The new integration test passes, but the new unit test fails.

oguzkocer: I meant a new unit test. The new unit test passes for me locally and it seems like it passed in CI as well. Am I missing something?

oguzkocer: Ah.. the commit is not in this branch. I thought it might be the case after I commented, sorry about that. Let me take a look.

oguzkocer: Right.. so, this doesn't work because we don't know the case of the original key. I just borrowed this from the link header lookup logic during the refactor and didn't pay attention to it afterwards.

---

I _think_ there are some cases this is an issue even in current `trunk`, because Rust is not the only one who might do a lookup, so using `http::HeaderMap` doesn't completely fix the issue. For example, if we expose the keys such as `X-WP-Total` and then Kotlin or Swift does the lookup, there is no guarantee that they'll do it in a case-insensitive way.

The solution to this would be to use a key such as `LowercaseString` which is what `http::HeaderName` does under the hood, but the problem is that'd also require using `uniffi::Object` because otherwise we can't control its construction.

There is also [unicase](https://docs.rs/unicase/latest/unicase/), but it has the same problem of requiring `uniffi::Object`.

crazytonyli:
I have fixed CI issues in #489. Rebasing this PR should resolve the CI failures.

wp_api/src/request.rs
crazytonyli: > For example, if we expose the keys such as X-WP-Total and then Kotlin or Swift does the lookup

IMHO, this issue is with Kotlin/Swift code. They should parse a string map using a proper HTTP header type, just as Rust should parse a string map using a proper HTTP header type.

oguzkocer: > this issue is with Kotlin/Swift code

This stance is against what we are trying to do with the `wordpress-rs` project. Whenever possible, Rust layer is responsible for making sure that Kotlin/Swift code can't go wrong.

---

I have an idea but I am not sure if it'll work through the FFI layer. I'll test it in a few hours.

PR #503, opened 2025-01-24
Remove an unnecessary clone

crazytonyli:


wp_api/src/request.rs
```diff
@@ -296,7 +296,7 @@ impl WpNetworkHeaderMap {
             .into_iter()
             .flat_map(|(header_name, values)| {
                 values.into_iter().flat_map(move |header_value| {
-                    Self::build_header_name_value(header_name.clone(), header_value)
+                    Self::build_header_name_value(&header_name, &header_value)
```
oguzkocer: We should replace this with `iter`:

```
let inner = hash_map
    .iter()
    .flat_map(|(header_name, values)| {
        values.iter().flat_map(|header_value| {
            Self::build_header_name_value(header_name, header_value)
        })
    })
    .collect::<Result<HeaderMap, WpNetworkHeaderMapError>>()?;
```

wp_api/src/request.rs
```diff
@@ -308,7 +308,7 @@ impl WpNetworkHeaderMap {
         let inner = hash_map
             .into_iter()
             .flat_map(|(header_name, header_value)| {
-                Self::build_header_name_value(header_name, header_value)
+                Self::build_header_name_value(&header_name, &header_value)
             })
             .collect::<Result<HeaderMap, WpNetworkHeaderMapError>>()?;
```
oguzkocer: We should replace this with `iter`:

```
let inner = hash_map
    .iter()
    .flat_map(|(header_name, header_value)| {
        Self::build_header_name_value(header_name, header_value)
    })
    .collect::<Result<HeaderMap, WpNetworkHeaderMapError>>()?;
```

oguzkocer (changes requested):

wp_api/src/request.rs
crazytonyli: Do you mind explaining the difference?

oguzkocer: Sure. `into_iter` takes ownership of the values whereas `iter` takes a reference.

In your implementation, you use `into_iter` and take ownership of `header_name` and `header_value` but turn them into reference to call `Self::build_header_name_value(&header_name, &header_value)` meaning the ownership wasn't required.

Probably the best way to understand this would be to add `let v = hash_map.get("foo");` after line `302` in your implementation. You'll notice that compiler says `hash_map` is moved due to the `into_iter` call, so you need to clone it if you want to keep using it. However, if you replace it with `iter`, it'll no longer be an issue.

In this specific case there won't be any difference between the two, because ownership rules are enforced at compile time, so the binary would work exactly the same. However, it wouldn't be idiomatic Rust to use `into_iter` when `iter` would suffice.

Hope that makes sense!

crazytonyli: Yeah, I get the ownership difference. Just don't see the difference (which is none as you explained?) they would make in this particular context.

> However, it wouldn't be idiomatic Rust to use into_iter when iter would suffice.

Hmmm, okay. Good to know!

crazytonyli: Updated in https://github.com/Automattic/wordpress-rs/pull/503/commits/2590cc4b44758b875140800dbc7896c66ef92668

wp_api/src/request.rs
```diff
@@ -293,10 +293,10 @@ impl WpNetworkHeaderMap {
         hash_map: HashMap<String, Vec<String>>,
     ) -> Result<Self, WpNetworkHeaderMapError> {
         let inner = hash_map
-            .into_iter()
+            .iter()
             .flat_map(|(header_name, values)| {
-                values.into_iter().flat_map(move |header_value| {
-                    Self::build_header_name_value(header_name.clone(), header_value)
+                values.iter().flat_map(move |header_value| {
```
oguzkocer:
```suggestion
                values.iter().flat_map(|header_value| {
```

oguzkocer (changes requested):

wp_api/src/request.rs
crazytonyli: Good catch!

oguzkocer (approved):

PR #507, opened 2025-01-24
List `/templates`

oguzkocer:
Follows established patterns to implement list [`/templates` endpoint.](https://developer.wordpress.org/rest-api/reference/wp_templates/#retrieve-a-template)

Note that `SparseTemplateContent` is not implemented as a contextual type even though some fields aren't available for some contexts. This is due to a weird combination of the `content` field being available for all 3 contexts, while `SparseTemplateContent` fields being available for only `edit` & `embed` contexts **and** `content` might be either a `String` or an `Object`. The implementation in this PR is the simplest solution I could come up with and I think it's a decent tradeoff because it won't cause any logic errors. The only drawback is that some fields will always be `None` for some contexts.

Note that I had to take a very long break during this implementation, so a lot of the details are not fresh on my mind. I did skim through the implementation a few times and didn't see anything missing/wrong, but I definitely might have missed something. Having said that, I have a lot more to implement for `/templates`, so I should be able to find any issues while doing that work. I just wanted to get this PR out as soon as possible as I need to switch to some login related work and I don't want this one to drag on even longer.

### DIFF
--- a/wp_api/src/templates.rs
+++ b/wp_api/src/templates.rs
@@ -1,0 +1,212 @@
+use crate::{
+    impl_as_query_value_from_as_str,
+    post_types::PostType,
+    posts::PostId,
+    url_query::{
+        AppendUrlQueryPairs, AsQueryValue, FromUrlQueryPairs, QueryPairs, QueryPairsExtension,
+        UrlQueryPairsMap,
+    },
+    EnumFromStrParsingError, UserId,
+};
+use serde::{Deserialize, Serialize};
+use std::str::FromStr;
+use strum_macros::IntoStaticStr;
+use wp_contextual::WpContextual;
+
+#[derive(
+    Debug,
+    Default,
+    Clone,
+    PartialEq,
+    Eq,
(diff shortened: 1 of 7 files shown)


wp_api/src/templates.rs
```diff
@@ -0,0 +1,225 @@
...
+        match s {
+            "draft" => Ok(Self::Draft),
+            "future" => Ok(Self::Future),
+            "pending" => Ok(Self::Pending),
+            "private" => Ok(Self::Private),
+            "publish" => Ok(Self::Publish),
+            value => Ok(Self::Custom(value.to_string())),
+        }
+    }
+}
```
crazytonyli: Use `strnum` for consistence

oguzkocer: Right. I implemented this before the `strum` changes and didn't think of it when I picked up the branch again. Addressed in 8d5958e0304cd2a1c56e9459cb6884d410787db4.

crazytonyli (approved):

PR #514, opened 2025-01-27
Rework auto discovery experience

oguzkocer:
This PR contains several changes we are making to the auto discovery experience. The following PRs were reviewed and merged into this feature branch: #513, #516, #517, #518, #522, #524.

Although this rework is not complete yet, we'll merge the feature branch with these changes and target `trunk` for our upcoming PRs.


jkmassel:
It's weird that the `plugins_update_check` fails, but we can deal with it later.

jkmassel (approved):
Approving this so we can integrate it with the `is_wordpress` checks – there's still some sharp edges but clients aren't using it yet so this shouldn't be a breaking change.

PR #517, opened 2025-01-29
Add Invalid SSL support and redirect tracking in Swift

jkmassel:


wp_api/src/request.rs
```diff
@@ -470,6 +480,22 @@ pub fn request_or_response_body_as_string(body: &[u8]) -> String {
...
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, uniffi::Record)]
+pub struct WpRedirect {
+    pub source: String,
+    pub destination: String,
+}
+
+impl WpRedirect {
+    #[uniffi::constructor]
```
oguzkocer: `uniffi::Record` types can't have constructors, so we should remove this attribute.
```suggestion
```

Cargo.toml
```diff
@@ -46,3 +46,4 @@ trybuild = "1.0"
 uniffi = "0.28"
 url = "2.5"
 uuid = "1.12"
+x509-cert = "0.3.0-pre.0"
```
oguzkocer, line 49: Do we need to use a pre-release version?

wp_api/src/ssl.rs
oguzkocer, line 1: I think there is a lot of room to improve the Rust usage in this module. I came up with [these suggestions](https://github.com/Automattic/wordpress-rs/commit/2edd9f7c8cce2318138076e2749da6d809ac5801). Please take a look at them and feel free to cherry-pick the commit if you like. Note that it might be easier to check the [new implementation](https://github.com/Automattic/wordpress-rs/blob/2edd9f7c8cce2318138076e2749da6d809ac5801/wp_api/src/ssl.rs) rather than follow the diff.

Also note that there is still a `TODO` and I've kept it in my suggestion. It'd be good to address that before we land the PR into the feature branch.

oguzkocer (changes requested):
I only reviewed the Rust code and in general it looks good to me. I made some suggestions to improve the new `ssl` module and requested an incorrect `uniffi::constructor` attribute to be removed. Please take a look at them and let me know what you think!

Cargo.toml
jkmassel: It'd be a lot weirder and less efficient not to – the current release version doesn't have any way to get at the fields inside it so we'd have to either do all of our own parsing, or dump it out to a string then parse it from there. This seemed like the less-bad option IMHO

oguzkocer: Sounds good. Hopefully we'll get a new release soon.

oguzkocer (approved):

PR #532, opened 2025-02-06
Add Login Spec

jkmassel:
As discussed with @oguzkocer – there are a lot of different things that can happen during login, so let's document them so we can ensure we're testing them for each HTTP implementation (the native Rust one, Apple platforms' `URLSession`, and Android)


wp_api/src/request.rs
```diff
@@ -151,9 +154,10 @@ impl WpNetworkRequestBody {
 }
 
 // Has custom `Debug` trait implementation
-#[derive(uniffi::Object)]
+#[derive(uniffi::Object, Clone)]
```
oguzkocer: I try to avoid deriving `Clone` as long as possible, so we don't accidentally abuse it. It's not the most costly `clone` since the header map is `Arc<T>`, but it's still pretty easy to avoid it, so I'd prefer if don't add it.
```suggestion
#[derive(uniffi::Object)]
```

wp_api/src/request.rs
```diff
@@ -162,10 +166,23 @@ pub struct WpNetworkRequest {
...
     pub fn request_id(&self) -> String {
         self.uuid.clone()
     }
 
+    /// Increments the retry count for this request by returning a copy of it
+    pub fn increment_retry_count(&self) -> Self {
+        let mut request = self.clone();
+        request.retry_count += 1;
+        request
+    }
```
oguzkocer: The following suggestion avoids deriving `Clone` and improves(?) the name of the function. We could maybe call it `retry` as well, because it'd be clear what it does from the types, but this might be a bit more clear at a first glance.

I don't like the name `increment_retry_count`, because it'd be something that takes `&mut self` and increment it in place. Although similar to `retry` suggestion, it's clear from the function signature that it returns a copy of the original request.

```suggestion
    /// Clones the request and increments the retry count
    pub fn clone_with_incremented_retry_count(&self) -> Self {
        Self {
            uuid: self.uuid.clone(),
            retry_count: self.retry_count + 1,
            method: self.method.clone(),
            url: self.url.clone(),
            header_map: Arc::clone(&self.header_map),
            body: self.body.clone(),
        }
    }
```

wp_api/src/login/url_discovery.rs
```diff
@@ -581,13 +614,14 @@ pub enum ParseApiRootUrlError {
 #[derive(Debug, thiserror::Error, uniffi::Error)]
 pub enum FetchApiDetailsError {
     #[error(
-        "Request execution failed!\nStatus Code: '{:?}'.\nResponse: '{}'",
+        "Request execution faiasdfled!\nStatus Code: '{:?}'.\nResponse: '{}'",
```
oguzkocer:
```suggestion
        "Request execution failed!\nStatus Code: '{:?}'.\nResponse: '{}'",
```

This is probably something you were testing and forgot to change it back? Or random typo 🙈

wp_api/src/login/url_discovery.rs
```diff
@@ -224,6 +227,36 @@ impl AutoDiscoveryAttemptResult {
...
+    }
+
+    /// Does the site look like a WordPress site?
+    fn is_wordpress_site(&self) -> bool {
+        self.is_wordpress_site.is_successful()
+    }
+}
+
+// Does the given URL look like it's on a local development environment for the purposes of the Login Spec?
+pub fn is_local_dev_environment_url(parsed_site_url: &ParsedUrl) -> bool {
```
oguzkocer: Is this only used by the testing suite? The implementation of this is pretty speculative, so if we need it for testing suite, I'd prefer if it's not part of the public API. I wouldn't mind even having separate implementations in Swift, Kotlin & Rust since it's so simple.

wp_api/src/login/url_discovery.rs
```diff
@@ -224,6 +227,36 @@ impl AutoDiscoveryAttemptResult {
...
+                    is_local_dev_environment_url(parsed_url)
+                } else {
+                    false
+                }
+            }
+        }
+    }
+
+    /// Does the site look like a WordPress site?
+    fn is_wordpress_site(&self) -> bool {
```
oguzkocer: If we are adding this helper, we should use it in `AutoDiscoveryResult::is_wordpress_site` as well.

wp_api/src/login/url_discovery.rs
```diff
@@ -224,6 +227,36 @@ impl AutoDiscoveryAttemptResult {
             Err(_) => None,
         }
     }
+
+    fn is_local_dev_environment(&self) -> bool {
```
oguzkocer: Please see my comment for `is_local_dev_environment_url` in this same file.

wp_api/src/login.rs
```diff
@@ -60,15 +61,75 @@ pub struct WpApiDetails {
...
+        namespace: "wordfence/v1".to_string(),
+        support_url: "https://www.wordfence.com/support/".to_string(), // TODO: Ensure this is correct with the WordFence folks
+    }]
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+pub struct KnownApplicationPasswordBlockingPlugin {
+    /// The name of the plugin.
+    pub name: String,
+
```
oguzkocer: Nitpick: Rest of the code base doesn't use these extra newlines. Although I don't have a strong preference either way, I'd prefer for us to stick to one style. I've seen this in your other implementations as well and I probably should have brought it up much sooner.

wp_api/src/login.rs
```diff
@@ -60,15 +61,75 @@ pub struct WpApiDetails {
...
+    }
+
+    /// Returns `true` if the site URL looks like a local development environment URL.
+    pub fn site_url_is_local_development_environment(&self) -> bool {
+        ParsedUrl::parse(self.url.as_str())
+            .is_ok_and(|parsed_url| is_local_dev_environment_url(&parsed_url))
+    }
+}
+
+pub fn known_application_password_blocking_plugins() -> Vec<KnownApplicationPasswordBlockingPlugin>
```
oguzkocer: I think we can make this an associated function for `KnownApplicationPasswordBlockingPlugin`. I've made this change in 5664111f6eaafeaa1a034ba374e546eed851635d also including the styling suggestion I mentioned. Please feel free to cherry-pick it if you agree with the suggestion.

wp_api/src/login.rs
```diff
@@ -60,15 +61,75 @@ pub struct WpApiDetails {
...
+            .collect()
+    }
+
+    /// Returns the site URL (as defined by the site itself, not by user input) as a string.
+    pub fn site_url_string(&self) -> String {
+        self.url.clone()
+    }
+
+    /// Returns `true` if the site URL looks like a local development environment URL.
+    pub fn site_url_is_local_development_environment(&self) -> bool {
```
oguzkocer: Please see my suggestion for `is_local_dev_environment_url`.

oguzkocer (commented):
@jkmassel I've left some comments about the Rust implementation. I haven't reviewed the documentation or the Swift changes yet.

wp_api/src/login/url_discovery.rs
jkmassel: I'm not sure if it needs to be public API at this point but it is used in two places to figure out which error message to show.

But it's an implementation detail so we can probably hide it from the public API

wp_api/src/login/url_discovery.rs
jkmassel: Done in 8cd8bef61d5e4f56188e531e9020e5f32b1f6d7f

wp_api/src/login.rs
jkmassel: Accepted!

wp_api/src/login.rs
jkmassel: Accepted as part of your other change

wp_api/src/login/url_discovery.rs
oguzkocer: I think we'll want to move the code that uses this to Rust and then we can hide it from the API. I am still worried of it being speculative, but maybe we can address that with some naming changes for both the function name and the error message.

Documentation/login-spec.md
```diff
@@ -0,0 +1,302 @@
...
+   - A domain ending in `.local`
+   - The hostname `localhost`
+
+Behavior:
+1. The system MUST reject the login attempt
+2. The system MUST display the error: "This site is a local development environment. You'll need to enable application passwords to connect to it with the app."
+
+Reference Implementation: http://localhost
+
+# 3: WordPress Admin URL Entry
```
oguzkocer, line 45: This says `WordPress Admin`, but it handles both `/wp-login` and `wp-admin` cases, should we rename it?

native/swift/Sources/wordpress-api/Foundation+Extensions.swift
```diff
@@ -17,3 +17,23 @@ public extension Date {
         wordpressDateFormatter.date(from: string)
     }
 }
+
+public extension TimeInterval {
+    static func fromRetryHeaderValue(_ value: String, now: Date = Date()) -> TimeInterval? {
```
crazytonyli: This function does not seem to be used anywhere except unit tests.

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -103,80 +62,166 @@ final class WpRequestExecutor: SafeRequestExecutor {
             ))
        }
     }
-    // swiftlint:enable function_body_length
+
+    private func fetch(request: URLRequest) async throws -> (Data, URLResponse) {
+        if #available(macOS 12.0, *) {
+            return try await self.session.data(for: request, delegate: executorDelegate)
```
crazytonyli: This function is available on all Apple platforms. Maybe we can change this function to be something like this?

```swift
#if os(Linux)
  return session.data(for: request)
#else
  return session.data(for: request, delegate: executorDelegate)
#endif
```

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -103,80 +62,166 @@ final class WpRequestExecutor: SafeRequestExecutor {
...
+            )
+        )
     }
 
-    private func getPeerCertificateChain(_ error: Error) throws -> [SSLCertificateInfo] {
+    private func getPeerCertificateChain(_ error: Error) -> [SSLCertificateInfo]? {
         let nserror = error as NSError
         let info = nserror.userInfo
 
         guard let certChainArray = info["NSErrorPeerCertificateChainKey"] as? NSArray else {
```
crazytonyli: We don't need the `parseCertificateChain` function below if changing this casting statement to `info["NSErrorPeerCertificateChainKey"] as? [SecCertificate]`

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// Uses the proposed site URL to scan the website it points to and find the Application Passwords login URL
+    ///
+    public func findLoginUrl(
+        forSite proposedSiteUrl: String
+    ) async throws(AutoDiscoveryAttemptFailure) -> ParsedUrl {
+        // All sites should have some form of authentication we can use
+        guard let apiDetails = try await loginAttempt(forSite: proposedSiteUrl).apiDetails() else {
+            preconditionFailure("It shouldn't be possible to hit this case – an error should be emitted instead")
```
crazytonyli: We should not need this statement if changing the `api_details()` function to `fn api_details(&self) -> Result<Arc<WpApiDetails>, AutoDiscoveryAttemptFailure>`

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
```diff
@@ -0,0 +1,119 @@
...
+    ) {
+        self.requestExecutor = requestExecutor
+        self.client = UniffiWpLoginClient(requestExecutor: requestExecutor, middlewarePipeline: middleware)
+    }
+
+    func loginAttempt(
+        forSite proposedSiteUrl: String
+    ) async throws(AutoDiscoveryAttemptFailure) -> AutoDiscoveryAttemptResult {
+
+        let discoveryResult = await client.apiDiscovery(siteUrl: proposedSiteUrl)
```
crazytonyli: Can we change this function to `fn api_discovery(&self, site_url: String) -> Result<Arc<AutoDiscoveryAttemptSuccess>, AutoDiscoveryAttemptFailure>` and do the if-else checks below within Rust?

crazytonyli (commented):
Regarding the Rust API comments, I tried to make the changes myself. But as I was making changes to them, it came to me that you guys may have tried that already. So, I haven't gone too deep into them, but I'm happy to make some proof of concept changes if needed.

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -18,83 +18,42 @@ extension SafeRequestExecutor {
...
     private let session: URLSession
-    private let redirectTracker = RedirectTracker()
+    private let executorDelegate: RequestExecutorDelegate
 
-    init(urlSession: URLSession) {
+    private let additionalHttpHeadersForAllRequests: [String: String]
+
+    init(
+        urlSession: URLSession,
+        additionalHttpHeadersForAllRequests: [String: String] = [:]
```
crazytonyli, line 27: This argument is not used.

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
```diff
@@ -0,0 +1,119 @@
+import Foundation
+import WordPressAPIInternal
+
+#if os(Linux)
+import FoundationNetworking
+#endif
+
+public final class WordPressLoginClient {
```
crazytonyli, line 8: BTW, there is not much going on in this type. Can we use `UniffiWpLoginClient` directly? You'll need to add the `loginURL(...)` and `credentials(...)` helper functions to it, but I don't think that's a terrible thing to do.

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
oguzkocer: The function currently returns `AutoDiscoveryResult/AutoDiscoveryUniffiResult` which is the result of combination of multiple attempts, so this suggestion wouldn't work. Happy to elaborate on it if you need more details.

wp_api/src/parsed_url.rs
```diff
@@ -161,6 +161,12 @@ impl From<Url> for ParsedUrl {
     }
 }
 
+impl From<ParsedUrl> for String {
+    fn from(input: ParsedUrl) -> Self {
+        input.url()
+    }
+}
+
```
oguzkocer: Usually it's preferred to implement the `Display` trait for this conversion, but shouldn't be a problem if you want to keep this.

native/kotlin/build.gradle.kts
```diff
@@ -94,6 +94,12 @@ fun setupJniAndBindings() {
         from("$cargoProjectRoot/test_media.jpg")
         into(generatedTestResourcesPath)
     }
+
+    tasks.register<Copy>("copySampleJSON") {
+        dependsOn(tasks.named("deleteTestResources"))
+        from("$cargoProjectRoot/native/swift/Tests/wordpress-api/Resources/Responses/localhost-json-root.json")
```
oguzkocer: Probably best to put this somewhere in the `native` directory, rather than under `swift` or `kotlin`.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/MockRequestExecutor.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlinx.coroutines.delay
+import okio.FileNotFoundException
+import uniffi.wp_api.MediaUploadRequest
+import uniffi.wp_api.RequestExecutor
+import uniffi.wp_api.WpNetworkHeaderMap
+import uniffi.wp_api.WpNetworkRequest
+import uniffi.wp_api.WpNetworkResponse
+import java.net.URI
+
+// Future improvement idea: This class could probably go in a Test Helpers package or something
```
oguzkocer: This is currently part of the public API, so I'd say rather than a future improvement idea, it should happen now. You can move it to `native/kotlin/api/kotlin/src/integrationTest/kotlin/` as this is similar to `native/kotlin/api/kotlin/src/integrationTest/kotlin/IntegrationTestHelpers.kt`

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -10,19 +10,34 @@ import okhttp3.OkHttpClient
...
 import java.io.File
+import java.net.UnknownHostException
+import javax.net.ssl.HostnameVerifier
+import javax.net.ssl.HttpsURLConnection
+import javax.net.ssl.SSLPeerUnverifiedException
+import javax.net.ssl.SSLSession
 
 class WpRequestExecutor(
-    private val okHttpClient: OkHttpClient = OkHttpClient(),
+    private var okHttpClient: OkHttpClient = OkHttpClient(),
```
oguzkocer: This change doesn't seem to be necessary. Instead, we should pass a default `WpRequestExecutorHostnameVerifier` to the `hostnameVerifier` so that it can later be updated.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -10,19 +10,34 @@ import okhttp3.OkHttpClient
...
     private val dispatcher: CoroutineDispatcher = Dispatchers.IO
 ) : RequestExecutor {
 
+    private var allowedHostnames: List<String> = emptyList()
+
+    fun addAllowedAlternativeNameForHostname(altname: String, hostname: String) {
+        allowedHostnames = allowedHostnames.plus(altname).plus(hostname)
+    }
```
oguzkocer: This should be tracked in `WpRequestExecutorHostnameVerifier`.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -91,4 +116,83 @@ class WpRequestExecutor(
...
+        val clientCertificates = HandshakeCertificates.Builder()
+            .addPlatformTrustedCertificates()
+            .addInsecureHost(allowedHostnames.first())
+            .build()
+
+        return okHttpClient.newBuilder()
+            .hostnameVerifier(WpRequestExecutorHostnameVerifier(allowedHostnames))
+            .sslSocketFactory(clientCertificates.sslSocketFactory(), clientCertificates.trustManager)
+            .build()
+    }
```
oguzkocer: We shouldn't build a new client. This defeats the whole purpose of using interfaces for things like `hostnameVerifier`. Instead, the instance that's used for `hostnameVerifier` should be updated, most likely outside of this executor.

Furthermore, there is a major issue with this; it overrides the `OkHttp` client provided by the client in the constructor.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -36,15 +51,25 @@ class WpRequestExecutor(
...
+                        requestHeaderMap = request.headerMap()
+                    )
+                }
+            } catch (e: SSLPeerUnverifiedException) {
+                handleInvalidSSLError(e, urlRequest)
+            } catch (e: UnknownHostException) {
+                handleUnknownHostException(e, urlRequest)
             }
+
+            error("Unknown Error occurred")
```
oguzkocer: This shouldn't be necessary and if I am not mistaken, the program execution can never make it here. If `try` succeeds it'll return the value, if the exception is caught it'll throw another exception and if it's not caught it'll stop the execution.

Compiler doesn't seem to be very smart about this as removing this line results in an error. However, if you change the helper functions to return the exception instead of throwing it and throw inside the catch blocks, then it'll stop complaining. In my opinion, this is also the correct way to do this and I was going to make the suggestion regardless of this line.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -91,4 +116,83 @@ class WpRequestExecutor(
...
+    private fun handleUnknownHostException(e: UnknownHostException, request: Request) {
+        val error = RequestExecutionErrorReason.NonExistentSiteError(
+            errorMessage = e.localizedMessage,
+            suggestedAction = "Check that the URL is valid and try again"
+        )
+
+        throwExceptionFor(error)
+    }
+
+    private fun throwExceptionFor(reason: RequestExecutionErrorReason) {
```
oguzkocer: I think it'd be better to implement a `buildRequestExecutionException` instead. It also ties back to my previous suggestion about removing the unnecessary `error("Unknown Error occurred")` line from `execute` function.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
oguzkocer: If I was working as a client developer and using this library function, I'd never thought that this function would be used as part of:

```
val clientCertificates = HandshakeCertificates.Builder()
    .addPlatformTrustedCertificates()
    .addInsecureHost(allowedHostnames.first())
    .build()
```

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -91,4 +116,83 @@ class WpRequestExecutor(
...
+            .addInsecureHost(allowedHostnames.first())
+            .build()
+
+        return okHttpClient.newBuilder()
+            .hostnameVerifier(WpRequestExecutorHostnameVerifier(allowedHostnames))
+            .sslSocketFactory(clientCertificates.sslSocketFactory(), clientCertificates.trustManager)
+            .build()
+    }
+
+    @Suppress("UnusedParameter")
```
oguzkocer:
```suggestion
```

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -91,4 +116,83 @@ class WpRequestExecutor(
...
+            siteCertificate = siteCertificate,
+            certificateChain = certificates.mapNotNull { it },
+            errorMessage = "Invalid certificate for domain",
+            suggestedAction = null
+        )
+
+        throwExceptionFor(error)
+    }
+
+    @Suppress("UnusedParameter")
```
oguzkocer: We shouldn't have unused parameters in an internal function and then suppress it. I think this kind of suppression is only reasonable if we are trying to avoid breaking public API.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -91,4 +116,83 @@ class WpRequestExecutor(
...
+            .hostnameVerifier(WpRequestExecutorHostnameVerifier(allowedHostnames))
+            .sslSocketFactory(clientCertificates.sslSocketFactory(), clientCertificates.trustManager)
+            .build()
+    }
+
+    @Suppress("UnusedParameter")
+    private fun handleInvalidSSLError(e: SSLPeerUnverifiedException, request: Request) {
+        // It's kind of weird and annoying that we need to make a second request to get
+        // this data, but it doesn't seem like we can get it from the response or the
+        // `SSLPeerUnverifiedException` directly.
```
oguzkocer: Maybe I am misunderstanding it, but this seems like a major security issue to me. We are retrying the request while allowing insecure connections without any explicit consent from the client. I think this is flow is only valid if it's done by the client after a user prompt. The information about invalid SSL certificate is definitely not important enough to throw away security considerations.

Please let me know if I am misunderstanding or misinterpreting it.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpLoginClient.kt
```diff
@@ -19,4 +19,29 @@ class WpLoginClient(
     suspend fun apiDiscovery(siteUrl: String): AutoDiscoveryUniffiResult = withContext(dispatcher) {
         internalClient.apiDiscovery(siteUrl)
     }
+
+    suspend fun loginUrl(siteUrl: String): String = withContext(dispatcher) {
```
oguzkocer: This implementation feels really off to me. (to be clear, not because of how you did it)

 I think we should add a helper function to Rust for it that returns a `Result` so that when we use it we either get the successful url or the most relevant error.

If you agree, I can open a PR about this targeting `trunk`, so you can rebase and update this function. We also probably should return a [Kotlin Result](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-result/) from this function.

oguzkocer (dismissed):
I haven't finished the review yet, but I left some pretty important comments about the Kotlin implementation which may affect the remaining files I haven't reviewed yet.

P.S: I usually review my comments before posting, but I am not able to do that at the moment and don't want to risk losing them. So, if there is anything poorly worded or plain incorrect, I apologize in advance as it was unintentional.

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
crazytonyli: Yes. I'd love more details.

I haven't looked the how Kotlin uses the `api_discovery` function. But the Swift function here turns the `AutoDiscoveryUniffiResult` into either a success result or a failure, which essentially is `Result<Arc<AutoDiscoveryAttemptSuccess>, AutoDiscoveryAttemptFailure>`.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
jkmassel: We're not actually re-sending the request – we're issuing a request to the same URL. 

The difference there is that no secret information is being transferred (for instance, the `Authorization` header will always be empty).

I could see an argument for stripping query params too, and we could make that change, but I wanted to start with that clarification.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpLoginClient.kt
jkmassel: I'm fine to move it into Rust – it needed to be implemented separately for Swift as well.

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
jkmassel: I'd be fine if we get there eventually, but the name is wrong – nobody consuming this library should ever see `UniffiWpLoginClient`. At the very least we'll have to typealias it

wp_api/src/parsed_url.rs
jkmassel: I'm not sure I want to commit to how `ParsedUrl` should handle `Display` at this point – right now we want it as a string, not to show it in UI so I think I'd still prefer this for now?

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
jkmassel: Not by our code, but this is a good back-door way to add a user agent (for instance).

Should that just be a separate PR?

crazytonyli: The same feature can be achieved using `URLSession`. But I guess this argument would allow adding headers _only_ for API requests originated from the library. I'm okay with it being in this PR if it's needed.

native/swift/Tests/wordpress-api/LoginTests.swift
```diff
@@ -7,37 +7,280 @@ import Testing
...
+
+            return true
+        })
+    }
+
+    @Test("Login Spec Example 9: Not a WordPress Site")
     func testNotWordPressSite() async throws {
-        let client = WordPressLoginClient(urlSession: .shared)
+        await #expect(performing: {
+            _ = try await client.findLoginUrl(forSite: "https://google.com")
```
crazytonyli: When the input is "google.com", the returned error is `ParseUrlError.RelativeUrlWithoutBase`. IMO that's an incorrect error. I'd expect the error to be `probablyNotAWordPressSite`.

oguzkocer: I tested this in Rust with the new `combinedResult` function from #652 and it returned `ProbablyNotAWordPressSite` as expected. Once this PR is updated, the issue should be resolved for Swift as well.

```
// Full error message for `google.com`
FindApiRoot {
    parsed_site_url: ParsedUrl {
        inner: Url {
            scheme: "https",
            cannot_be_a_base: false,
            username: "",
            password: None,
            host: Some(
                Domain(
                    "google.com",
                ),
            ),
            port: None,
            path: "/",
            query: None,
            fragment: None,
        },
    },
    find_api_root_failure: ProbablyNotAWordPressSite,
}
```

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
jkmassel: This was adjusted to be done in Rust

native/swift/Sources/wordpress-api/Foundation+Extensions.swift
jkmassel: Removed in 8e99aa795f8abea52c430b1eb303e536aba6c0a7 – this was moved to Rust!

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
jkmassel: I'm hoping we can limit the number of things that a developer has to handle themselves in `URLSession`, but I guess we'll see?

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
jkmassel: Addressed in 23fdee98623daf4994c57f3886a8023bb90b2418 – this was because all the development happened on macOS

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
jkmassel: Nice! Done in c07115cc535a72a1b600492b762d5f510812df8d

native/swift/Tests/wordpress-api/LoginTests.swift
crazytonyli: @jkmassel Do you mind adding `google.com` as a test case, which should be fairly easy with `@Test(arguments: ["google.com", "https://google.com"])`?

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    public protocol AuthenticatorProtocol {
+        func authenticate(url: URL, callbackURL: URL) async throws -> URL
+    }
```
crazytonyli: This protocol can be removed.

crazytonyli (approved):
Left a couple of minor comments.

native/kotlin/build.gradle.kts
jkmassel: I'd say let's do this as soon as we have a second item

PR #533, opened 2025-02-06
Fix Plugin Update Test failures

jkmassel:
There are some issues with parsing responses from the WP.org API update server.

This PR addresses those.


oguzkocer (approved):
~I believe these test cases belong together. It's fine to make them either integration tests or unit tests, because we can make a case for why we should pick one over the other. However, since they do the same thing, we should pick only one of them.~

~Approving the PR so it can be merged once one of the tests are moved.~

**Edit:** They are already both unit tests, I am just blind 🤦‍♂️

PR #536, opened 2025-02-06
Implement `is_wordpress_site` for api discovery

oguzkocer:
This is the first iteration of the `is_wordpress_site` implementation, which checks if the given site is a WordPress site following a few conditions:

* Whether the site has API link header `https://api.w.org/`
* Whether `/wp-json` returns a parseable JSON including `namespaces` field
* Whether the source HTML has a `generator` tag that mentions `WordPress`
* Whether the source HTML mentions `wp-content` or `wp-includes` in the `href` attribute in `link` tags or `src` attribute in `script` tags

---

In this PR, `is_wordpress_site` is built as a separate check from api discovery. However, we discussed this with @jkmassel and decided to combine them into a single API to make it easier to use. We also have a few other improvements coming up, such as changing some details about the constructed attempts, and possibly adding more checks.


jkmassel (approved):
Looks good – discussed next steps offline:

1. Add more test cases
2. Support checking for the `<link rel="https://api.w.org/" [...] />` tag

PR #538, opened 2025-02-07
Improve constructed attempts for api discovery and handle parsing link tag for api root url

oguzkocer:
This PR improves the constructed api discovery attempts by removing the `wp-admin`, `wp-admin.php`, `wp-admin/`, `wp-login`, `wp-login.php` & `wp-login/` suffixes.

It also checks the link tags in the source HTML to find the api root url following this pattern: `<link href="http://localhost/wp-json/" rel="https://api.w.org/">`. Since we always parse HTML, I've made this as the first thing we do, so if we find the link tag, we don't request the header anymore.

This setup doesn't seem correct to me, but I think it works. So, instead of trying to figure everything out in one go, I'd like to see if this does what we need to do and if so, then improve it.

---

We have had one last case remaining in https://github.com/Automattic/wordpress-rs/pull/265 where the API discovery didn't work. I think with the addition of this new implementation, it now works. However, the authorization url in #265 was `https://jetpack.wpmt.co/wp-admin/authorize-application.php` whereas according to this implementation it's `https://aggressive-caching.wpmt.co/wp-admin/authorize-application.php`.

### DIFF
--- a/wp_api/src/login/login_client.rs
+++ b/wp_api/src/login/login_client.rs
@@ -67,12 +67,20 @@ impl WpLoginClient {
         attempt: AutoDiscoveryAttempt,
     ) -> AutoDiscoveryAttemptResult {
         let attempt_site_url = attempt.attempt_site_url.as_str();
+        let parse_html_result = self
+            .fetch_site(attempt_site_url)
+            .await
+            .map(|r| IsWordPressSiteParseHtmlResult::parse_response(&r.body_as_string()));
         let api_link_header_result = self.find_api_root_url(attempt_site_url).await;
         let discovery_result = self
-            .inner_attempt_api_discovery(attempt_site_url, api_link_header_result.clone())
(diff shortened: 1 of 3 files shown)


jkmassel (approved):
> This setup doesn't seem correct to me, but I think it works. So, instead of trying to figure everything out in one go, I'd like to see if this does what we need to do and if so, then improve it.

This is fine - it handles more valid cases and we can refine for efficiency later once we have a robust test suite.

> We have had one last case remaining in https://github.com/Automattic/wordpress-rs/pull/265 where the API discovery didn't work. I think with the addition of this new implementation, it now works. However, the authorization url in https://github.com/Automattic/wordpress-rs/pull/265 was https://jetpack.wpmt.co/wp-admin/authorize-application.php whereas according to this implementation it's https://aggressive-caching.wpmt.co/wp-admin/authorize-application.php.

That's correct!

PR #539, opened 2025-02-09
Introduce `WpNaiveDateTime` & `WpGmtDateTime`

oguzkocer:
This PR introduces `WpNaiveDateTime` and `WpGmtDateTime` types for use across all endpoints. It's currently only used for `/posts`, but I plan to switch to using these types at other endpoints in a follow-up.

We're using two different types because the WordPress API returns both `date` and `date_gmt` types for most of its endpoints. The `date_gmt` field contains a date with timezone information, whereas the `date` field is in the user's timezone, which isn't known when we parse the date. That's why we're using `WpNaiveDateTime`, which requires timezone information.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,6 +340,7 @@ dependencies = [
  "iana-time-zone",
  "js-sys",
  "num-traits",
+ "serde",
  "wasm-bindgen",
  "windows-targets 0.52.6",
 ]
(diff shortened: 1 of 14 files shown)


oguzkocer:
Closing in favor of https://github.com/Automattic/wordpress-rs/pull/550.

PR #543, opened 2025-02-11
Update uniffi to 0.29

crazytonyli:


oguzkocer:
@crazytonyli I assume you're testing to see if the update takes care of the Swift issues, which is why this PR is targeting `date-type` branch. However just in case that wasn't the intention, I think we should update `uniffi` in `trunk` and then rebase the date type PR.

crazytonyli:
Yep. I just wanted to double-check that it fixes the issue in your PR. I have open a new PR to trunk: https://github.com/Automattic/wordpress-rs/pull/545.

PR #546, opened 2025-02-11
WpGmtDateTime to Foundation.Date

crazytonyli:


wp_api/src/date.rs
```diff
@@ -22,6 +22,14 @@ impl WpGmtDateTime {
     }
 }
 
+impl WpGmtDateTime {
+    fn from_timestamp(seconds: i64) -> Self {
```
jkmassel, line 26: Should this just be a `From<i64>` trait?

PR #547, opened 2025-02-12
Add Swift Docs

jkmassel:
Adds auto-documentation support for Swift.

Builds the docs and uploads them to Buildkite for each build.

Later, we'll push this somewhere interesting.

To Test:
- Ensure CI passes

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,7 +28,9 @@ var package = Package(
             targets: ["WordPressAPI"]
         )
     ],
-    dependencies: [],
+    dependencies: [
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -28,7 +28,9 @@ var package = Package(
             targets: ["WordPressAPI"]
         )
     ],
-    dependencies: [],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
+    ],
```
crazytonyli, line 33: I wish SPM had "development dependencies". There is no need to download this dependency when used in the app.

crazytonyli (approved):

PR #549, opened 2025-02-12
Add `https` auto discovery attempt for `http` urls

oguzkocer:
If the url starts with `http`, but not `https`, we now add a new `AutoHttps` api discovery attempt type. The PR also adds `http://optional-https.wpmt.co` as a test case, which was resulting in a parsing error before this PR. Finally, the PR adds some helpers to `AutoDiscoveryAttempt` and improves the readability of the unit tests - although it's still quite difficult.


jkmassel (approved):
This change makes the `HTTP URL with HTTPS Support` example in #532 work, so I'm approving it.

PR #550, opened 2025-02-13
Introduce `WpGmtDateTime`

oguzkocer:
This PR introduces `WpGmtDateTime` which is a wrapper for [`chrono::DateTime<chrono::offset::Utc>`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html).

We use `uniffi::custom_type!` macro to convert it into `java.util.Date` in Kotlin bindings using `i64` as the underlying value. To make sure this conversion is done correctly, we export a couple assertion functions in the native bindings. These assertion functions are used in Kotlin to make sure the dates are converted correctly between Rust and Kotlin. We'll do the same for Swift in a follow up PR.

### DIFF
--- a/wp_api/src/date.rs
+++ b/wp_api/src/date.rs
@@ -28,3 +28,45 @@ uniffi::custom_type!(WpGmtDateTime, i64, {
     lower: |date_time| date_time.0.timestamp(),
     try_lift: |seconds| Ok(WpGmtDateTime::from_timestamp(seconds)),
 });
+
+// Assertion functions that should only be used by the native test suite
+// These are hidden from the Rust public API, but will be visible/usable in the generated bindings
+mod native_test_helper {
+    use super::WpGmtDateTime;
+    use chrono::offset::FixedOffset;
+
+    const OFFSET_FOR_EXAMPLE_DATE: i32 = 60 * 60 * 2;
+    const EXAMPLE_DATE: &str = "2020-08-14T15:00:00+02:00";
+
+    #[uniffi::export]
+    fn assertion_example_date_that_can_be_used_to_verify_conversion_between_rust_and_native(
+    ) -> WpGmtDateTime {
+        EXAMPLE_DATE
+            .parse::<WpGmtDateTime>()
+            .expect("Example date is parseable")
+    }
+
+    #[uniffi::export]
+    fn assert_date_is_converted_from_native_to_rust_correctly(date: WpGmtDateTime) {
+        assert_eq!(
+            date.0
+                .with_timezone(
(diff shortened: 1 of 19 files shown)


wp_api/src/date.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct WpGmtDateTime(#[serde(with = "wp_utc_date_format")] pub DateTime<Utc>);
+
+impl WpGmtDateTime {
+    pub fn from_timestamp(seconds: i64) -> Self {
+        let date_time =
+            DateTime::<Utc>::from_timestamp(seconds, 0).unwrap_or(DateTime::<Utc>::UNIX_EPOCH);
+        Self(date_time)
+    }
```
crazytonyli, line 14: What do you think about adding a few unit tests for this function with different kinds of values (zero, positive, negative)?

crazytonyli (approved):

wp_api/src/date.rs
oguzkocer: I don't mind it, but would it be useful? We'd be testing the underlying implementation for `DateTime::<Utc>` and I generally don't add those tests as I feel they don't add any value.

Btw, I am not too thrilled about the `.unwrap_or(DateTime::<Utc>::UNIX_EPOCH)` part, but with `custom_type!` I think we either do this or panic. Out of them, I'd prefer us to panic so that if there are any issues, we'd be aware of it. Otherwise, this may only show up as a support request. I kept your implementation from that temporary PR and I was thinking of bringing it up in the PR description.

oguzkocer:
I am going to merge this as is so that we can follow it up with [the PR that adds Swift support.](https://github.com/Automattic/wordpress-rs/pull/551)

There is an open discussion about further unit tests and whether we should panic if we can't create a date from timestamp. If any further actions are necessary for these discussions, I'll open follow up PRs.

wp_api/src/date.rs
crazytonyli: I thought ensuring we always use "second" as the unit might be useful. Plus, adding test for some distant past and future dates might be helpful to ensure `WpGmtDateTime` continues to support a large range of dates.

The only reason that the parsing can fail is that the date is super distant. I think the 1970 date is an okay fallback for this parsing error because the original super distant is probably a mistake anyway. At bare minimal, we should not panic when parsing them.

oguzkocer: Addressed in https://github.com/Automattic/wordpress-rs/pull/556.

PR #551, opened 2025-02-13
Convert WpGmtDateTime to Foundation.Date

crazytonyli:


native/swift/Tests/wordpress-api/DateTests.swift
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Test
+    func swiftToRust() {
+        assertDateIsConvertedFromNativeToRustCorrectly(date: testDate)
+    }
+
+    @Test
+    func rustToSwift() {
+        let date = assertionExampleDateThatCanBeUsedToVerifyConversionBetweenRustAndNative()
```
oguzkocer, line 19: We should call `assertDateIsConvertedFromNativeToRustCorrectly(date)` here, [as we do in Kotlin](https://github.com/Automattic/wordpress-rs/pull/550/files#diff-911a4250f804d3c25efa620109c83f6175af77584a68aced551579f845f76891R23-R24), so that we can verify the round trip between platforms.

crazytonyli: I have added another unit test in https://github.com/Automattic/wordpress-rs/pull/551/commits/5fe0e49dae1ba230a0b18cfeba2a476d67540af1

oguzkocer (approved):

PR #557, opened 2025-02-16
Add a public API to convert URLSession to an RequestExecutor

crazytonyli:
`WordPressOrgClient` needs a `RequestExecutor`. We need a public API to create such instances. This PR adds a simple function to convert `URLSession` to a `RequestExecutor`. I don't mind publicizing `WpRequestExecutor`. It does not make much difference compared to this PR's change.

### DIFF
--- a/native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
+++ b/native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
@@ -5,12 +5,6 @@ import WordPressAPIInternal
 import FoundationNetworking
 #endif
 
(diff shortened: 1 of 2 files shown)


jkmassel (changes requested):
`WordPressLoginClient` and `WordPressAPI` use `init(urlSession` as the public API – could this do the same?

crazytonyli:
@jkmassel I have replaced the original solution with a `WordPressLoginClient(urlSession:)` initializer. Let me know what you think.

jkmassel (approved):

PR #558, opened 2025-02-16
Use plugin slug types instead of String

crazytonyli:
What says in the title.

### DIFF
--- a/wp_api/src/plugins.rs
+++ b/wp_api/src/plugins.rs
@@ -104,7 +104,7 @@ impl Display for PluginSlug {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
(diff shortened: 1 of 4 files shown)


oguzkocer:
CI has failed, so I am removing myself from the reviewer list. Please feel free to re-request a review once the CI is green.

**P.S:** I use email notifications to keep track of review requests and if I can't review a PR when I open the email, I have to snooze the email and keep checking the PR to see if it's ready for review which is troublesome if it's a frequent issue. That's why I always try to either locally verify that everything is good, or open the PR but wait for CI to request a review. You don't have the do the same, but I need some workaround in my process. So, from now on, I'll be removing myself from reviewer list if the PR is not reviewable and wait for your re-request. Hope that's OK with you 🙇‍♂️

oguzkocer (approved):

PR #559, opened 2025-02-17
Add uniffi bindings to ser/de `PluginInformation`

crazytonyli:
Add a macro to generate two functions to ser/de a Rust instance to binary: `serialize_xxx` and `deserialize_xxx`. These two functions are internal APIs and are only exported as uniffi bindings.

I tried to use `bincode`, but due to [some known issues](https://github.com/bincode-org/bincode/blob/0ab1f715ae824b84010f0d020a0ecd47227223e2/src/features/serde/mod.rs#L42-L51), it can't ser/de `PluginInformation`. Even though `serde_json` may be slow, it's guaranteed to be correct since it's used to ser/de all types in this library.

### DIFF
--- a/wp_api/src/uniffi_serde.rs
+++ b/wp_api/src/uniffi_serde.rs
@@ -1,0 +1,30 @@
+#[derive(Debug, uniffi::Object, thiserror::Error)]
+#[error("{inner:?}")]
+pub(crate) struct UniffiSerializationError {
+    inner: serde_json::Error,
+}
+
+impl From<serde_json::Error> for UniffiSerializationError {
+    fn from(inner: serde_json::Error) -> Self {
+        Self { inner }
+    }
+}
+
+#[macro_export]
+macro_rules! uniffi_export_serialization {
+    ($func_name:ident, $type:ty) => {
+        paste::paste! {
+            #[uniffi::export]
+            fn [<serialize_ $func_name>](value: $type) -> Result<Vec<u8>, Arc<$crate::uniffi_serde::UniffiSerializationError>> {
+                serde_json::to_vec(&value)
+                    .map_err(|e| Arc::new(e.into()))
+            }
+
+            #[uniffi::export]
+            fn [<deserialize_ $func_name>](value: Vec<u8>) -> Result<$type, Arc<$crate::uniffi_serde::UniffiSerializationError>> {
+                serde_json::from_slice(&value)
+                    .map_err(|e| Arc::new(e.into()))
+            }
+        }
+    };
+}
--- a/wp_api/src/uniffi_serde.rs
+++ b/wp_api/src/uniffi_serde.rs
@@ -1,12 +1,14 @@
-#[derive(Debug, uniffi::Object, thiserror::Error)]
-#[error("{inner:?}")]
-pub(crate) struct UniffiSerializationError {
-    inner: serde_json::Error,
+#[derive(Debug, uniffi::Error, thiserror::Error)]
+pub(crate) enum UniffiSerializationError {
+    #[error("{reason:?}")]
+    Serde { reason: String },
 }
 
 impl From<serde_json::Error> for UniffiSerializationError {
-    fn from(inner: serde_json::Error) -> Self {
-        Self { inner }
+    fn from(error: serde_json::Error) -> Self {
+        UniffiSerializationError::Serde {
+            reason: error.to_string(),
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


jkmassel (approved):
This looks fine to me – are you thinking that Swift will have its own mechanism for storing the data?

Or is that just coming in a future PR?

wp_api/src/uniffi_serde.rs
```diff
@@ -0,0 +1,30 @@
+#[derive(Debug, uniffi::Object, thiserror::Error)]
```
oguzkocer: As far as I know, we use `uniffi::Error` type for all other error types. Could you elaborate why we should use `uniffi::Object` here? There is no exported function to read the error details at the moment, so I don't know what the intention is.

Even if we need to wrap `serde_json::Error` type, I think my preference would be for us to use `uniffi::Error` for the error type itself and put the `uniffi::Object` into one of the variants. That way we can also add some of the expected error variants as the `enum` variant.

Also, I vaguely remember not using `uniffi::Error` in `Result<T, E>` causing issues with `uniffi` - possibly in Kotlin? I unfortunately can't remember the details and maybe they have changed recently.

wp_api/src/uniffi_serde.rs
```diff
@@ -0,0 +1,30 @@
...
+        Self { inner }
+    }
+}
+
+#[macro_export]
+macro_rules! uniffi_export_serialization {
+    ($func_name:ident, $type:ty) => {
+        paste::paste! {
+            #[uniffi::export]
+            fn [<serialize_ $func_name>](value: $type) -> Result<Vec<u8>, Arc<$crate::uniffi_serde::UniffiSerializationError>> {
```
oguzkocer: I am hoping we'll switch to `uniffi::Enum`, so `Arc` won't be necessary anymore. However, if we don't, we should use `std::sync::Arc` here. It looks like [I didn't do this correctly in `api_client`](https://github.com/Automattic/wordpress-rs/blob/22e854728af76c230b3feae92193273fe102242d/wp_api/src/api_client.rs#L183) which I'll open a PR to fix.

wp_api/src/wordpress_org/plugin_directory.rs
```diff
@@ -328,4 +331,33 @@ mod tests {
...
+    fn serialization_round_trip() {
+        let json_string =
+            include_str!("../../tests/plugin-directory/plugin-with-expected-types.json");
+        let plugin = serde_json::from_str::<PluginInformation>(json_string).unwrap();
+        let serialized = serialize_plugin_information(plugin).unwrap();
+        let deserialized = deserialize_plugin_information(serialized).unwrap();
+
+        let expected = serde_json::from_str::<PluginInformation>(json_string).unwrap();
+        assert_eq!(deserialized.name, expected.name);
+        assert_eq!(deserialized.author, expected.author);
```
oguzkocer: I think it'd be worth adding `PartialEq` and `Eq` to `PluginInformation` and its inner types, and then use:
```
assert_eq!(deserialized, expected);
```

I tested this locally and it worked as expected.

wp_api/src/wordpress_org/plugin_directory.rs
```diff
@@ -1,8 +1,8 @@
 use super::de::deserialize_default_values;
-use serde::Deserialize;
-use std::{collections::HashMap, fmt::Debug};
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, fmt::Debug, sync::Arc};
```
oguzkocer: Importing `sync::Arc` shouldn't be necessary. See my comment about using `std::sync::Arc`.

oguzkocer (changes requested):
I left a few comments one of which is probably worth discussing before we merge the PR.

wp_api/src/uniffi_serde.rs
crazytonyli: I just thought it'd be a simpler solution than a one-variant enum.

crazytonyli:
@jkmassel Yep. The app could call these functions to save/parse data to/from the file system.

wp_api/src/uniffi_serde.rs
crazytonyli: I have changed it to an enum in https://github.com/Automattic/wordpress-rs/pull/559/commits/c93a8c73730154f80540870a27518624124484a1

oguzkocer (approved):

PR #572, opened 2025-02-20
Move Login Application Errors to Rust

jkmassel:
Moves a bunch of application-layer login errors out of #532, pushing them down into Rust so the implementation can be shared between Swift and Kotlin.

This is a separate PR to keep things small and easy-to-review.

### DIFF
--- a/wp_api/src/login/url_discovery.rs
+++ b/wp_api/src/login/url_discovery.rs
@@ -6,7 +6,7 @@ use crate::{
 };
 use scraper::{Html, Selector};
 use serde::Deserialize;
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, fmt::Display, sync::Arc};
 
 pub(crate) const API_ROOT_LINK_HEADER: &str = "https://api.w.org/";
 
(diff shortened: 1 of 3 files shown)


wp_api/src/parsed_url.rs
```diff
@@ -11,6 +13,20 @@ impl ParsedUrl {
     }
 }
 
+impl Display for ParsedUrl {
```
oguzkocer, line 16: I think this goes a little bit against the common practice of wrapper types implementing the `Display` of the inner type. Although that may be fine in some cases where it's obvious, I feel like in this one it can trip one up as they were expecting the [forwarded implementation of `Url`.](https://docs.rs/url/latest/url/struct.Url.html#impl-Display-for-Url)

Would using a different method for `ParsedUrl` address your needs? That way we can also make it available to `uniffi` clients.

crazytonyli (approved):

wp_api/src/login/url_discovery.rs
```diff
@@ -426,6 +427,33 @@ pub enum AutoDiscoveryAttemptFailure {
...
+    #[error("This site is a local development environment. You'll need to enable application passwords to connect to it with the app.")]
+    SiteIsLocalDevelopmentEnvironment {
+        parsed_site_url: ParsedUrl,
+        api_root_url: ParsedUrl,
+    },
+    #[error("Application Passwords is not enabled for this site – this is likely because we can't establish a secure connection to it. Please add an SSL certificate to this site and try again.")]
+    ApplicationPasswordsDisabledForHttpSite {
+        parsed_site_url: ParsedUrl,
+        api_root_url: ParsedUrl,
+    },
```
oguzkocer: All these cases seem to be special cases for `ApplicationPasswordsNotSupported`. If I am not mistaken, they don't represent separate states in the api discovery which is what `AutoDiscoveryAttemptFailure` is for.

I think we should move these to a new enum such as `ApplicationPasswordsNotSupportedReason`. That way, we won't need to put `parsed_site_url` and `api_root_url` fields in every variant since it can be retrieved from the parent. It should also significantly clean up the rest of the code from this PR.

oguzkocer (changes requested):

wp_api/src/parsed_url.rs
jkmassel: Easy enough to just have a separate method – see 15722a70934793825cc1a863c72f359bca07d7d4

oguzkocer: I am sorry, I didn't do a good job of explaining my concern. Usually, you'd expect the following implementation in a wrapper type:

```
impl Display for ParsedUrl {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.inner)
    }
}
```

Notice how we just forward the inner type's implementation. This is not just for `Display` trait, but a common practice for all traits. As I mentioned, there are cases where this is broken, but they tend to be obvious why it's implemented in a different way.

So, what I was suggesting was to drop the `Display` implementation from the current PR - or replace it with the one in this comment. Then, use the helper method wherever you need to convert `ParsedUrl` to pretty url. Of course this only works if you don't need to rely on `Display` which is what my original question was about.

Hope that clears it up.

jkmassel: Oh yikes – I meant to do what was in your example and somehow had the wrong thing 🤦 

I've updated it to resolve this

wp_api/src/login/url_discovery.rs
```diff
@@ -420,14 +421,36 @@ pub enum AutoDiscoveryAttemptFailure {
...
+}
+
+impl ApplicationPasswordsNotSupportedReason {
+    fn error_message(&self, parsed_site_url: impl Display) -> String {
+        match self {
+            Self::ApplicationPasswordBlockedByPlugin { plugin } => format!("Unable to login to {} – the {} plugin might have disabled Application Passwords. Please visit {} to learn more", parsed_site_url, plugin.name, plugin.support_url),
+            Self::ApplicationPasswordBlockedByMultiplePlugins => format!("Unable to login to {} – there are multiple installed plugins that might have disabled Application Passwords. Please disable them and try again.", parsed_site_url),
+            Self::SiteIsLocalDevelopmentEnvironment => "This site is a local development environment. You'll need to enable application passwords to connect to it with the app.".to_string(),
+            Self::ApplicationPasswordsDisabledForHttpSite => "Application Passwords is not enabled for this site – this is likely because we can't establish a secure connection to it. Please add an SSL certificate to this site and try again.".to_string(),
+    }
```
oguzkocer, line 450: Auto-formatting had some trouble here due to lengthy strings, but I think it's fine for now. It should be fixed with the upcoming localization changes.

oguzkocer (approved):

PR #573, opened 2025-02-21
Add 429 support to Rust

jkmassel:
This PR is a proposal for how to add automated 429 retry (and a mechanism to bail out if there are too many retry responses). 

We can later apply it to the main API client, because we should have some mechanism of handling that there too, but I'd like to start here, get feedback on the approach, then carry on.

This PR deliberately doesn't:

1. Handle non-integer `Retry-After` values
2. Do anything about `401` or `403` responses, which IMHO should be handled in the same place – treating these like a transport error makes sense because we can't do anything unless they're resolved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ resolver = "2"
 [workspace.dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
+async-std = "1.10"
 base64 = "0.22"
 chrono = "0.4"
 clap = "4.5"
--- a/wp_api/src/login/login_client.rs
+++ b/wp_api/src/login/login_client.rs
@@ -15,8 +15,10 @@ use crate::{
         endpoint::WpEndpointUrl, RequestExecutor, RequestMethod, WpNetworkHeaderMap,
         WpNetworkRequest, WpNetworkResponse,
     },
-    ParseUrlError, ParsedUrl, RequestExecutionError,
+    ParseUrlError, ParsedUrl, RequestExecutionError, RequestExecutionErrorReason,
 };
+use async_std::task;
+use std::time::Duration;
 use std::{str, sync::Arc};
 use uuid::Uuid;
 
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -16,6 +16,7 @@ resolver = "2"
 [workspace.dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
+async-std = "1.10"
```
crazytonyli, line 19: I thought we didn't want to introduce an async runtime, which may interfere with the one in uniffi-rs?

wp_api/src/login/login_client.rs
```diff
@@ -329,19 +329,49 @@ impl WpLoginClient {
...
+            if retry_count >= 3 {
+                return Err(RequestExecutionError::RequestExecutionFailed {
+                    status_code: Some(response.status_code),
+                    redirects: None,
+                    reason: RequestExecutionErrorReason::MisconfiguredRateLimitError {},
+                });
+            }
+
+            if response.is_http_429() {
+                let retry_after = response.get_retry_after();
```
crazytonyli, line 404: Make sure it's the value is in a given range? So that, we don't retry too soon and too often, also don't wait forever.

crazytonyli:
This retry mechanism hides the fact that the app is retrying for the users instead of manually pulling to refresh. When it happens, users will see a loading view for a long time (depending on the `retry-after` value). I wonder if it'd be better to just straight up telling users that "please try again in xxx minutes".

BTW, is there a real world example? I'm guessing this is not a native WordPress feature, but it's typically configured in apache or nginx? _I think_ typical rate limit error would mean needing to wait for minutes rather than seconds. That would mean users may see a long loading view, which may gives an impression that the app is slow or stuck.

Cargo.toml
jkmassel: IINM the stuff in `async-std` should work in any async context because it's just a `Future`, but I might have that wrong?

crazytonyli: I'm not questioning whether the crate (or the `task::sleep` async function) would work in this library. I remember Oguz mentioned (maybe in one of the PR reviews) that we don't want to introduce another async runtime (like `tokio`) to this library. `async-std` is just another async runtime, like `tokio`. @oguzkocer Please correct me if I am wrong about this.

oguzkocer: I don't have any experience using `async-std` either, but afaik `async-std` is a little different from `tokio` because it can be run on other executors, including `tokio`. Furthermore, both of them are sort of like ecosystems where some parts of them are related to executors (more so in `tokio`) and some of them are `futures` utilities.

In my opinion, in `wp_api`, we should try to use `futures` as much as possible. I wouldn't be against adding a dependency for `futures` related utilities, but I think it's best to avoid `async-std`, `tokio` etc as much as possible.

crazytonyli: Does `futures` have a `sleep` function? Its API doc says there is no timer API.

I did a little bit of digging. The `task::sleep` function internally calls [`async_io::Timer::after`](https://docs.rs/async-io/latest/async_io/struct.Timer.html#method.after). Maybe we can use that function directly?

BTW, unlike `futures`, this function uses its own thread instead of the existing async runtime.

> The first time [Async](https://docs.rs/async-io/latest/async_io/struct.Async.html) or [Timer](https://docs.rs/async-io/latest/async_io/struct.Timer.html) is used, a thread named “async-io” will be spawned. - [API doc](https://docs.rs/async-io/latest/async_io/index.html#implementation)

oguzkocer: My understanding is that `sleep` is a run-time dependent action unless we put the whole thread to sleep, which is not what we want. Considering we are working with different platforms and don't want to rely on anything runtime dependent, I think the easiest and for sure method we can use here is to add a `sleep` function to `RequestExecutor` trait and let each platform provide their own implementation. This should be super easy to implement in all 3 platforms where we have specific runtime executors and it'll keep `wp_api` runtime agnostic.

At some point we might want to extract this new `sleep` function to a new utility async trait where we might add more helpers. However, I don't think there is any need to immediately introduce a new trait.

jkmassel:
Replaced by #585

PR #574, opened 2025-02-22
Localization support

oguzkocer:
This PR introduces localization support, building on the initial prototype from https://github.com/Automattic/wordpress-rs/pull/488.

* `wp_localization_macro::wp_messages` generates bindings from the main `ftl` file, so we can use static functions instead of string keys.
* `wp_localization_macro::WpDeriveLocalizable` derive macro implements the `Display` trait if the type implements `wp_localization::WpSupportsLocalization` trait. It also generates `uniffi` bindings to localize the type using a free-floating function with the name `localize_{type_name}`.
* Implementing `wp_localization::WpSupportsLocalization` is how we'll bind a type to a localized message using the bindings generated in `wp_localization::WpMessages`.
* `wp_localization` depends on `fluent`, but it provides a tool agnostic API to `wp_api`, so we can replace `fluent` without changing anything about how we localize our types in `wp_api`.

Here is an example usage:

```rust
#[derive(Debug, Clone, thiserror::Error, uniffi::Error, WpDeriveLocalizable)]
pub enum ParseApiRootUrlError {
    ApiRootLinkHeaderNotFound {
        status_code: u16,
        header_map: Arc<WpNetworkHeaderMap>,
    },
}

impl WpSupportsLocalization for ParseApiRootUrlError {
    fn message_bundle(&self) -> wp_localization::MessageBundle {
        match self {
            ParseApiRootUrlError::ApiRootLinkHeaderNotFound {
                status_code,
                header_map,
            } => WpMessages::api_root_link_header_not_found(
                status_code.to_string(),
                format!("{:#?}", header_map),
            ),
        }
    }
}
```

Notice how we are no longer using the `#[error("..")` attribute from `thiserror` which would implement the `Display` trait. Instead, we only implement `WpSupportsLocalization` and add the `WpDeriveLocalizable`.

Once this is implemented, here are few usage options:
```rust
// Rust
let error = ParseApiRootUrlError::ApiRootLinkHeaderNotFound {
    status_code: 404,
    header_map: some_header_map(),
};
let string_with_default_locale = error.to_string();
let string_with_default_locale = error.localize(None);
let string_with_specified_locale = error.localize(Some(wp_localization::WpLocale::TrTR));
```
```kotlin
// Kotlin
val error = ParseApiRootUrlException.ApiRootLinkHeaderNotFound(
    statusCode = 404u,
    headerMap = WpNetworkHeaderMap.fromMap(mapOf())
)
val stringWithDefaultLocale = localizeParseApiRootUrlError(error, locale = null)
val stringWithSpecifiedLocale = localizeParseApiRootUrlError(error, locale = WpLocale.TR_TR)
```

I believe this is a great starting point for localization. As far as I can tell, it should be safe to use this in production. Having said that, there is still quite a bit room for improvement. There are `fluent` types we don't handle at all, we could use more tests, the bindings don't seem to be generated after `ftl` file is updated until `cargo build`, (although there is a build script that tries to do that) and uniffi bindings are free-floating functions. (even though initially they were contained in a `UniffiLocalizable` type) If I don't call this done here I feel like I might spend another few days on it and probably won't have enough to show for it. To keep things efficient, I'd like to re-visit these topics and any other improvements on a need-to basis.


wp_localization/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        locales: "./localization",
+        fallback_language: "en-US"
+    };
+}
+
+#[wp_localization_macro::wp_messages]
+pub struct WpMessages {}
+
+#[derive(Debug)]
+pub struct MessageBundle {
```
crazytonyli, line 17: The term "Bundle" gives the impression that it represents a list of many messages. Maybe something like `LocalizableMessage` would be better?

wp_localization/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    // Keep the fields private to avoid clients from looking up messages that do not exist.
+    key: &'static str,
+    args: Option<HashMap<&'static str, String>>,
```
crazytonyli, line 20: Use [`FluentValue`](https://docs.rs/fluent/latest/fluent/enum.FluentValue.html) instead of String. I believe it's used to support other fluent syntax (like plural).

native/kotlin/api/kotlin/src/integrationTest/kotlin/LocalizationTest.kt
```diff
@@ -0,0 +1,27 @@
...
+            statusCode = 404u,
+            headerMap = WpNetworkHeaderMap.fromMap(mapOf())
+        )
+        assertEquals(
+            "Api root link header not found!\nStatus Code: '\u2068404\u2069'\nHeader Map: '\u2068WpNetworkHeaderMap {\n    inner: {},\n}\u2069'",
+            localizeParseApiRootUrlError(error, locale = null)
+        )
+        assertEquals(
+            "Api kök bağlantı başlığı bulunamadı!\nDurum kodu: '\u2068404\u2069'\nBaşlık Haritası: '\u2068WpNetworkHeaderMap {\n    inner: {},\n}\u2069'",
+            localizeParseApiRootUrlError(error, locale = WpLocale.TR_TR)
```
crazytonyli, line 24: I don't think we should use a strong type `WpLocale` in the public API. At least, we should have a public API that accepts a language id string, which Rust maps to an existing `<locale>.ftl` (either the "correct one" or the default one): `WPLocal.from_language_id(string: String) -> WPLocal`. Otherwise, Swift & Kotlin would need to implement this mapping themselves.

crazytonyli: BTW, this conversion exists in my prototype PR.

wp_localization/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    #[strum(serialize = "tr-TR")]
+    TrTR,
+}
+
+impl WpLocale {
+    pub fn as_language_id(&self) -> unic_langid::LanguageIdentifier {
+        Into::<&str>::into(self).parse().expect(
+            "All locales are unit tested to ensure they can be converted to LanguageIdentifier",
+        )
+    }
```
crazytonyli, line 86: I think we should be able to use the `langid!` macro to avoid parsing the same string repeatedly at runtime. For example, generating the `WpLocale` type in the build script and adding a `fn as_language_id() -> &'static LanguageIdentifier`.

wp_localization/src/lib.rs
```diff
@@ -0,0 +1,199 @@
+use fluent_bundle::FluentValue;
+use fluent_templates::Loader;
+use std::{collections::HashMap, fmt::Debug, fmt::Display};
+use strum_macros::{EnumIter, IntoStaticStr};
+
+fluent_templates::static_loader! {
```
crazytonyli, line 6: I was hoping we could implement this ourselves so that we can get rid of this third-party library. I guess it could be one of the next steps.

wp_api/src/login/url_discovery.rs
```diff
@@ -606,33 +608,57 @@ pub(crate) fn construct_attempts(input_site_url: String) -> Vec<AutoDiscoveryAtt
...
+    }
+}
+
+#[derive(Debug, thiserror::Error, uniffi::Error, WpDeriveLocalizable)]
 pub enum FetchApiDetailsError {
-    #[error(
-        "Request execution failed!\nStatus Code: '{:?}'.\nResponse: '{}'",
-        status_code,
-        reason
-    )]
```
crazytonyli, line 628: It would be nice to implement the same syntax like `#[localized_error(...)]`, to replace the `impl WpSupportsLocalization for ...`

native/kotlin/api/kotlin/src/integrationTest/kotlin/LocalizationTest.kt
oguzkocer: That conversion will be just a `uniffi` exported function with the current setup. This PR doesn't include it because there is no test case implemented for it yet and I wanted to wrap up the initial PR. It's a simple enough addition as a follow up.

wp_localization/src/lib.rs
oguzkocer: How about `MessageComponents`? I think `MessageBundle` is the correct term here because `Bundle` represents the components for a single message, but I can see how it can be interpreted in the way you mentioned.

I don't mind `LocalizableMessage` either if you strongly prefer it, but I think that gives the impression that this is a final "product" type, not a builder type.

wp_localization/src/lib.rs
oguzkocer: This might be a little tricky to do. Let me think about it.

wp_localization/src/lib.rs
oguzkocer: I missed the fact that `langid!` can be used to create constants. Good catch, I'll address this tomorrow.

wp_localization/src/lib.rs
oguzkocer: I haven't looked into this at all. So, although I agree with you that it'd be good not to rely on a 3rd party, I don't think it's very impactful to take this on at the moment.

wp_api/src/login/url_discovery.rs
oguzkocer: I've considered this and an initial version of this might be doable in a reasonable time, but I felt that it might get annoying to maintain it. Because, unlike `thiserror`, this is a very small part of our project.

Furthermore, I am not sure if there is enough value for such an implementation in this case. In my opinion, it's not any simpler to use than manually implementing `WpSupportsLocalization` and whenever we need a helper to build the string, it's actually easier to include that in a manual `WpSupportsLocalization` implementation.

So, at the very least, this is outside of the scope of this PR and I don't think I'll be tackling it unless I run out of things to do for when I am blocked.

crazytonyli (approved):

oguzkocer:
@crazytonyli The PR got auto-merged, but I intend to open a follow up PR with your feedback. I am AFK for a bit, so it might be a few days.

PR #584, opened 2025-02-27
Add HTTP Header Parsing Helpers

jkmassel:
Adds some helpers for parsing 401, 403, and 429 responses

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,1 @@
+large-error-threshold = 192
--- a/wp_api/src/api_error.rs
+++ b/wp_api/src/api_error.rs
@@ -1,3 +1,4 @@
+use crate::request::{HttpAuthMethod, HttpAuthMethodParsingError};
 use crate::{
     request::{request_or_response_body_as_string, WpRedirect},
     ssl::SSLCertificateInfo,
@@ -450,20 +451,22 @@ pub enum RequestExecutionErrorReason {
     },
     #[error(
         "The server at {} requires authentication. Please provide your username and password.",
-        url
+        hostname
     )]
     HttpAuthenticationRequiredError {
-        url: String,
+        hostname: String,
         server_message: Option<String>,
     },
     #[error(
         "The server at {} rejected your credentials. Please provide a valid username and password.",
-        url
+        hostname
     )]
     HttpAuthenticationRejectedError {
-        url: String,
-        server_message: Option<String>,
+        hostname: String,
+        method: Option<HttpAuthMethod>,
     },
+    #[error("The server is sending invalid HTTP authentication information. Please check your site's HTTP authentication configuration.")]
+    MisconfiguredHttpAuthenticationError { issue: HttpAuthMethodParsingError },
     #[error("The server is rate limiting requests in a way that will never succeed. Please check your site's rate limit configuration.")]
     MisconfiguredRateLimitError {},
     #[error("{}", error_message)]
@@ -520,6 +523,18 @@ impl From<MediaUploadRequestExecutionError> for WpApiError {
             MediaUploadRequestExecutionError::MediaFileNotFound { file_path } => {
                 Self::MediaFileNotFound { file_path }
             }
+        }
+    }
+}
+
+impl From<HttpAuthMethodParsingError> for WpApiError {
+    fn from(value: HttpAuthMethodParsingError) -> Self {
+        WpApiError::RequestExecutionFailed {
+            status_code: None,
+            redirects: None,
+            reason: RequestExecutionErrorReason::MisconfiguredHttpAuthenticationError {
+                issue: value,
+            },
         }
     }
 }
--- a/wp_api/src/request.rs
+++ b/wp_api/src/request.rs
@@ -783,15 +783,15 @@ mod tests {
     }
 
     #[rstest]
-    #[case("Basic realm=\"example\", charset=\"UTF-8\"", HttpAuthMethod::Basic(BasicAuthenticationDetails {
+    #[case(r#"Basic realm="example", charset="UTF-8""#, HttpAuthMethod::Basic(BasicAuthenticationDetails {
         realm: Some("example".to_string()),
         requires_utf8: true,
     }))]
-    #[case("Basic realm=\"example\"", HttpAuthMethod::Basic(BasicAuthenticationDetails {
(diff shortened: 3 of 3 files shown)


wp_api/src/request.rs
```diff
@@ -15,7 +18,6 @@ pub mod endpoint;
 
 const CONTENT_TYPE_JSON: &str = "application/json";
 const CONTENT_TYPE_MULTIPART: &str = "multipart/form-data";
-const LINK_HEADER_KEY: &str = "Link";
```
jkmassel, line 18: We can use constants from `http::header`, which will avoid cluttering up this file with a bunch of extras (I'm using two more header names in this PR, and figured I'd align the older code while I was at it)

wp_api/src/request.rs
```diff
@@ -278,6 +288,11 @@ impl WpNetworkHeaderMap {
             .and_then(|h| h.parse().ok())
     }
 
+    pub fn insert(&mut self, header_name: HeaderName, header_value: String) {
```
jkmassel: Added because `from_map` doesn't properly handle cases where the header value (validly) contains a comma

.clippy.toml
```diff
@@ -0,0 +1 @@
+large-error-threshold = 192
```
jkmassel, line 1: I'm not super happy about this, but I'm not sure we can easily avoid it either? Wrapping everything in `Arc<T>` is kinda lousy, and we don't have a deep error stack – each of these will be allocated once per request, which is IMHO reasonable?

crazytonyli: One alternative could be to use the `GenericError` for all errors that we don't expect apps to handle. By "handle," I mean the app may fall back to another path when the error happens. For example, the app can't do anything about the `DeviceIsOfflineError` error other than present an error message. We can remove it and use `GenericError` instead.

Now that we have `wp_localization`, we can change `GenericError(String)` to be `GenericError(MessageBundle)`, so that the error message is localized in Rust.

wp_api/src/request.rs
```diff
@@ -593,6 +780,83 @@ mod tests {
         assert_header_map_values(&header_map, "User-Agent", vec![]);
     }
 
+    #[rstest]
+    #[case("Basic realm=\"example\", charset=\"UTF-8\"", HttpAuthMethod::Basic(BasicAuthenticationDetails {
+        realm: Some("example".to_string()),
+        requires_utf8: true,
+    }))]
```
crazytonyli, line 789: You can use `r#"Basic realm="example"..."#` to avoid escaping quotes `\"`

wp_api/src/api_error.rs
```diff
@@ -450,20 +451,22 @@ pub enum RequestExecutionErrorReason {
...
+        hostname
     )]
     HttpAuthenticationRejectedError {
-        url: String,
-        server_message: Option<String>,
+        hostname: String,
+        method: Option<HttpAuthMethod>,
     },
+    #[error("The server is sending invalid HTTP authentication information. Please check your site's HTTP authentication configuration.")]
+    MisconfiguredHttpAuthenticationError { issue: HttpAuthMethodParsingError },
```
crazytonyli, line 469: Maybe we don't have to pre-parse the `www-authentication` header? Instead, we can include the header string in the error variant and expose an API to parse the string to `Result<HttpAuthMethod, HttpAuthMethodParsingError>`. By doing that, we can remove this `MisconfiguredHttpAuthenticationError` variant.

wp_api/src/request.rs
```diff
@@ -278,6 +288,11 @@ impl WpNetworkHeaderMap {
             .and_then(|h| h.parse().ok())
     }
 
+    pub fn insert(&mut self, header_name: HeaderName, header_value: String) {
+        self.inner
+            .insert(header_name, header_value.parse().unwrap());
```
crazytonyli: Return an `Result`, instead of `unwrap`? Or, we can make this function only available to unit tests, which seems to be the only place it's currently used.

.clippy.toml
jkmassel: The `DeviceIsOfflineError` is there because the app might want to do something with the type information.

We could reasonably replace it with a `type` field on `GenericError`, and that could help. 

I'm just not sure how much we want to de-specialize our error types? I worry we'll just end up overloading `GenericError` with a ton of extra data and end up right back where we started.

wp_api/src/api_error.rs
jkmassel: Yeah, this is a weird one – the reason it exists is because we can't cover the entire world of HTTP Auth responses we might get. We could defer to the client for this, but we're modelling the rest of the `400` error space, so it seems odd to leave this one case out.

wp_api/src/request.rs
jkmassel: Sure, I've moved it into the test mod. If we move it back out, it might make sense to have `header_value` be `std::ascii` to avoid the need for the `Result`?

wp_api/src/request.rs
```diff
@@ -439,6 +449,175 @@ impl WpNetworkResponse {
...
+            .map(|d| d.seconds)
+    }
+
+    pub fn is_http_authentication_required(&self) -> bool {
+        self.status_code == 401 || self.status_code == 403
+    }
+
+    pub fn get_http_auth_method(
+        &self,
+    ) -> Result<Option<HttpAuthMethod>, HttpAuthMethodParsingError> {
```
crazytonyli, line 472: `Option<Result<HttpAuthMethod, HttpAuthMethodParsingError>>` might make more sense? `None` means the header does not exists and `Some(result)` means we tried to parse the header. Or, just return `Result<...>` type by adding a "missing header" to the error?

wp_api/src/request.rs
```diff
@@ -439,6 +449,175 @@ impl WpNetworkResponse {
...
+impl FromStr for HttpAuthMethod {
+    type Err = HttpAuthMethodParsingError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts = s.splitn(2, ' ').collect::<Vec<&str>>();
+
+        let method = parts[0];
+
+        let mut details = HashMap::new();
+        parts[1].split(',').for_each(|part| {
```
crazytonyli, line 568: It's possible that `parts` contain less than two elements. In that case, would `parts[0]` and `parts[1]` would panic? Maybe we should do some array length check beforehand?

crazytonyli (approved):

PR #585, opened 2025-02-28
Add Middleware

jkmassel:
Adds a middleware pipeline to the `LoginClient` and `WpOrg` client.

This doesn't yet exist for the `WpApiClient`. While it's probably applicable and would work great, we're leaving that for another day so that we can ship login.

The middleware pipeline is a common solution to avoid tightly coupling a lot of logic in the library, while still retaining a "batteries included" philosophy. This is similar to other API libraries like:

[OkHttp](https://square.github.io/okhttp/features/interceptors/)
[Guzzle](https://docs.guzzlephp.org/en/stable/handlers-and-middleware.html#middleware)
[Alamofire](https://github.com/Alamofire/Alamofire/blob/master/Documentation/AdvancedUsage.md#adapting-and-retrying-requests-with-requestinterceptor)

We might want to add more later, but for now, there are two middlewares bundled with the library:
- HTTP 429 auto-retry
- HTTP Basic Authentication

Kotlin and Swift both also have a "debug" middleware that'll print requests and responses to the log. This is useful if developers want to put a breakpoint in the request process, but also works as a template to demonstrate how to implement native middleware if needed.


wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+
+#[derive(Debug, uniffi::Object)]
+pub struct WpApiMiddlewarePipeline {
+    inner: Arc<_WpApiMiddlewarePipeline>,
+}
+
+#[derive(Debug)]
+struct _WpApiMiddlewarePipeline {
+    middlewares: Vec<Arc<dyn WpApiMiddleware>>,
+}
```
crazytonyli, line 10: This struct seems redundant. Is there any reason not to use `Vec<Arc<dyn WpApiMiddleware>>` in `WpApiMiddlewarePipeline` directly?

wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    pub async fn process(
+        &self,
+        request_executor: Arc<dyn RequestExecutor>,
+        response: WpNetworkResponse,
+        request: Arc<WpNetworkRequest>,
+    ) -> Result<WpNetworkResponse, WpApiError> {
+        let mut response = response;
+
+        for middleware in self.inner.middlewares.iter() {
```
crazytonyli: I think `.iter()` is redundant here?

wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+
+        Ok(response)
+    }
+}
+
+impl Default for WpApiMiddlewarePipeline {
+    fn default() -> Self {
+        Self::new(Vec::new())
+    }
+}
```
crazytonyli: I think you can replace this manual implementation with a `#[derive(Default)]` macro

wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+
+#[derive(Debug)]
+struct InnerPipelineBuilder {
+    middlewares: Vec<Arc<dyn WpApiMiddleware>>,
+}
+
+#[uniffi::export]
+impl WpApiMiddlewarePipelineBuilder {
```
crazytonyli: I don't feel like this type is super useful? It's building a list, which is pretty easy to do in Rust or native code?

wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+}
+
+fn async_sleep(duration: std::time::Duration) -> impl Future<Output = ()> {
+    let (tx, rx) = futures::channel::oneshot::channel::<i32>();
+
+    thread::spawn(move || {
+        thread::sleep(duration);
+        let _ = tx.send(0); // ignore error because it is perfectly ok when the receiver is dropped
+    });
```
crazytonyli: It _feels_ a bit wasteful to create a thread just to let it sleep and do nothing. What do you think about delegating this sleep to the native platforms? 

For example:

```rust
#[uniffi::foreight_trait]
trait RetryMiddlewareSupport {
  async sleep(duration: Duration)
}

struct RetryMiddleware {
  ...
  support: Arc<dyn RetryMiddlewareSupport>
}
```

Native platforms can easily implement `RetryMiddlewareSupport` and pass an instance to `RetryMiddle`'s initializer.

wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+        &self,
+        request_executor: Arc<(dyn RequestExecutor + 'static)>,
+        response: WpNetworkResponse,
+        request: Arc<WpNetworkRequest>,
+    ) -> Result<WpNetworkResponse, WpApiError> {
+        if !response.is_http_authentication_required() {
+            return Ok(response);
+        }
+
+        let new_request = request.adding_http_authentication(&self.username, &self.password);
```
crazytonyli: When using `HttpAuthenticationMiddleware` (instead of adding the header when building requests), we would send two requests for endpoints that require authentication: one without the authentication header and the other with the authentication header.

What do you think about allowing handling outgoing requests, too, so that we can insert headers before sending requests?

crazytonyli: Also, it's best to check request.url first before adding the header. We don't want to send the application password to any random sites.

wp_api/src/request.rs
```diff
@@ -213,12 +214,33 @@ impl WpNetworkRequest {
...
+        );
+
+        WpNetworkRequest {
+            uuid: self.uuid.clone(),
+            retry_count: self.retry_count,
+            method: self.method.clone(),
+            url: self.url.clone(),
+            header_map: Arc::new(new_header_map.into()),
+            body: self.body.clone(),
+        }
```
crazytonyli, line 242: We should be able to avoid all these clones by making this function take `self`'s ownership.

wp_api/src/middleware.rs
jkmassel: This is probably the way to go for sure

wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+
+            if response.is_rate_limit_exceeded() && self.max_retries != 0 {
+                let retry_after = response.get_retry_after();
+
+                if let Some(mut retry_after) = retry_after {
+                    // If the server sends some super-long value, we don't want to wait that long
+                    if retry_after > self.max_retry_wait_seconds {
+                        retry_after = self.max_retry_wait_seconds;
+                        async_sleep(Duration::from_secs(retry_after)).await;
+                    }
```
oguzkocer: This would only sleep if `retry_after` is longer than `max_retry_wait_seconds`. In my suggestion PR, I am changing this as:

```
request_executor.sleep(
    // If the server sends some super-long value, we don't want to wait that long
    Duration::from_secs(std::cmp::min(
        retry_after,
        self.max_retry_wait_seconds,
    ))
    .as_millis(),
);
```

Since this includes a logic change, I wanted to leave a note just in case I am misunderstanding something.

wp_api/src/middleware.rs
```diff
@@ -0,0 +1,364 @@
...
+    async fn process(
+        &self,
+        request_executor: Arc<(dyn RequestExecutor + 'static)>,
+        response: WpNetworkResponse,
+        request: Arc<WpNetworkRequest>,
+    ) -> Result<WpNetworkResponse, WpApiError> {
+        let mut retry_count = 0;
+        let mut response = response;
+
+        loop {
```
oguzkocer: I think there are quite a few logic errors in this middleware:

1. We return an error in `retry_count >= self.max_retries` case without checking the response. This means if the last response was `200`, we end up turning that into a `RequestExecutionErrorReason::MisconfiguredRateLimitError` error.
2. We only increase the `retry_count` if `response.is_rate_limit_exceeded() && self.max_retries != 0`. That means if we get any response status other than `429` it'll end up in an infinite loop. I don't see how this retry middleware can ever successfully return a response.
3. The middleware only gets processed when there is a `WpNetworkResponse` - as opposed to `Result<WpNetworkResponse, RequestExecutionError>` - and it checks if the response status is `429`, but I believe the request executors will return `Err(RequestExecutionError)` in such a case, so I don't think this middleware would ever get processed. This heavily depends on the request executor implementation which should be something we should strongly avoid.

wp_api/src/middleware.rs
jkmassel: Uniffi doesn't nicely handle a collection of heterogenous objects that all conform to a protocol, which is why this exists

wp_api/src/middleware.rs
jkmassel: Yeah, when I implemented `async_sleep`, I think I dropped the reference to it one line too high – it should be where the commented out code is below

native/swift/Sources/wordpress-api/Extensions.swift
```diff
@@ -0,0 +1,28 @@
...
+    }
+}
+
+extension WpNetworkResponse {
+    init(data: Data, request: WpNetworkRequest, response: URLResponse) throws {
+        guard let response = response as? HTTPURLResponse else {
+            preconditionFailure("We should never wind up here")
+        }
+
+        self =  WpNetworkResponse(
```
crazytonyli, line 20: This is an extra whitespace after `=`.

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -158,6 +151,16 @@ final class WpRequestExecutor: SafeRequestExecutor {
...
     }
     // swiftlint:enable force_cast
+
+    // swiftlint:disable force_try
+    func sleep(millis: UInt64) async {
+        if #available(macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0, *) {
+            try! await Task.sleep(for: .milliseconds(millis))
+        } else {
+            try! await Task.sleep(nanoseconds: millis * 1000)
+        }
```
crazytonyli, line 161: I think it's best to `try?` here. I believe `sleep` throws a `CancellationError`, when the task is cancelled while `sleep`ing.

jkmassel: Yep that's right! The underlying protocol doesn't return a `Result` type so we can't make this function `throw`-ing, but it's on the list to fix this up for sure

oguzkocer (approved):

PR #586, opened 2025-03-03
Implement Jetpack connection flow

crazytonyli:
The new Jetpack connection endpoints (see pf5801-1q1-p2) are in the Jetpack plugin and WP.com API. They are implemented in the two new crates (jetpack & wp_com).

Other than following the existing patterns in `wp_api` to implement the new endpoints, I added a `JetpackConnectionClient` in the `jetpack` crate. It provides two high-level functions `connect_site` and `connect_user` to perform Jetpack connection flows, which use the new endpoints internally.


jetpack/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "jetpack"
+version = "0.1.0"
+edition = "2021"
```
oguzkocer: We've updated the edition of all other packages to `2024` recently.

wp_api/src/request/endpoint.rs
```diff
@@ -78,6 +78,10 @@ impl TryFrom<&str> for ApiBaseUrl {
 }
 
 impl ApiBaseUrl {
+    pub fn with_api_url(url: Url) -> Self {
```
oguzkocer, line 87: We probably should drop the `impl From<Url> for ApiBaseUrl` implementation because having both of these might create confusion.

Looking back, I think the way `impl From<Url> for ApiBaseUrl` implemented wasn't a good idea in the first place. The `ApiBaseUrl` only has a `Url` field, so the expectation for `impl From<Url> for ApiBaseUrl` would be to set that field directly, instead of modifying it the way it's done at the moment.

Let me know if you want to address that in this PR or if you'd rather I open a PR against `trunk`.

wp_com/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "wp_com"
+version = "0.1.0"
+edition = "2021"
```
oguzkocer: We've updated the edition of all other packages to `2024` recently.

jetpack/src/connection.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub status: String,
+}
+
+/// Connect a WordPress site to Jetpack.
+///
+/// Please note, the endpoints used are available on Jetpack 14.2 and above.
+#[derive(Debug, uniffi::Object)]
+pub struct JetpackConnectionClient {
+    request_executor: Arc<dyn RequestExecutor>,
+    jetpack_client: JetpackApiClient,
```
oguzkocer, line 90: This setup seems a little backwards to me. Shouldn't we add `JetpackConnectionClient` to `JetpackApiClient` instead? Instead of `JetpackApiClient`, this type should probably take a `ConnectionRequestExecutor` and we should remove the `connection: Arc<ConnectionRequestExecutor>` field from `JetpackApiClient`.

All these suggestions are under the assumption that we are not supposed to use the `ConnectionRequestExecutor` directly, but rather use the `JetpackConnectionClient` because it provides further processing. Let me know if that's a misunderstanding.

oguzkocer (commented):
Looks great! I only left one important suggestion/question about the relationship between `JetpackConnectionClient` and `JetpackApiClient`, and a few minor ones.

jetpack/src/connection.rs
crazytonyli: My thinking is that `JetpackApiClient` and `WpApiClient` are for accessing individual endpoints, and `JetpackConnectionClient` is a higher-level API that connects sites to Jetpack which internally uses `JetpackApiClient` to access Jetpack endpoints and potentially uses `WpApiClient` too to access the core endpoints.

The type names here are probably confusing. The connection in `JetpackConnectionClient` indicates the type is for connecting a site to Jetpack, where as the connection in `JetpackApiClient.connect()` (a.k.a `ConnectionRequestExecutor`) means that the type implements endpoints under `jetpack/v4/connection`.

wp_api/src/request/endpoint.rs
crazytonyli: It's best to refactor it in a separate PR, which I can look into later.

jetpack/src/connection.rs
oguzkocer: I could see either approach making sense. I always thought `JetpackApiClient` and `WpApiClient` would be the overarching container, but it can also be treated as the clients that allow raw requests as you've done so here.

I am fine with leaving it as is for now and we can re-consider it if we find any other pros/cons to either approach.

wp_com/Cargo.toml
oguzkocer: @crazytonyli Looks like you might have missed updating this in 8ffd50cb3e94cc5c55541c16f2f3bc1f58dce35d.

oguzkocer (approved):
Approving, so it can be merged by @crazytonyli once `wp_com` crate's edition is updated to `2024`.

wp_com/Cargo.toml
crazytonyli: 🙈 Addressed in https://github.com/Automattic/wordpress-rs/pull/586/commits/d7e68b148b2da254770c0d271940e1e10ae394d5

PR #594, opened 2025-03-04
uniffi::Error types conform to LocalizedError

crazytonyli:
All uniffi::Error types now implement the `WpSupportsLocalization` trait. In Swift bindings, they all conform to the `LocalizedError` protocol.

This PR changes are mostly based on what we have at the moment: just adding more message keys to the en-US/main.ftl file and implementing more `WpSupportsLocalization`.

There is one refactor on the `WpLocale`: changing it from enum to a struct, which is exported as String in native bindings. Native platforms can now pass any arbitrary string (typically from Android/iOS system API), and `WpLocale` will resolve them into a language that the library supports.


native/swift/Tests/wordpress-api/LocalizationTests.swift
```diff
@@ -0,0 +1,16 @@
...
+import WordPressAPIInternal
+
+class LocalizationTests {
+    @Test
+    func testParsingError() {
+        do {
+            let _ = try ParsedUrl.parse(input: "not-url")
+            Issue.record("Got an unexpected successful result")
+        } catch {
+            #expect(error.localizedDescription == "URL is invalid")
```
crazytonyli: Just in case it's unclear, this is this PR's end goal, where `error.localizedDescription` returns error messages for app users.

wp_api/src/api_error.rs
```diff
@@ -11,39 +13,30 @@ where
     fn as_parse_error(reason: String, response: String) -> Self;
 }
 
-#[derive(Debug, PartialEq, Eq, thiserror::Error, uniffi::Error)]
+#[derive(Debug, PartialEq, Eq, uniffi::Error, WpDeriveLocalizable)]
```
crazytonyli: I had to delete `thiserror::Error` because of some compiling errors. But there are no more compiling errors when keeping the `thiserror::Error` macro. I'm not exactly sure what has changed, but I'm happy to add the macro back if needed.

wp_api/src/login/url_discovery.rs
```diff
@@ -411,47 +433,53 @@ pub enum FetchAndParseApiRootFailure {
...
+                .unwrap_or(WpMessages::application_passwords_not_supported()),
         }
     }
 }
 
-#[derive(Debug, Clone, uniffi::Error)]
+#[derive(Debug, Clone, uniffi::Error, WpDeriveLocalizable)]
 pub enum ApplicationPasswordsNotSupportedReason {
     ApplicationPasswordBlockedByPlugin {
+        site_url: Arc<ParsedUrl>,
```
oguzkocer: I specifically moved the common fields out of these sub error variants and this PR reverts that without any benefits. I'd like to stick to the single source of truth principle here which should be easy to do so by following the design in current `trunk`.

Also a little bit of general feedback; Making these types of changes as part of unrelated PRs can be easy to miss especially if they are not specifically brought up in the PR description.

wp_api/src/login/url_discovery.rs
```diff
@@ -411,47 +433,53 @@ pub enum FetchAndParseApiRootFailure {
...
         plugin: KnownApplicationPasswordBlockingPlugin,
     },
-    ApplicationPasswordBlockedByMultiplePlugins,
+    ApplicationPasswordBlockedByMultiplePlugins {
+        site_url: Arc<ParsedUrl>,
```
oguzkocer: See my previous comment about moving the common fields out of sub error variants.

wp_api/src/login/login_client.rs
```diff
@@ -187,11 +187,12 @@ impl WpLoginClient {
                 if plugins.len() == 1 {
                     // If there's only one candidate, we can show more information in the error message
                     Some(ApplicationPasswordsNotSupportedReason::ApplicationPasswordBlockedByPlugin {
+                        site_url: parsed_site_url.clone(),
```
```suggestion
```

wp_api/src/login/login_client.rs
```diff
@@ -187,11 +187,12 @@ impl WpLoginClient {
...
                         plugin: plugins.first().expect("Already verified there is one plugin").clone(),
                     })
                 } else {
                     // If there's more than one, for now we'll just give a generic error
-                    Some(ApplicationPasswordsNotSupportedReason::ApplicationPasswordBlockedByMultiplePlugins)
+                    Some(ApplicationPasswordsNotSupportedReason::ApplicationPasswordBlockedByMultiplePlugins { site_url: parsed_site_url.clone() })
```
```suggestion
                    Some(ApplicationPasswordsNotSupportedReason::ApplicationPasswordBlockedByMultiplePlugins)
```

wp_api/src/parsed_url.rs
```diff
@@ -45,32 +47,27 @@ impl ParsedUrl {
...
     RelativeUrlWithCannotBeABaseBase,
-    #[error("a cannot-be-a-base URL doesn’t have a host to set")]
     SetHostOnCannotBeABaseUrl,
-    #[error("URLs more than 4 GB are not supported")]
     Overflow,
 }
 
+impl WpSupportsLocalization for ParseUrlError {
+    fn message_bundle(&self) -> MessageBundle {
+        WpMessages::url_parsing_error()
```
oguzkocer, line 77: Is this an accidental change? It seems to be stripping importing information about the error variants.

wp_localization/localization/en-US/main.ftl
```diff
@@ -1,10 +1,65 @@
...
+
+http_authentication_rejected_error = The server at {$url} rejected your credentials. Please provide a valid username and password.
+
+misconfigured_http_authentication_error = The server is sending invalid HTTP authentication information. Please check your site's HTTP authentication configuration
+
+misconfigured_rate_limit_error = The server is rate limiting requests in a way that will never succeed. Please check your site's rate limit configuration
+
+oauth_response_url_error_missing_site_url = The given URL is missing the `site_url` query parameter
+oauth_response_url_error_missing_username = The given URL is missing the `username` query parameter
+oauth_response_url_error_missing_password = The given URL is missing the `password` query parameter
```
oguzkocer: We can use a single key for this.

wp_localization/Cargo.toml
```diff
@@ -10,11 +10,11 @@ uniffi = { workspace = true , features = [ "build", "cli" ] }
 fluent-bundle = { workspace = true }
 fluent-langneg = { workspace = true }
 fluent-templates = { workspace = true }
-icu_locid = { workspace = true }
-thiserror = { workspace = true }
 paste = { workspace = true }
+rstest = { workspace = true }
```
oguzkocer: This should be a dev dependency.

wp_localization/localization/en-US/main.ftl
```diff
@@ -1,10 +1,65 @@
-api_root_link_header_not_found = Api root link header not found!
-  Status Code: '{$status_code}'
-  Header Map: '{$header_map}'
+api_root_link_header_not_found = WordPress REST API link is not found in the site response
```
oguzkocer: This is an unused message as it was recently dropped in `trunk`.

wp_localization/localization/en-US/main.ftl
```diff
@@ -1,10 +1,65 @@
...
 
-fetch_api_details_request_execution_failed = Request execution failed!
-  Status Code: '{$status_code}'
-  Reason: '{$reason}'
+generic_error = Something went wrong
```
oguzkocer: I think we should rename this to `wp_api_error_generic_error` because generic error messages may depend on context.

wp_localization/localization/en-US/main.ftl
```diff
@@ -1,10 +1,65 @@
...
+non_existent_site_error = A server with the specified hostname could not be found
+
+http_authentication_required_error = The server at {$url} requires authentication. Please provide your username and password
+http_forbidden_error = The server at {$url} denied access to the requested resource. Please check your site's configuration
+
+http_authentication_rejected_error = The server at {$url} rejected your credentials. Please provide a valid username and password.
+
+misconfigured_http_authentication_error = The server is sending invalid HTTP authentication information. Please check your site's HTTP authentication configuration
+
+misconfigured_rate_limit_error = The server is rate limiting requests in a way that will never succeed. Please check your site's rate limit configuration
```
oguzkocer: Most of these are missing `.` at the end. There are also others messages in the rest of the file.

I _think_ there is supposed to be a `.` at the end if there is another sentence ending with punctuation (`.`, `!` etc) within the message, because otherwise it looks incomplete. In other cases, I think it depends on the context and other messages used for that error, but in any case, we should at least fix the ones with multiple sentences as I believe they were (or most of them were) correct in the original message.

wp_localization/localization/tr-TR/main.ftl
```diff
@@ -1,3 +1,71 @@
-api_root_link_header_not_found = Api kök bağlantı başlığı bulunamadı!
-  Durum kodu: '{$status_code}'
-  Başlık Haritası: '{$header_map}'
+api_root_link_header_not_found = WordPress REST API bağlantısı site yanıtında bulunamadı
```
oguzkocer: This message was removed recently as noted in the main translation file.

I am not going to review the rest of this file as I assume it's AI translated. Probably worth doing another pass once the changes in the `main.ftl` are finalized.

oguzkocer (changes requested):

wp_api/src/login/url_discovery.rs
crazytonyli: The reason for adding this URL is not to revert anyone's changes. It's simply because I'd like to preserve the existing `error_message` messages in the new `message_bundle` function, which would require the site URL. Since we can't add a URL argument to the `message_bundle` function, adding an URL to the error variant is the only way that I can think of. Let me know if there are other solutions though.

wp_api/src/parsed_url.rs
crazytonyli: Considering this message is displayed on the UI, I don't think we need to be too specific about the technical reasons, like "host is missing," "scheme is missing", etc. I thought a generic "URL is invalid" should be enough for the users to correct the URL, like if they missed a `:` or a `/` in the `https://`.

wp_localization/localization/en-US/main.ftl
crazytonyli: Good catch. I have added `.` to all messages expect the `just` one, which is hack to return the input argument as "localized" message.

wp_localization/localization/tr-TR/main.ftl
crazytonyli: I was hoping you could review them because they are AI translated, to examine the quality of the AI translation.

wp_api/src/login/url_discovery.rs
oguzkocer: The existing implementation already showcases how this can be done, but I'll create a commit in another branch for you to cherry pick and comment here with the hash.

wp_api/src/api_error.rs
oguzkocer: I believe we need to keep `thiserror::Error` because that's what implements the `Error` trait. I am actually a little surprised this works without `thiserror::Error` because [I thought `uniffi::Error` required `Error` to be implemented.](https://mozilla.github.io/uniffi-rs/0.27/udl/errors.html)

> For UniFFI to expose this error, your error type (`E`) must be an `enum` and implement `std::error::Error` ([thiserror](https://crates.io/crates/thiserror) works!).

wp_api_integration_tests/src/lib.rs
```diff
@@ -148,7 +148,7 @@ pub trait AssertResponse {
     fn assert_response(self) -> Self::Item;
 }
 
-impl<T: std::fmt::Debug, E: std::error::Error> AssertResponse for Result<T, E> {
+impl<T: std::fmt::Debug, E: std::fmt::Display + std::fmt::Debug> AssertResponse for Result<T, E> {
```
oguzkocer: I was a little confused about this change when I first saw it, but didn't raise it because I didn't understand why it was required. I only saw [this comment](https://github.com/Automattic/wordpress-rs/pull/594/files#r2010983929) just now and It think this is the reason for it. Once the error types derive `thiserror::Error`, we can revert this change.

wp_api/src/parsed_url.rs
oguzkocer: Unfortunately there isn't as clear of a separation as we'd like between debug error messages and UI error messages. So, I could see this showing up in the logs for devs as well.

I see your point though, so if it's an intentional change that you want to keep, I'll leave it at that for now.

wp_localization/localization/tr-TR/main.ftl
oguzkocer: Unfortunately I don't think I'd be able to gauge their correctness, because I haven't been exposed to any technical terms that are used in Turkey. A bunch of the translations feel weird to me, but I don't know if that's because it's translated unnaturally or because I am just unfamiliar with the terms.

Having said all that, outside the unfamiliar technical term translations, they seem generally reasonable to me. 🤷‍♂️ 

I am sorry for not being able to be much help on this.

wp_api/src/login/url_discovery.rs
oguzkocer: You can cherry pick dbd841063e15372eb06475b44ca474f037e02c2c to address the issue.

crazytonyli: Hmmm, is it okay to remove `uniffi::Error` from those enums? I thought the reason they were declared as `uniffi::Error` is that some Rust functions may throw those errors.

oguzkocer: Yeap, it's OK. In fact, they should have been `uniffi::Enum`s, but the login process went through too many changes and we are left with some unintentional declarations.

These errors should never be thrown from any of the uniffi exported functions because they are only meaningful within the context of their outer errors. Using `uniffi::Enum` should help avoid accidentally throwing them. It's best to think of these as `reason`s for an error rather than being errors themselves.

crazytonyli: Their naming probably can be improved. There are `XxxError`, `XxxFailure`, and `XxxReason` types. I can see that the `XxxReason` types should not be thrown as errors, but the `XxxError` and `XxxFailure` seem pretty similar if we just look at their names.

wp_api/src/parsed_url.rs
crazytonyli: > Unfortunately there isn't as clear of a separation as we'd like between debug error messages and UI error messages.

The `Debug` derived macro is for debug messages, right? We should be able to print the detailed error information while debugging?

oguzkocer: These are currently not exported through `uniffi`, but [it looks like `Debug` is one of the traits we can export](https://mozilla.github.io/uniffi-rs/latest/types/interfaces.html). So, you're right, we can treat them as UI messages.

wp_api/src/login/url_discovery.rs
oguzkocer: Agreed. I think we can use `XxxReason` as the naming scheme for these types. Feel free to make that change in this PR or I can open a new one once this is merged.

wp_api/src/api_error.rs
crazytonyli: I'm not sure if it's a bug or by design, but it appears the `Display` trait is all uniffi needs at the moment. Both `uniffi::Enum` and `uniffi::Error` can be used as uniffi APIs's return `Result<.., E>`, as long as they implement `Display`.

oguzkocer: Regardless, I think it's a good idea for us to implement the `Error` trait which is a Rust ecosystem staple. We don't lose anything by it, so I can't see any reason not to.

oguzkocer (approved):
The CI error should be a simple fix, so once that's addressed, I think this is good to merge.

wp_api/src/login/url_discovery.rs
crazytonyli: I think this PR is large enough. If you don't mind, I'll merge this PR and leave the naming changes to you.

oguzkocer: Don't mind at all. In fact, I'd really like to merge this PR asap as what it has conflicts with what I am working on at the moment 😅

PR #598, opened 2025-03-10
Add Parsing Improvements

jkmassel:
While testing against a list of known WordPress sites, I came across several parsing failures. 

This PR contains all of the improvements I made during this process, including adding another entry for plugins that block Application Password authentication. The commit messages have a bit more detailed info on why the changes were needed.

Includes unit tests, so if CI passes this should be good to go.


wp_api/src/login.rs
```diff
@@ -123,12 +125,20 @@ pub struct KnownApplicationPasswordBlockingPlugin {
...
+                name: "Wordfence".to_string(),
+                namespace: "wordfence/v1".to_string(),
+                // TODO: Ensure this is correct with the WordFence folks
+                support_url: "https://www.wordfence.com/support/".to_string(),
+            },
+            Self {
+                name: "Hostinger Tools".to_string(),
+                namespace: "hostinger-tools-plugin/v1".to_string(),
+                // TODO: Ensure this is correct with the Hostinger folks
+                support_url: "https://wordpress.org/support/plugin/hostinger/".to_string(),
```
crazytonyli, line 153: An alternative way is to hard-code the WordPress.org plugin directory slug (which is unlikely to change) instead. We can get the support URL from the plugin directory, which should always be correct.

```
curl -s 'https://api.wordpress.org/plugins/info/1.2/?action=plugin_information&slug=hostinger' | jq -r '.support_url'
```

wp_serde_helper/src/lib.rs
```diff
@@ -129,6 +174,45 @@ where
...
+    where
+        E: de::Error,
+    {
+        if v.to_lowercase().trim() == "false" {
+            return Ok(None);
+        }
+
+        Ok(Some(v.to_string()))
+    }
+}
```
crazytonyli, line 207: I think implementing a Visitor is a bit overkill here. An alternative could be something like this:

```rust
enum Value {
  String { value: String }
  Bool { value: bool }
}

fn deserialize_false_or_string() {
  let value = Value::deserialize(deserializer)
  match value {
    ...
  }
}
```

---

BTW, I wonder if the `WpDeserialize` macro might be helpful in parsing WpApiDetails.

wp_api/src/login.rs
jkmassel: We could do that, but it's another `await` which I'd be nervous about.

For now, this URL won't change (IINM plugin slugs are immutable in the directory) but I'm pretty sure we'll end up sending them to something like `https://support.hostinger.com/$SOME_PATH`

wp_serde_helper/src/lib.rs
oguzkocer: `WpDeserialize` would only work if we mark all fields as `Option`. Otherwise, we have to manually implement `WpDeserialize` does.

`WpDeserialize` uses `DeserializeEmptyVecOrT` which doesn't handle `false` or `String`, so it wouldn't be helpful even if we could use `WpDeserialize` for `WpApiDetails`.

oguzkocer: > I think implementing a Visitor is a bit overkill here. An alternative could be something like this

I think it's fine to implement a Visitor for this. It's more performant and the implementation is not particularly complicated.

wp_api/src/login.rs
```diff
@@ -332,6 +363,39 @@ mod tests {
...
+
+        if result.is_err() {
+            // Logging the error should give us a line number or missing field name to figure out the problem
+            println!(
+                "Failed to parse json as `WpApiDetails`: {}",
+                result.as_ref().err().unwrap()
+            );
+        }
+
+        assert!(result.is_ok());
```
oguzkocer:
```suggestion
        assert!(
            result.is_ok(),
            "Failed to parse json as `WpApiDetails`: {:#?}",
            result
        );
```

wp_api/src/login.rs
```diff
@@ -332,6 +363,39 @@ mod tests {
             .is_empty());
     }
 
+    #[rstest]
+    #[case("api-details/invalid-json-01.json")]
+    #[case("api-details/invalid-json-02.json")]
+    fn test_valid_json_documents(#[case] input: &str) {
```
oguzkocer: The test files are called `invalid-json-..` whereas the test is called `test_valid_json_documents`. The assertion seems to expect the files to be valid, so could we rename the files to match the expectation?

wp_api/src/login.rs
```diff
@@ -332,6 +363,39 @@ mod tests {
...
+                result.as_ref().err().unwrap()
+            );
+        }
+
+        assert!(result.is_ok());
+    }
+
+    fn test_json(input: &str) -> Result<Vec<u8>, std::io::Error> {
+        let mut file_path = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        file_path.push("tests");
```
oguzkocer: The path `./tests` have a special meaning for cargo. Could we rename the folder to `test-data`?

I am a little reluctant to do IO as part of unit tests, but I think it's fine to leave it as is for now.

wp_api/src/login/login_client.rs
```diff
@@ -110,14 +110,15 @@ impl WpLoginClient {
             }
         };
         let api_details: WpApiDetails =
-            match serde_json::from_slice::<WpApiDetails>(&fetch_api_details_response.body) {
+            match WpApiDetails::try_from(fetch_api_details_response.body) {
                 Ok(api_details) => api_details,
                 Err(error) => {
+                    // println!("Error parsing API details: {}", String::from_utf8(fetch_api_details_response.body).unwrap());
```
```suggestion
```

wp_api/src/login.rs
```diff
@@ -51,14 +52,29 @@ pub struct WpApiDetails {
...
     pub site_icon_url: Option<String>,
 }
 
+impl TryFrom<Vec<u8>> for WpApiDetails {
+    type Error = serde_json::Error;
+
+    fn try_from(mut value: Vec<u8>) -> Result<Self, Self::Error> {
+        // If the body starts with the UTF-8 BOM, remove it
+        if value.starts_with(&[0xEF, 0xBB, 0xBF]) {
+            value = value[3..].to_vec();
```
oguzkocer:
```suggestion
            value.drain(0..3);
```

oguzkocer (changes requested):

wp_api/src/login/login_client.rs
jkmassel: It'd be super handy for debugging if there were a way to fetch the body from the error type

wp_api/src/login.rs
jkmassel: Addressed in https://github.com/Automattic/wordpress-rs/pull/598/commits/a3e84220c137e41bb5ae6d38c18ad0f5c1fda868

wp_api/src/login.rs
jkmassel: Addressed in https://github.com/Automattic/wordpress-rs/pull/598/commits/a3e84220c137e41bb5ae6d38c18ad0f5c1fda868

I think disk-based IO is generally ok – I'd rather keep the tests readable for parsing larger JSON docs

oguzkocer (approved):

PR #607, opened 2025-03-12
Add endpoint guessing

jkmassel:
This is a proof-of-concept PR demonstrating endpoint guessing. The vast majority of sites don't use a custom REST API root, so we should just be able to tack `wp-json` onto the end of the user-provided input and have it "just work".

This increases our success rate in a small sample set by 1%, and increases our useful error message rate by at least 6%.

### DIFF
--- a/wp_api/src/login/login_client.rs
+++ b/wp_api/src/login/login_client.rs
@@ -188,21 +188,29 @@ impl WpLoginClient {
                 parsed_site_url,
                 api_root_url,
             })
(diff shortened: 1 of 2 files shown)


crazytonyli:
This constant probably should be removed too.

https://github.com/Automattic/wordpress-rs/blob/6b13f6635f1e8ea00f382c8492bffdd2baa6f014/wp_api/src/request/endpoint.rs#L19

jkmassel:
Closing in favour of #611

PR #611, opened 2025-03-13
[Api Discovery] Try root `/wp-json` if we can't find link header

oguzkocer:
I started this PR as suggestions for https://github.com/Automattic/wordpress-rs/pull/607, but since I pretty much re-wrote the entire thing - and implemented a few other improvements, I think it's easier to review the PR by targeting `trunk` directly. I still kept the original commit from @jkmassel just as a reference.

In some cases we can't find the api root url from the link header or by parsing the home page, but we can still access it by trying `{site_url}/wp-json`. This PR adds this as a fallback mechanism to `find_api_root_url` step.

Note that we were already making a request to `{site_url}/wp-json` as part of `is_wordpress_site` discovery. So, I moved this to an earlier step and re-used its result instead of making a separate request. Having said that, in some cases we are still making 2 requests to `/wp-json`, one before we find the api root and one afterwards. It should be possible to re-use the result of the initial request when the api root is the same as `{site_url}/wp-json`, so we'll want to fix that in a separate PR.

I've also included a general improvement to the api discovery by parsing the attempt url at the beginning and then short cutting the entire api discovery attempt if parsing fails. This shouldn't result any practical changes, but it saves us from parsing the same url multiple times and it also simplifies the error handling in api discovery helpers.

### DIFF
--- a/wp_api/src/login/login_client.rs
+++ b/wp_api/src/login/login_client.rs
@@ -181,22 +181,34 @@ impl WpLoginClient {
         fetch_wp_json_result: Result<FetchWpJsonSuccess, FetchWpJsonFailure>,
     ) -> Result<FindApiRootLinkHeaderSuccess, FindApiRootLinkHeaderFailure> {
         if let Some(api_root_url) = api_root_url_from_link_tag {
-            Ok(FindApiRootLinkHeaderSuccess {
+            return Ok(FindApiRootLinkHeaderSuccess {
                 parsed_site_url,
                 api_root_url,
-            })
-        } else if let Ok(success) = self
+            });
+        }
+        match self
             .fetch_and_parse_api_root_url_header(parsed_site_url.clone())
             .await
         {
-            Ok(success)
-        } else if let Ok(fetch_wp_json_success) = fetch_wp_json_result {
-            Ok(FindApiRootLinkHeaderSuccess {
-                parsed_site_url,
-                api_root_url: fetch_wp_json_success.wp_json_url,
-            })
-        } else {
-            Err(FindApiRootLinkHeaderFailure::ApiRootNotFound { parsed_site_url })
+            Ok(s) => Ok(s),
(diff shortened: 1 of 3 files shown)


wp_api/src/login/login_client.rs
```diff
@@ -162,35 +174,41 @@ impl WpLoginClient {
...
+            .await
+        {
+            Ok(s) => Ok(s),
+            Err(fetch_and_parse_api_root_url_header_err) => {
+                // If we can't find the link header, but we were able to fetch `/wp-json` from the
+                // attempt url, we assume that it's the correct api root url
+                //
+                // We don't immediately rely on this because there might be cases where there are
+                // multiple `/wp-json` paths in which case we want to prioritize the one returned
+                // from the link header
```
jkmassel: More specifically, there might be a `/wp-json` path _and_ the actual REST API root (which could be something else).

Not sure if we _need_ to update this comment – I'll leave that to you

wp_api/src/login/login_client.rs
```diff
@@ -239,8 +258,7 @@ impl WpLoginClient {
         }
     }
 
-    // Fetches the site's homepage with a HEAD request, then extracts the Link header pointing
-    // to the WP.org API root
+    // Fetches the site's homepage with a HEAD request
```
jkmassel:
```suggestion
    // Fetches the site's homepage headers with a HEAD request
```

jkmassel (approved):
Approved, but I disabled auto merge for one very small suggestion

wp_api/src/login/login_client.rs
oguzkocer: Addressed in 30cfa0427745e558e20a4d9cca693aa45e468e6c.

wp_api/src/login/login_client.rs
oguzkocer: Addressed in 30cfa0427745e558e20a4d9cca693aa45e468e6c.

PR #613, opened 2025-03-14
Add `ReqwestRequestExecutor` behind `reqwest-request-executor` Cargo feature

oguzkocer:
This PR moves the `wp_api_integration_tests::AsyncWpNetworking` to `wp_api` crate by renaming it to `reqwest_request_executor::ReqwestRequestExecutor`. It also removes the `ReqwestRequestExecutor::Default` implementation in favor `ReqwestRequestExecutor::new(danger_accept_invalid_certs: bool)`, since in production accepting invalid certificates is probably not the client would want. Note that, both `reqwest_request_executor` module and the `reqwest` dependency are behind `reqwest-request-executor` Cargo feature.

`ReqwestRequestExecutor` in its current state is probably not ready to be used in production. We have a couple unwraps we should remove and turn into appropriate errors. We are also lacking some of the extra error handling Swift client has. Having said that, I think it's best to make this move without any functional changes and open subsequent PRs for improvements. @jkmassel is already working on such improvements at the moment, so I don't want to create any conflicts with that.


jkmassel (approved):
This looks like a pretty simple move of the request executor in to `wp-api`, so I think we should marge it then I'll rebase #610 on top of it and try out the test suite

Issue #605, opened 2025-03-12
Handle wp_error for /wp-json


jkmassel:
There are plugins that emit a `WP_Error ` when the REST API is disabled. This is pretty standard behaviour (return a valid response shape, or else a `WP_Error`) for the rest of the API, so we should handle it as an application-level error in the library.

Here are two example responses:

```json
{
  "code": "rest_forbidden",
  "message": "REST API access is restricted.",
  "data": {
    "status": 403
  }
}
```

```json
{
  "code": "rest_not_logged_in",
  "message": "You are not currently logged in.",
  "data": {
    "status": 401
  }
}
```

Example Plugins:
- https://wordpress.org/plugins/disable-json-api/
- https://wordpress.org/plugins/disable-wp-rest-api/

For login specifically, we should have the ability to scan the `code` field and have a list of signatures (similar to what we use for the list of plugins that block Application Passwords) and tie that back to a specific plugin so we can tell the user "You have plugin xxxx installed, and the JSON API is disabled because of it. You'll need to disable this plugin to use the app"

oguzkocer:
@jkmassel I've addressed this issue in #625. However, I am not really sure what actions I need to take about this following section:

> For login specifically, we should have the ability to scan the `code` field and have a list of signatures (similar to what we use for the list of plugins that block Application Passwords) and tie that back to a specific plugin so we can tell the user "You have plugin xxxx installed, and the JSON API is disabled because of it. You'll need to disable this plugin to use the app"

Do you want to add a `maybe_caused_by_plugins: Vec<KnownRestApiBlockingPlugin>` field to `AutoDiscoveryAttemptFailure::WpError`?

PR #625, opened 2025-03-18
Implement `AutoDiscoveryAttemptFailure::WpError`

oguzkocer:
Addresses #605. During api discovery, if `/wp-json` returns a `WpError`, we now parse that into `AutoDiscoveryAttemptFailure::WpError` error variant. Since this case is difficult to cover with an integration test, the PR adds a unit test to verify that we are able to parse this JSON correctly.

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,1 +1,1 @@
-large-error-threshold = 192
+large-error-threshold = 256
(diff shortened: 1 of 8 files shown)


jkmassel (approved):
I tested this by cherry-picking it into https://github.com/Automattic/wordpress-rs/tree/add/cli-tool and testing a production site with the REST API disabled, and a test site that returns a `403` with no other details.

Both worked

PR #626, opened 2025-03-18
[Api Discovery] Remove `AutoRemoveWpAdminSuffix` & `AutoRemoveWpLoginSuffix` attempts 

oguzkocer:
As requested by @jkmassel on Slack; this PR removes `AutoRemoveWpAdminSuffix` & `AutoRemoveWpLoginSuffix` individual attempt types to combine them into `AutoStrippedHttps` singular attempt type.

This means, api discovery for urls such as `http://localhost/wp-admin` will no longer succeed, because `http://localhost` will not be attempted and `https://localhost` doesn't work. The argument is that if the site is not using `https`, it'll likely result in `AutoDiscoveryAttemptFailure::ApplicationPasswordsNotSupported` error with `ApplicationPasswordsNotSupportedReason::ApplicationPasswordsDisabledForHttpSite` reason.

### DIFF
--- a/wp_api/src/login/url_discovery.rs
+++ b/wp_api/src/login/url_discovery.rs
@@ -30,11 +30,16 @@ impl AutoDiscoveryAttempt {
         Self::new(attempt_site_url, AutoDiscoveryAttemptType::UserInput)
     }
 
-    fn with_auto_https_attempt_type(attempt_site_url: impl Into<String>) -> Self {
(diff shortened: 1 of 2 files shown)


jkmassel (approved):
> api discovery for urls such as http://localhost/wp-admin will no longer succeed

I think this makes a lot of sense – if you want to use an `http` URL, it must be the exact correct one. 

Given the fact that (almost) nobody will be using `localhost` (because who runs a server on their device), I think we can certainly live with the loss of this case

PR #627, opened 2025-03-20
Remove `is_wordpress_site` discovery in favor of `ProbablyNotAWordPressSite` error variant

oguzkocer:
As requested by @jkmassel, this PR removes the separate `is_wordpress_site` discovery in favor of introducing a new error variant `ProbablyNotAWordPressSite`. The goal for this change is to simplify the error handling for clients where they previously had to combine the result for `is_wordpress_site` and `api_discovery`.

Since we don't have to do multiple discoveries, this results in a significant simplification of the process. The main reason for the previous complexity was due to some requests having to be made in early on just so its results can be used by both discoveries. Due to this, the diff may be a little harder to understand, so I suggest taking a look at the final implementation instead.

The PR also introduces changes to the `AutoDiscoveryAttemptFailure` error to present a more structured error type.

I ended up dropping some of the error helpers from `AutoDiscoveryAttemptFailure` as I thought they either didn't make much sense anymore or they weren't necessary. We can re-visit that decision according to our needs. I also temporarily removed the Kotlin localization tests because `ParseApiRootUrlError::ApiRootLinkHeaderNotFound` has been removed by this PR. I intend to re-introduce these tests once we update all the errors to use localizations.


jkmassel (approved):
I tested this atop `add/cli-tool` and got zero instances of `Unhelpful Error` 🥳

PR #637, opened 2025-03-25
Remove ApiBaseUrl

crazytonyli:
Addresses https://github.com/Automattic/wordpress-rs/pull/586#discussion_r2010437994.

Considering we have found [the real wp-json API](https://github.com/Automattic/wordpress-rs/blob/633141be69129c7c1adbd0f670bc7946bdc60b70/wp_api/src/login/url_discovery.rs#L362) in the url discovery process, I think we should use it to create `WpApiClient`, instead of using a "site URL" which will be used to construct a hardcoded `<site-url>/wp-json` as the API root URL which may be incorrect. That's basically the main change in this PR (see the changes in the `WpApiRequestBuilder::new` function), and the rest of the PR is just adopting that API change.

This PR introduces a breaking change. There are a couple of things the client apps need to do:
1. They should store the API root URL in the URL discovery result during logging into the site, so that the API root URL can reused later on.
2. If the app is already logged in, but no the API root URL is not stored (which is the production app environment), they should just use hardcoded `<site-url>/wp-json` as the API root URL.


oguzkocer (commented):
I'd be fine with dropping the `ApiBaseUrl` type. However, I wonder if it's worth using a wrapper type for `ApiRootUrl` instead of using `ParsedUrl` directly. We already do this internally in the login flow, so maybe we should consider exposing that and re-using it everywhere. In any case, that change doesn't need to happen in this PR and unless we come across issues while using `ParsedUrl` directly, I don't feel strongly about introducing a new type for this.

I didn't approve since it can't be merged yet. Let me know if you need help with the Kotlin integration changes. Happy to do that directly in this PR or open a new PR against this one. Also happy to just review your changes if you want to update it yourself.

wp_derive_request_builder/src/generate.rs
```diff
@@ -444,7 +444,7 @@ pub struct ConfigStaticTypes {
 impl ConfigStaticTypes {
     fn new(crate_ident: &Ident) -> Self {
         Self {
-            api_base_url: quote! { std::sync::Arc<#crate_ident::request::endpoint::ApiBaseUrl> },
+            api_base_url: quote! { std::sync::Arc<#crate_ident::ParsedUrl> },
```
crazytonyli: The generated code still uses `api_base_url` as argument names, which can be fixed by a search-and-replace: `api_base_url` -> `api_root_url`. If the approach is acceptable, I can create a separate PR for that or push another commit.

oguzkocer: You can push it directly to this PR. Since I already reviewed the PR, I am now catching up on the changes commit by commit. Thanks for checking 👍

crazytonyli:
@oguzkocer I couldn't get my Android Studio working. Can you please help to address the issues in the kotlin project? _I think_ adding `wp-json` to the site URL in the integration tests is all it needs.

oguzkocer:
@crazytonyli I've addressed the Kotlin changes in 93ca3c4c08cd7cf5436918dddee211b66f9e5d32.

When I search for `api_base_url` in the project, I still see quite a few usages mostly as part of `fixture_api_base_url`, but also a few in `api_client.rs`. Would you like to address them in this PR?

oguzkocer (approved):

PR #639, opened 2025-03-26
Use `FluentValue` instead of `String` for localization arguments

oguzkocer:


wp_localization_macro/src/wp_messages.rs
```diff
@@ -151,21 +151,21 @@ mod bindings {
         let args = entry.placeables.iter().map(|placeable| {
             let placeable_ident = format_ident!("{}", placeable.0);
             quote! {
-                #placeable_ident: impl Into<String>,
+                #placeable_ident: impl Into<fluent_bundle::FluentValue<'a>>,
```
crazytonyli, line 154: Nitpick: use `fluent::FluentValue`? `fluent_bundle` probably exports `fluent` APIs.

crazytonyli (approved):

wp_localization_macro/src/wp_messages.rs
oguzkocer: At the moment, [we are using `fluent_bundle` directly](https://github.com/Automattic/wordpress-rs/blob/5eeee1a758fd6b03bba9ca60c7e7835cc1671653/wp_localization/Cargo.toml#L10) and [don't depend on `fluent`](https://github.com/Automattic/wordpress-rs/blob/5eeee1a758fd6b03bba9ca60c7e7835cc1671653/Cargo.toml#L28-L31).

Furthermore, this value is used in [`fluent-templates::Loader::lookup_complete`](https://docs.rs/fluent-templates/latest/fluent_templates/trait.Loader.html#tymethod.lookup_complete) which expects [`fluent_bundle::FluentValue`](https://docs.rs/fluent-bundle/0.15.3/fluent_bundle/types/enum.FluentValue.html).

There might be a `From` implementation to convert `fluent::FluentValue` values somewhere, so using that might also work, but at the moment our setup seems correct to me.

PR #641, opened 2025-03-29
Introduce `wp_localization_validation` crate to validate localization files

### LINKED ISSUE #575: Validate translations
We'll extract the parsing code from `wp_localization_macro` to a new crate and then build a tool that compares the default localization file with translations.

* There should be no keys in translation files that are not also present in the default localization file.
* For any given key, the placeables (`$value`) in the translation file should match the default localization file on an unordered basis.
* The report should include the translation percentage for all languages.

oguzkocer:
Closes #575. Adds a new binary tool to validate the localization files. It handles the following issues:

* There can only be one `ftl` file per language - Reported as `FtlSetupError::MultipleFilesForLanguage`. If in the future we need multiple files, we'll need to update our tooling to allow that and then we can remove this validation.
* There can't be any keys in translation files that are not also present in the default localization file - Reported as `ValidationError::KeyNotInDefault`
* There can't be any duplicate keys for any language - Reported as `ValidationError::DuplicateKeys`
* For any given key, the `placeables` ($value) in the translation file must match the default localization file on an unordered basis - Reported as `ValidationError::MismatchedPlaceables`
* If in `strict` mode, all keys must be translated - Reported as `ValidationError::MissingTranslations`.

If there are no critical issues found, it'll print the translation percentage for each language.

---

I have some minor grievances about the implementation, especially how finding duplicate keys and converting to `HashMap` happens together and outside the other validation logic. However, it'd be inefficient to split them up, so I left it alone for now. On the other hand, I initially implemented the validations in a more efficient way, but found the code annoying to follow, so I later split them up. The tool ended up being so fast that I really shouldn't have thought about performance at all, but unless there is a request during review, I think I am just going to leave it be.

_P.S:_ Since this is an isolated implementation, I used it to evaluate Cursor AI and I actually kept some of the code written by it. However, since I had to fight it quite a bit, the whole PR ended up being a mix of how I usually implement things and how Cursor AI did. That is probably part of the reason why I have grievances about my own PR 🙈 Fwiw, I am still in the process of integrating AI into my workflow, but I gave up on Cursor AI at least for now.


wp_localization_validation/src/ftl_file.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    #[error("Default language '{0}' not found")]
+    DefaultLanguageNotFound(TranslationLanguage),
+
+    #[error("No FTL files found in language directory '{lang_id}'")]
+    NoFtlFiles { lang_id: TranslationLanguage },
+}
+
+// Forward the `Display` trait implementation
+impl Debug for FtlSetupError {
```
crazytonyli, line 82: We can probably just use the derive macro instead, like almost all other types?

wp_localization_validation/src/ftl_file.rs
```diff
@@ -0,0 +1,86 @@
...
+///   tr-TR/
+///     main.ftl
+/// ```
+///
+/// # Returns
+///
+/// A map of language codes to their translation entries.
+pub fn parse_localization_files(
+    localization_dir: &str,
+) -> Result<HashMap<TranslationLanguage, Vec<TranslationEntry>>, FtlSetupError> {
```
crazytonyli, line 25: Can we move this function to the parser crate? We could potentially reuse it in the wp_localization crate's build script.

crazytonyli (approved):

wp_localization_validation/src/ftl_file.rs
oguzkocer: That wouldn't work in this case, otherwise I wouldn't have manually implemented it 😅

As the line comment states, we are forwarding the implementation for the `Display` trait for this `Debug` trait implementation. If we derived the `Debug` trait, then it'd generate a `String` that uses the `Debug` formatting for each of its fields whereas when we forward the `Display` trait implementation we get the error message from the `#[error("")]` implementation.

In a library crate, we probably wouldn't want to do this unless it is an edge case. However, there is a good reason to do this in a binary crate and that's to be able to use the `String` from `Display` trait from `fn main() -> Result<(), FtlSetupError>`. When we return a `Result` from `fn main()`, if there is an error during program execution, it'll use the `Debug` formatting to display the error message. In this case, using the `Display` formatting is more accurate, so we forward it to `Debug` implementation.

Here is how you can see the difference:
1. Create a second ftl file for `en-US` so the program execution fails during setup: `cp wp_localization/localization/en-US/main.ftl wp_localization/localization/en-US/aa.ftl`
2. Run the validation tool and take a look at the error message: `cargo run --bin wp_localization_validation -- --localization-folder ./wp_localization/localization/`
3. Remove this manual `Debug` implementation and derive it instead
4. Run the validation tool again and notice how it's now showing just the error variant and its fields

```
// Error message from Step 2
Error: Multiple files found for language 'en-US'
// Error message from Step 4
Error: MultipleFilesForLanguage { lang_id: TranslationLanguage("en-US") }
```

---

Just to re-iterate, if this error type was in a library crate I think it'd be better to have separate `Debug` and `Display` implementations. In that case, we could have manually changed which formatting is used in `fn main()`. However, since this is a binary crate, this approach makes things more convenient as we can now return the error with `?` just like we do for any other function.

Does that make sense?

wp_localization_validation/src/ftl_file.rs
oguzkocer: I think that'd be fine, but we can leave that to when we need it or when we are updating the build script especially since the PR is already merged.

PR #652, opened 2025-04-07
Simplify uniffi api discovery result

oguzkocer:
Since api discovery is done through multiple attempts, it can be a little difficult to work with its result especially when it results in an error. To help combat this, this PR introduces the concept of "error importance" for `AutoDiscoveryAttemptFailure` so that we can identify which error would be the most useful to share with the user. This is currently closely related to when the error occurs in the api discovery process where the errors from later steps are treated as more important.

This concept is applied as a helper in the Rust API using the new `combined_result` function. However, for native clients, it removes the previous `AutoDiscoveryUniffiResult` in favor of directly returning this combined result. This means the clients can no longer show the result of each attempt, but that wasn't being useful at the moment. If we ever want to make the result of all attempts available, it'll be a simple change.

### DIFF
--- a/wp_api/src/login/url_discovery.rs
+++ b/wp_api/src/login/url_discovery.rs
@@ -4,9 +4,15 @@ use crate::{
     login::KnownApplicationPasswordBlockingPlugin,
     request::{ResponseBodyType, WpRedirect},
 };
+use itertools::Itertools;
 use scraper::{Html, Selector};
 use serde::Deserialize;
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    cmp::Ordering,
+    collections::HashMap,
+    num::{NonZero, NonZeroUsize},
+    sync::Arc,
+};
 use wp_localization::{MessageBundle, WpMessages, WpSupportsLocalization};
 use wp_localization_macro::WpDeriveLocalizable;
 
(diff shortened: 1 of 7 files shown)


wp_api/src/login/url_discovery.rs
```diff
@@ -129,6 +109,24 @@ impl AutoDiscoveryResult {
...
+            success.api_discovery_result.as_ref()
+        } else {
+            Err(self
+                .attempts
+                .iter()
+                .flat_map(|(_, a)| a.api_discovery_result.as_ref().err())
+                // Sort in descending order so the most important error is returned
+                .sorted_by(|a, b| b.compare_importance(a))
+                .next()
+                .expect("If the discovery was unsuccessful, there is at least one error"))
```
crazytonyli, line 127: Nitpick: I think you can get rid of this unnecessary `expect` if deleting `find_successful` and re-implementing it within this function.

crazytonyli (approved):

wp_api/src/login/url_discovery.rs
oguzkocer: I considered that, but that function should still be useful for Rust clients. I also considered duplicating the implementation, but the expect didn't bother me enough to do that 🤷

PR #658, opened 2025-04-09
Export WpLoginClient to replace WordPressLoginClient

crazytonyli:
We need to return the `AutoDiscoveryAttemptSuccess.apiRootUrl` to the clients, so that the API root URL can be used (and stored) to create `WordPressAPI` instances.

Exposing the discovery result from the `WordPressLoginClient` wrapper makes the wrapper itself a very thin shell for the Rust type `WpLoginClient`, so I decided to remove the wrapper and `UniffiWpLoginClient` and export `WpLoginClient` directly as `uniffi::Object`. I can get rid of this change if it's inappropriate, which means we'll need to return the `AutoDiscoveryAttemptSuccess.apiRootUrl` in the wrapper in another way.

### DIFF
--- a/native/swift/Example/Example/UI/LoginView.swift
+++ b/native/swift/Example/Example/UI/LoginView.swift
@@ -69,8 +69,8 @@ struct LoginView: View {
                     callbackUrl: "x-wordpress-app://login-callback"
                 )
 
-                let loginClient = WpLoginClient(urlSession: .shared)
-                let details = try await loginClient.apiDiscovery(siteUrl: url)
+                let loginClient = WordPressLoginClient(urlSession: .shared)
+                let details = try await loginClient.details(ofSite: url)
                 let loginUrl = try await details.loginURL(for: application)
 
                 let callbackUrl = try await self.webAuthenticationSession.authenticate(
--- a/native/swift/Example/Example/UI/LoginView.swift
+++ b/native/swift/Example/Example/UI/LoginView.swift
@@ -69,8 +69,9 @@ struct LoginView: View {
                     callbackUrl: "x-wordpress-app://login-callback"
                 )
 
-                let loginClient = WordPressLoginClient(urlSession: .shared)
-                let loginUrl = try await loginClient.loginURL(forSite: url, application: application)
+                let loginClient = WpLoginClient(urlSession: .shared)
+                let details = try await loginClient.apiDiscovery(siteUrl: url)
+                let loginUrl = try await details.loginURL(for: application)
 
                 let callbackUrl = try await self.webAuthenticationSession.authenticate(
                     using: loginUrl,
@@ -78,7 +79,7 @@ struct LoginView: View {
                 )
 
                 let loginDetails = try loginClient.credentials(from: callbackUrl)
-                try await loginManager.setLoginCredentials(to: loginDetails)
+                try await loginManager.setLoginCredentials(to: loginDetails, apiRootURL: details.apiRootUrl.asURL())
             } catch let err {
                 handleLoginError(err)
             }
--- a/native/swift/Example/Example/WordPressAPI+Extensions.swift
+++ b/native/swift/Example/Example/WordPressAPI+Extensions.swift
@@ -11,13 +11,7 @@ extension WordPressAPI {
                 throw CocoaError(.validationMissingMandatoryProperty)
             }
 
-            let parsedUrl = try ParsedUrl.parse(input: defaultSiteUrl)
-            let apiRootUrl = try ParsedUrl.parse(
-                input: parsedUrl
-                    .asURL()
-                    .appendingPathComponent("wp-json")
-                    .absoluteString
-            )
+            let apiRootUrl = try ParsedUrl.parse(input: defaultSiteUrl)
 
             guard let loginCredentials = try await loginManager.getLoginCredentials() else {
                 throw CocoaError(.xpcConnectionInvalid)
--- a/native/swift/Example/Example/WordPressAPI+Extensions.swift
+++ b/native/swift/Example/Example/WordPressAPI+Extensions.swift
@@ -7,7 +7,7 @@ extension WordPressAPI {
         get async throws {
             let loginManager = await LoginManager()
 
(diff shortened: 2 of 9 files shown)


native/swift/Example/Example/UI/LoginView.swift
```diff
@@ -69,16 +69,17 @@ struct LoginView: View {
...
+                let loginUrl = try await details.loginURL(for: application)
 
                 let callbackUrl = try await self.webAuthenticationSession.authenticate(
                     using: loginUrl,
                     callbackURLScheme: "x-wordpress-app"
                 )
 
                 let loginDetails = try loginClient.credentials(from: callbackUrl)
-                try await loginManager.setLoginCredentials(to: loginDetails)
+                try await loginManager.setLoginCredentials(to: loginDetails, apiRootURL: details.apiRootUrl.asURL())
```
crazytonyli, line 82: The changes in this file are how login should work: the client needs both the login URL and the API root URL.

oguzkocer (dismissed):
There shouldn't be any need to make changes to the Rust code because `apiRootUrl` is available in `AutoDiscoveryAttemptSuccess`.

We have intentionally separated the login clients for Rust and uniffi and this PR is removing that without any clear explanation to its reasons. We have a Kotlin login client that re-uses the `WpLoginClient` name which was one reason to keep them separate. Another one is that it's consistent with the `UniffiWpApiClient` and `WpApiClient` separation. Both of these separations provide a clear boundary between Rust and uniffi clients which is lacking in `uniffi::Object` types. It also makes it easier to keep the public API the same while changing the internals without worry.

native/swift/Example/Example/WordPressAPI+Extensions.swift
```diff
@@ -11,13 +11,7 @@ extension WordPressAPI {
...
             }
 
-            let parsedUrl = try ParsedUrl.parse(input: defaultSiteUrl)
-            let apiRootUrl = try ParsedUrl.parse(
-                input: parsedUrl
-                    .asURL()
-                    .appendingPathComponent("wp-json")
-                    .absoluteString
-            )
+            let apiRootUrl = try ParsedUrl.parse(input: defaultSiteUrl)
```
oguzkocer, line 14: Maybe I am misunderstanding what is happening here, but this doesn't look correct to me at a first glance. Maybe the name `defaultSiteUrl` is what's confusing me since it's being cast to `apiRootUrl`?

crazytonyli: Addressed in https://github.com/Automattic/wordpress-rs/pull/658/commits/7aa025b413456829119d0a495b027605f8ae8c7d

oguzkocer: Thank you! 🙇‍♂️ 

I'll leave the actual review to @jkmassel since this PR is only related to Swift.

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
```diff
@@ -25,21 +25,22 @@ public final class WordPressLoginClient {
         self.client = UniffiWpLoginClient(requestExecutor: requestExecutor, middlewarePipeline: middleware)
     }
 
+    /// Uses the proposed site URL to scan the website and return login related information about the site.
+    ///
+    public func details(
+        ofSite proposedSiteUrl: String
+    ) async throws -> AutoDiscoveryAttemptSuccess {
```
jkmassel, line 32: This should probably just return `client.apiDiscovery(siteUrl: proposedSiteUrl).apiDetails`

IINM there's no other useful information in `AutoDiscoveryAttemptSuccess` anyway

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
```diff
@@ -52,20 +53,42 @@ public final class WordPressLoginClient {
...
 
-    /// Convert the callback URL into a set of authentication credentials
-    ///
-    public func credentials(from callbackUrl: URL) throws -> WpApiApplicationPasswordDetails {
-        try extractLoginDetailsFromUrl(url: callbackUrl.absoluteString)
+}
+
+public extension WpApiDetails {
+
+    var applicationPasswordAuthenticationUrl: ParsedUrl {
```
jkmassel, line 79: We definitely don't want to crash for this case – there are many sites that don't support application passwords.

IINM though, Rust should emit an error for this case, so we wouldn't hit this anyway (AFAIK)

jkmassel (changes requested):

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
crazytonyli: `api_root_url` is in `AutoDiscoveryAttemptSuccess` though: https://github.com/Automattic/wordpress-rs/blob/cb6c2c17a973ba9cb121937e0e7b035af6f704a9/wp_api/src/login/url_discovery.rs#L359-L363

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
oguzkocer: That is right. If I am not mistaken, we return an error if authentication url can't be found. However, the API still returns an `Option` for this, so it's sort of annoying.

I'll see if I can somehow expose this without it being `Option` when we get a successful discovery result.

crazytonyli: @jkmassel Good point. I have moved the function to `AutoDiscoveryAttemptSuccess` in https://github.com/Automattic/wordpress-rs/pull/658/commits/537c490fe364e772093359c0ae9e7ce3381c0e9a, which only exists if application password authentication is available.

native/swift/Sources/wordpress-api/WordPressLoginClient.swift
jkmassel: Ah you're right. Never mind then :)

jkmassel (approved):

PR #660, opened 2025-04-09
Extract `ReqwestRequestExecutor` changes from `Add Web Tool` PR#653

oguzkocer:
Extracts `ReqwestRequestExecutor` changes from #653 with only very minor changes/improvements. I'll leave some line comments for where there are changes from the original PR.

Note that, all credit for this implementation belongs to @jkmassel.


Cargo.toml
```diff
@@ -31,7 +31,11 @@ fluent-bundle = "0.15"
 fluent-langneg = "0.13"
 fluent-syntax = "0.11"
 fluent-templates = "0.13"
+h2 = "0.4"
+hickory-resolver = "0.25"
 http = "1.3"
+hyper = "0.14.32" # Use the same version as `reqwest`
```
oguzkocer, line 37: As discussed on Slack, we are pinning this to the version used by the current `reqwest` version rather than using `*`.

wp_api/src/ssl.rs
```diff
@@ -81,3 +89,9 @@ pub struct SSLCertificateIssuer {
     pub organization: Option<String>,
     pub country: Option<String>,
 }
+
+impl From<x509_cert::time::Time> for WpGmtDateTime {
+    fn from(date_time: x509_cert::time::Time) -> Self {
+        WpGmtDateTime::from_timestamp(date_time.to_unix_duration().as_secs() as i64)
```
oguzkocer, line 95: The original PR had the following implementation:
```
impl From<u64> for WpGmtDateTime {
    fn from(seconds: u64) -> Self {
        Self::from_timestamp(seconds.try_into().unwrap())
    }
}
```

I like simply casting it to `i64` here because it's a clean cast. The alternative is to add a `From` implementation, but I don't think this ^ `From` implementation is the best way to do it, we should instead cast to `i64` and forward to the existing implementation.

wp_api/src/reqwest_request_executor.rs
```diff
@@ -1,29 +1,50 @@
...
 #[derive(Debug)]
 pub struct ReqwestRequestExecutor {
     client: reqwest::Client,
 }
 
+impl Default for ReqwestRequestExecutor {
+    fn default() -> Self {
+        Self::new_with_default_timeout(false)
+    }
+}
```
oguzkocer, line 27: I decided to add a `Default` implementation and use that in the test suite to make it clear what our expected default behavior is.

There is one test in `test_login_immut` that tests `http://wordpress-1315525-4803651.cloudwaysapps.com` which requires accepting dangerous certificates. I've separated this test to make it clear it's outside the default behavior.

wp_api/src/reqwest_request_executor.rs
```diff
@@ -143,3 +159,227 @@ impl RequestExecutor for ReqwestRequestExecutor {
...
+    }
+
+    fn as_hyper_error(&self) -> Option<&HyperError> {
+        self.find::<HyperError>()
+    }
+}
+
+// It's probably possible to have a single implementation for all of these, but we can do that later
+trait FindsError {
+    fn find<E: Error + 'static>(&self) -> Option<&E>;
```
oguzkocer, line 356: Removed the `'a` lifetime from the original implementation because `clippy` was complaining about it.

wp_api/src/reqwest_request_executor.rs
```diff
@@ -143,3 +159,227 @@ impl RequestExecutor for ReqwestRequestExecutor {
...
+        if error.is_incomplete_message() {
+            // The server terminated the connection unexpectedly
+            return RequestExecutionErrorReason::GenericError {
+                error_message: error.to_string(),
+            };
+        }
+
+        RequestExecutionErrorReason::GenericError {
+            error_message: error.to_string(),
+        }
```
oguzkocer, line 306: Replaced the `todo!("Unhandled hyper error: {:?}", error);` from the original implementation since we can't have `todo!()`s in production code.

wp_api/src/reqwest_request_executor.rs
```diff
@@ -143,3 +159,227 @@ impl RequestExecutor for ReqwestRequestExecutor {
...
+            };
+        }
+
+        RequestExecutionError::RequestExecutionFailed {
+            status_code: None,
+            redirects: None,
+            reason: RequestExecutionErrorReason::GenericError {
+                error_message: error.to_string(),
+            },
+        }
```
oguzkocer, line 242: Replaced the following from original implementation since we can't have `todo!()`s in production code:
```
println!("================================================");
println!("Error: {:?}", error);
println!("Error: {:?}", error.source());
todo!("Unhandled error – aborting");
```

wp_api/src/reqwest_request_executor.rs
```diff
@@ -143,3 +159,227 @@ impl RequestExecutor for ReqwestRequestExecutor {
...
+            }) => RequestExecutionErrorReason::InvalidSslError {
+                inner: InvalidSslError::CertificateNotValidForName {
+                    hostname: expected.to_str().to_string(),
+                    presented_hostnames: presented.to_vec(),
+                },
+            },
+            _ => RequestExecutionErrorReason::GenericError {
+                error_message: error.to_string(),
+            },
+        }
```
oguzkocer, line 262: Simplified this from the original implementation as suggested by clippy. It should be functionally the same as what the original PR had:

```
match error {
    TlsError::InvalidCertificate(certificate_error) => match certificate_error {
        CertificateError::NotValidForNameContext {
            expected,
            presented,
        } => {
            crate::RequestExecutionErrorReason::InvalidSslError {
                inner: InvalidSslError::CertificateNotValidForName {
                    hostname: expected.to_str().to_string(),
                    presented_hostnames: presented.to_vec(),
                },
            }
        }
        _ => RequestExecutionErrorReason::GenericError {
            error_message: error.to_string(),
        },
    },
    _ => RequestExecutionErrorReason::GenericError {
        error_message: error.to_string(),
    },
}
```

wp_api/src/ssl.rs
jkmassel: u64 -> i64 should always have a loss of precision, no?

oguzkocer: Technically yes, practically no because this is a date. Regardless, there is no way to avoid that because the function we use [chrono::DateTime::from_timestamp](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_timestamp) accepts `i64`.

This is actually why I preferred having the cast here as we know that `date_time` is a valid date and can be cast without a problem. If we add a `From<u64>` implementation then it's less clear how it should be handled.

oguzkocer:
Closed in favor of #661.

PR #666, opened 2025-04-11
Login spec tests in Rust

oguzkocer:
Uses #610 as a template to build on. Some of the tests in #610 were marked as `TODO`, in which case I followed the Kotlin test implementation.

The PR adds a flexible `MockRequestExecutor` implementation. It also adds some `WpNetworkResponse` helpers similar to the ones in Kotlin.

It removes the previous login tests since the spec tests cover the same basis.


wp_api_integration_tests/tests/test_login_remote.rs
```diff
@@ -0,0 +1,443 @@
...
+#[parallel]
+async fn login_spec_2_local_development_environment() {
+    // Spec Example 2
+    let executor = MockExecutor::with_execute_fn(|request| match request.url().0.as_str() {
+        "http://localhost/" | "https://localhost/" => {
+            Ok(response_helpers::with_api_root("http://localhost/wp-json"))
+        }
+        "http://localhost/wp-json" | "https://localhost/wp-json" => {
+            Ok(response_helpers::json_response_from_path(Path::new(
+                "../native/swift/Tests/wordpress-api/Resources/Responses/localhost-json-root.json",
```
oguzkocer, line 41: Kotlin, Swift & Rust tests are using this path. We should move it to a common place, but I wanted to separate that into its own PR.

wp_api_integration_tests/tests/test_login_remote.rs
```diff
@@ -0,0 +1,443 @@
...
+
+#[tokio::test]
+#[parallel]
+async fn login_spec_16_invalid_url() {
+    // Spec Example 16
+    assert!(matches!(
+        login_err("https://valid-looking-url-but-not-actually.foo")
+            .await
+            .to_find_api_root_failure(),
+        FindApiRootFailure::FetchHomepage { .. }
```
jkmassel: This should be `NonExistentSiteError`

jkmassel (changes requested):
Looks good – just one adjustment required

wp_api_integration_tests/tests/test_login_remote.rs
oguzkocer: Should be addressed in b19d26fd4b3f8aed0038675746989fbc502cf09f.

jkmassel (approved):

PR #678, opened 2025-04-15
Adds CLI tool for api discovery

oguzkocer:
Builds on @jkmassel's prototype to add a CLI tool to run api discovery. It supports passing an individual site or a csv file for batch auto discovery. I've kept the output mostly the same, but updated how the api discovery result is used and also introduced `tokio_stream` to be able to batch the requests.


jkmassel (approved):
I've run this against a couple of test sets and it appears to work well.

There are some potential areas for improvement, but they're related to the library not the CLI so I think we can chase those down separately.

PR #680, opened 2025-04-15
Implement XML-RPC endpoint discovery

crazytonyli:
The XML-RPC endpoint discovery process uses the results of the REST API discovery. That means we can be sure that the site is a WordPress site and use the `wp-json` results in the XML-RPC discovery process.

The XML-RPC endpoint URL can be found in the process below:
1. Parse the site homepage HTML to get an RSD link, which is typically `<site-url>/xmlrpc.php?rsd`
2. Parse the RSD link content, which is an XML, to get the XML-RPC endpoint URL.

We'll check the discovered URL and the default `<site-url>/xmlrpc.php` (if they are not the same). We can verify if it's disabled by sending a `system.listMethods` XML-PRC call.

Here are a couple of examples.
```
// Success
$ curl -s --data-raw '<?xml version="1.0"?><methodCall><methodName>system.listMethods</methodName></methodCall>' https://www.nasa.gov/xmlrpc.php

// Failure
$ curl -s --data-raw '<?xml version="1.0"?><methodCall><methodName>system.listMethods</methodName></methodCall>' https://www.whitehouse.gov/xmlrpc.php
```


wp_api_integration_tests/tests/test_login_remote.rs
```diff
@@ -352,6 +353,30 @@ async fn login_spec_17_invalid_https_with_exception_works() {
...
+    // curl: (92) HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)
+    let result = xmlrpc_url("https://xmlrpc-disabled.wpmt.co").await;
+    assert!(result.is_err());
+    assert!(matches!(
+        result.unwrap_err(),
+        XMLRPCDiscoveryError::Disabled {
+            reason: XMLRPCDisabledReason::ByHost
+        }
+    ));
+}
```
crazytonyli, line 383: I want to add another test case for the `XMLRPCDisabledReason::ByPlugin` error. I haven't configured the test site yet, but will add one once the test site is ready.

oguzkocer (approved):
Looks good. I could offer minor suggestions for Rust, but I don't think any of them are worth holding up the PR, so I'd rather we merge it sooner rather than later.

PR #681, opened 2025-04-16
Add middleware support to the rest api client types

crazytonyli:
This PR adds the middleware support in `WpLoginClient` to other API client types. The purpose of this change is to support using a middleware to handle invalid application passwords, which I plan to implement on the app side, so that the app can show a re-authentication UI if users revokes the application password that the app is using.


oguzkocer (approved):
From the diff it looks about right.

PR #682, opened 2025-04-17
Adds user agent for Rust and Kotlin clients

oguzkocer:
Also addresses issues in the testing setup due to WordPress `6.8` and fixes an issue related to setting the user is agent for `reqwest` executor.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,10 +17,10 @@ let libwordpressFFI: Target = libwordpressFFIVersion.target
 var package = Package(
     name: "WordPressAPI",
     platforms: [
-        .iOS(.v15),
+        .iOS(.v16),
         .macOS(.v12),
         .tvOS(.v16),
-        .watchOS(.v8)
+        .watchOS(.v9)
     ],
     products: [
         .library(
--- a/native/kotlin/api/kotlin/src/integrationTest/kotlin/MediaEndpointTest.kt
+++ b/native/kotlin/api/kotlin/src/integrationTest/kotlin/MediaEndpointTest.kt
@@ -72,7 +72,7 @@ class MediaEndpointTest {
             requestBuilder.media().create(
                 params = MediaCreateParams(title = title),
                 "test_media.jpg",
-                "image/jpg"
+                "image/jpeg"
             )
         }.assertSuccessAndRetrieveData().data
         assertEquals(title, response.title.rendered)
--- a/native/swift/Tests/wordpress-api/UserAgentTests.swift
+++ b/native/swift/Tests/wordpress-api/UserAgentTests.swift
@@ -27,7 +27,7 @@ struct UserAgentTests {
     #elseif os(Linux)
     @Test("User agent contains architecture")
     func testThatUserAgentContainsArchitecture() throws {
-        #expect(UserAgent.postfix.contains("aarch64"))
+        #expect(UserAgent.postfix.contains("aarch64") || UserAgent.postfix.contains("x86_64")) // CI is `x86_64`
     }
     #endif
 }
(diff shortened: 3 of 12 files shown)


native/kotlin/api/kotlin/src/integrationTest/kotlin/MediaEndpointTest.kt
```diff
@@ -72,7 +72,7 @@ class MediaEndpointTest {
             requestBuilder.media().create(
                 params = MediaCreateParams(title = title),
                 "test_media.jpg",
-                "image/jpg"
+                "image/jpeg"
```
oguzkocer, line 75: It's a little concerning that `image/jpg` no longer works in WordPress `6.8`. It might be worth looking into implementing a `SupportedFileContentType` `enum` to make it easier for clients to set this field.

cc @jkmassel

jkmassel: Is this part of the release notes, or just observed behaviour?

This feels more like a server thing...

oguzkocer: It was observed behaviour. Without this change, create media test is failing with:

> WpError(errorCode=uniffi.wp_api.WpErrorCode$CustomException: v1=rest_upload_image_type_not_supported, errorMessage=The web server cannot generate responsive image sizes for this image. Convert it to JPEG or PNG before uploading., statusCode=400, response={"code":"rest_upload_image_type_not_supported","message":"The web server cannot generate responsive image sizes for this image. Convert it to JPEG or PNG before uploading.","data":{"status":400}})

I believe it's a new `WpError` type, because we don't have an enum variant for it yet.

oguzkocer: It's probably related to [this change](https://make.wordpress.org/core/2025/03/25/miscellaneous-developer-changes-in-wordpress-6-8/#media-control-uploading-of-unsupported-media-typesin-6-8-a-new-filter-wp-prevent-unsupported-mime-type-uploads-controls-the-behavior-of-the-editor-and-media-library-when-users-upload-an-image-type-that-the-serve-does-not-support-by-default-users-will-see-an-error-message-that-this-image-cannot-be-processed-by-the-web-server-convert-it-to-jpeg-or-png-before-uploading-developers-can-return-false-from-the-filter-to-enable-uploading-of-these-images-however-sub-sized-images-will-not-be-created-61167).

jkmassel: Ah yep – [looks like a bug](https://make.wordpress.org/core/2025/03/25/miscellaneous-developer-changes-in-wordpress-6-8/#comment-47628)

I'm fine to keep our tests passing with this fix though

Package.swift
```diff
@@ -17,10 +17,10 @@ let libwordpressFFI: Target = libwordpressFFIVersion.target
 var package = Package(
     name: "WordPressAPI",
     platforms: [
-        .iOS(.v15),
+        .iOS(.v16),
```
jkmassel, line 20: I'm not a big fan of bumping the minimum Swift version just for tests, but iOS 15 doesn't support Swift Regex literals. The fact that Swift features aren't back ported that far makes me think nobody needs that.

This would be easy enough to walk back, so I don't mind making the change and we can re-evaluate later.

native/swift/Tests/wordpress-api/UserAgentTests.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import Testing
+import WordPressAPI
+
+@Suite("User Agent Tests")
+struct UserAgentTests {
+    #if canImport(Darwin)
+    @Test("User agent contains bundle name and version")
+    func testThatDefaultUserAgentContainsBundleNameAndVersion() throws {
+        #expect(UserAgent.postfix.contains(/xctest\/(\d+.\d+)/))
```
jkmassel, line 10: It's feasible that Apple ships `xctest/x.y.z` at some point, but I figure we can address that down the road if it happens

jkmassel (approved):

PR #684, opened 2025-04-20
Don't allow `None` timeout for `ReqwestRequestExecutor`

oguzkocer:
Previous implementation ignores `None` values and sets it to `10` seconds by default. If we want to have a default value, this is not the correct way to do it.

Since this issue is causing the nightly tests to break, I'd like to admin merge the simple solution, but I'd be happy to implement a default timeout if you'd like @jkmassel.

### DIFF
--- a/wp_api/src/reqwest_request_executor.rs
+++ b/wp_api/src/reqwest_request_executor.rs
@@ -15,6 +15,8 @@ use reqwest::multipart::Part;
 use rustls::{CertificateError, Error as TlsError};
 use std::{error::Error, sync::Arc, time::Duration};
 
+const DEFAULT_TIMEOUT: u64 = 10;
+
 #[derive(Debug)]
 pub struct ReqwestRequestExecutor {
     client: reqwest::Client,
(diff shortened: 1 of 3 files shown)


jkmassel (changes requested):
I don't think not having a default is feasible? At some point the connection should time out because it's just never going to work.

10 seconds is a _lot_ of time, but we could try raising it to 30 seconds. I'd imagine that's like 95th percentile for server timeout configurations?

WDYT?

oguzkocer:
In my opinion, since this is a `reqwest` executor which doesn't have a default timeout, I think we should allow clients to opt into that behavior. So, if it was up to me, I'd keep this behavior. Having said that, if you feel strongly about any other behavior, I can implement that. I don't think this is worth arguing over.

One strong objection I have is that we don't do this:
```
pub fn new(danger_accept_invalid_certs: bool, timeout: Option<Duration>) -> Self {
        Self {
            client: reqwest::Client::builder()
                .timeout(timeout.unwrap_or(Duration::from_secs(10))) <--- This
                ...
        }
    }
```

If the argument is an `Option`, we shouldn't unwrap it and set a default value. If we always want a timeout, this method should take a `Duration` and another method could set the default timeout of our choice. (for example, `new_with_default_timeout`)

jkmassel:
> this method should take a Duration and another method could set the default timeout of our choice. (for example, new_with_default_timeout)

No objection to this from my side.

One suggestion for an objective approach to this – try removing the timeout completely and running against one of the CSV test sets? I believe there were enough mis-configured sites in there that some of them would take multiple minutes to return. I don't think that's really feasible for the CLI or web tools. 

If it's not a problem in those contexts, we could revisit whether it's necessary. FWIW AFAIK `URLSession` uses 60 seconds by default and `OkHttp` uses 10 seconds by default. I'm rather convinced that you need _some_ value for the timeout, otherwise you could point the tool at a malicious endpoint that just keeps connections open for ages and then your tool is broken. I'm not as fussed about what the specific timeout is though – I think both `URLSession` and `OkHttp` defaults are reasonable.

oguzkocer:
> One suggestion for an objective approach to this – try removing the timeout completely and running against one of the CSV test sets? I believe there were enough mis-configured sites in there that some of them would take multiple minutes to return. I don't think that's really feasible for the CLI or web tools.

I've already done this before opening the PR. Having said that, I understand your argument about forcing clients to have some sort of timeout. As long as the method signature matches that, I am good with that. I'll make the change.

jkmassel (approved):

PR #689, opened 2025-04-22
Remove the user id argument from the application password introspect API

crazytonyli:
Since the application password introspect API can only be used on the current user, I think we should remove the user id argument and hard-code the current user instead.

### DIFF
--- a/native/kotlin/api/kotlin/src/integrationTest/kotlin/ApplicationPasswordsEndpointTest.kt
+++ b/native/kotlin/api/kotlin/src/integrationTest/kotlin/ApplicationPasswordsEndpointTest.kt
@@ -36,7 +36,7 @@ class ApplicationPasswordsEndpointTest {
     @Test
     fun testApplicationPasswordRetrieveCurrentRequest() = runTest {
         val applicationPasswordList = client.request { requestBuilder ->
-            requestBuilder.applicationPasswords().retrieveCurrentWithEditContext(FIRST_USER_ID)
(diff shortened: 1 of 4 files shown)


oguzkocer (approved):
This sounds reasonable to me. Probably best to double check with @jkmassel as well just in case we are missing something.

Kotlin issue is probably about an integration test similar to the ones you updated for Rust. Happy to help with it if you need.

jkmassel:
So you're saying the `/introspect` API is locked to the current user? (ie – no alternative user ID will work – it has to be the same one as the current user?)

What about [the other API](https://developer.wordpress.org/rest-api/reference/application-passwords/#retrieve-a-application-password-3) that looks up by UUID?

oguzkocer:
> What about [the other API](https://developer.wordpress.org/rest-api/reference/application-passwords/#retrieve-a-application-password-3) that looks up by UUID?

According to [this test](https://github.com/Automattic/wordpress-rs/blob/aa3e88f0bebabebc19765a622a0afe0a8ff8b672/wp_api_integration_tests/tests/test_application_passwords_err.rs#L27-L42) that [uses that endpoint](https://github.com/Automattic/wordpress-rs/blob/aa3e88f0bebabebc19765a622a0afe0a8ff8b672/wp_api/src/request/endpoint/application_passwords_endpoint.rs#L27-L28), a user can't retrieve another user's application passwords.

jkmassel (approved):

PR #690, opened 2025-04-22
Adds `application_passwords_authentication_url` field to `AutoDiscoveryAttemptSuccess`

oguzkocer:
Until now, even though api discovery fails if the application passwords authentication url is missing, we still had to use the `find_application_passwords_authentication_url` function that returns `Option<String>` and manually unwrap it.

This PR adds `application_passwords_authentication_url` as a non-optional field to `AutoDiscoveryAttemptSuccess` to avoid this. This change will also force a compiler error if we decide to change this behavior whereas unwrapping `find_application_passwords_authentication_url` might have resulted in runtime crashes.


jkmassel (approved):
Solid improvement!

PR #693, opened 2025-04-23
Introduce `WpAppNotifier`

oguzkocer:
Introduces `WpAppNotifier` trait with the `requested_with_invalid_authentication` function. This will get triggered by request executor if the response status is `401` and the newly added `is_valid_authentication` returns `IsValidAuthenticationResult::Unauthorized`.

The new `is_valid_authentication` function makes a request to [`GET /wp/v2/users/<user_id>)/application-passwords/introspect` ](https://developer.wordpress.org/rest-api/reference/application-passwords/#retrieve-a-application-password-2) and return either `Authenticated`, `Unauthorized`, `ErrWpApiError` or `ErrRequestExecutionError`. We only return `Authenticated` or `Unauthorized` if we parse the response and find what we are looking for.

Also, each request executor now has a `has_valid_authentication` function which internally calls the `is_valid_authentication` to make it easy to manually check the authentication state.


wp_api/src/lib.rs
```diff
@@ -212,6 +213,27 @@ pub enum IntegerOrString {
...
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, uniffi::Enum)]
+pub enum WpAppNotification {
+    UnauthorizedRequest,
+}
+
+impl WpAppNotification {
+    pub fn from_wp_network_response(response: &WpNetworkResponse) -> Option<Self> {
+        if response.status_code == 401 {
+            Some(Self::UnauthorizedRequest)
```
crazytonyli: For the purpose of finding out whether the current application password is still valid, only a 401 status code check is not enough, since not all 401 responses mean that an invalid application password is used.

We'll probably want to use a stricter check like

```rust
fn is_unauthorized_error(response: &WpNetworkResponse) -> bool {
    if response.status_code == 401 {
        if let Some(original_error) = WpError::try_parse(&response.body.clone()) {
            return original_error.code == WpErrorCode::Unauthorized;
        }
    }

    false
}
```

wp_api_integration_tests/tests/test_app_notifier_immut.rs
```diff
@@ -0,0 +1,63 @@
...
+    assert!(
+        notifier.assertion.load(Ordering::Relaxed),
+        "Failed to notify with `WpAppNotification::UnauthorizedRequest`"
+    );
+}
+
+fn api_client_as_unauthenticated_with_notifier(notifier: Arc<FooAppNotifier>) -> WpApiClient {
+    WpApiClient::new(
+        test_site_url(),
+        WpAuthentication::None,
```
crazytonyli: This test set up is not suitable for the purpose of detecting an invalid application password. We should pass a hard-coded username and _invalid_ application password pair here.

oguzkocer: Addressed in d3b589560d191f146b1da64a8c91282f91b3225d.

wp_api/src/lib.rs
oguzkocer: I don't think we always get back `WpErrorCode::Unauthorized`. For example,

```
> curl --user test@example.com:invalid "http://localhost/wp-json/wp/v2/users?context=edit"
> {"code":"rest_forbidden_context","message":"Sorry, you are not allowed to list users.","data":{"status":401}}
```

So, at least for now, I have changed the condition to be true for both `WpErrorCode::Unauthorized` and `WpErrorCode::ForbiddenContext` in 80859035c89daccdf9bb3ab1982ff75ffad1aeab. Let me know what you think!

crazytonyli: Maybe we should check whether the HTTP response status code is 401 instead? So far, it looks like the core would always return a 401 status code, but may return a different `{"code": ...}` in the response JSON.

oguzkocer: The initial implementation was checking for `401`, but I tried updating it as per your previous suggestion. I agree though, I think we should only check the status code since we are going to make a request to check if the application password is valid.

crazytonyli: Ah right. The original approach of creating `WpAppNotification` from HTTP response probably won't work. But I see it's updated to invoke a callback after application password check. 👍

wp_api/src/lib.rs
```diff
@@ -212,6 +215,32 @@ pub enum IntegerOrString {
     String(String),
 }
 
+#[uniffi::export(with_foreign)]
+#[async_trait::async_trait]
+pub trait WpAppNotifier: Send + Sync + std::fmt::Debug {
+    async fn unauthorized_request(&self);
```
crazytonyli: Should this be more specific, like `authentication_becomes_invalid`?

wp_api/src/lib.rs
```diff
@@ -212,6 +215,32 @@ pub enum IntegerOrString {
...
+    authentication: crate::WpAuthentication,
+) -> bool {
+    if status_code == 401
+        && (wp_error_code == &WpErrorCode::Unauthorized
+            || wp_error_code == &WpErrorCode::ForbiddenContext)
+    {
+        let r = ApplicationPasswordsRequestBuilder::new(api_root_url, authentication)
+            .retrieve_current_with_edit_context()
+            .into();
+        request_executor.execute(r).await.is_ok()
```
crazytonyli: Technically, receiving a `RequestExecutorError` does not mean the request is unauthorized. I wonder if we should explicitly check the `WpErrorCode` instead here?

wp_derive_request_builder/src/generate.rs
```diff
@@ -78,7 +80,17 @@ fn generate_async_request_executor(
                 pub async #fn_signature -> Result<#response_type_ident, #error_type> {
                     use #crate_ident::middleware::PerformsRequests;
                     #request_from_request_builder
-                    self.perform(std::sync::Arc::new(request)).await?.parse()
+                    let response = self.perform(std::sync::Arc::new(request)).await?;
+                    let response_status_code = response.status_code;
+                    let parsed_result: Result<#response_type_ident, #error_type> = response.parse();
+                    if let Err(ref err) = parsed_result {
+                        if let Some(wp_error_code) = err.wp_error_code() {
+                            if #crate_ident::is_unauthorized_request(response_status_code, wp_error_code, self.request_executor.clone(), self.api_root_url.clone(), self.authentication.clone()).await {
```
crazytonyli: As discussed above, these if statements can be changed to `if is_unauthorized_request == 401 && #crate_ident::is_unauthorized_request(...) {`

wp_api/src/lib.rs
oguzkocer: Addressed in 6bbb9bb80dd1a9712b8ffa751155cb33b7f18f30.

oguzkocer: `becomes_invalid` may be a little misleading because we currently don't check if the authentication is valid when we first build the client. Maybe we should go with `requested_with_invalid_authentication`?

crazytonyli: Sounds good to me!

oguzkocer: Addressed in a93523cd4064b3adfa52ea0ef3d61120152be51d.

wp_api/src/request.rs
```diff
@@ -753,6 +756,62 @@ pub fn is_valid_json(value: impl AsRef<str>) -> bool {
...
+) -> IsValidAuthenticationResult {
+    let request = ApplicationPasswordsRequestBuilder::new(api_root_url, authentication_provider)
+        .retrieve_current_with_edit_context()
+        .into();
+    match request_executor.execute(request).await {
+        Ok(response) => {
+            let parsed_res: Result<
+                ApplicationPasswordsRequestRetrieveCurrentWithEditContextResponse,
+                WpApiError,
+            > = response.parse();
```
crazytonyli: I _think_ you can reuse the `ApplicationPasswordsRequestExecutor` to do this parsing for you?

wp_api/src/request.rs
```diff
@@ -753,6 +756,62 @@ pub fn is_valid_json(value: impl AsRef<str>) -> bool {
     serde_json::from_str::<serde_json::Value>(value.as_ref()).is_ok()
 }
 
+#[uniffi::export]
```
crazytonyli, line 759: Nitpick: I don't think this function needs to be exported? I presume we could change the arguments to `&Arc<...>` and avoid some `clone`s.

wp_api/src/request.rs
oguzkocer: That'd potentially introduce a recursion here, because `ApplicationPasswordsRequestExecutor` will also potentially trigger the method. We also may introduce other behavior, middleware etc that impact what `ApplicationPasswordsRequestExecutor` does. Doing it this way avoids all of that, because we only build the request and manually execute it.

wp_api/src/request.rs
oguzkocer: All of the clones here are `Arc`s so there is no downside to cloning them.

I think having the function exported is a good thing. In fact, it'd be good to have some integration tests for these at least in Kotlin because it's now an important part of the flow.

wp_api/src/request.rs
crazytonyli: This reminds me that we don't need to send an introspect API call if the original response is `WpErrorCode::Unauthorized`.

oguzkocer: This should address that: 99990ac904542c14f7c1af03d033c712d1eebf20

crazytonyli (approved):

PR #695, opened 2025-04-24
Introduce `WpApiClientDelegate`

oguzkocer:
Introduces `WpApiClientDelegate` as a container for trait implementations that we pass into `WpApiClient`. It also introduces `WpAuthenticationProvider` which allows static, dynamic and modifiable authentication strategies.

Dynamic authentication allows modifying authentication per request whereas modifiable authentication allows updating the authentication after it's set as it's backed by a `Mutex`. Dynamic authentication is flexible enough to build a modifiable authentication on top of it, but I think it's better to have a separate one for it. That's because with dynamic authentication, the credentials have to pass FFI for every request which shouldn't cost _much_, but it's also not free. On the other hand, with the modifiable authentication, it only passes FFI when it's updated.

This setup allows clients to choose whichever authentication method is best suited for their needs and will most likely cover all use cases. We might end up adding a few more helpers, but my expectation is that we won't have to re-visit this in the future. 🤞


jetpack/src/connection.rs
```diff
@@ -179,7 +166,7 @@ impl JetpackConnectionClient {
 
     pub async fn connect_user(
         &self,
-        wp_com_authentication: WpAuthentication,
+        auth_provider: Arc<WpAuthenticationProvider>,
```
crazytonyli: Unlike the authentication property in `WpApiClient` stored in the API client instance, this `wp_com_authentication` argument is discarded after use. I think we can leave it as `WpAuthentication`? Internally, we can wrap the `WpAuthentication` argument in a static auth `WpAuthenticationProvider`, which is then passed to  `WpApiClientDelegate`?

wp_api/src/api_client.rs
```diff
@@ -164,6 +146,18 @@ impl WpApiClient {
...
 
+// IMPORTANT: This type may be aggressively cloned, so all of its fields must be cheap to clone!
+#[derive(Clone, uniffi::Record)]
+pub struct WpApiClientDelegate {
+    pub auth_provider: Arc<WpAuthenticationProvider>,
+    pub request_executor: Arc<dyn RequestExecutor>,
+    pub middleware_pipeline: Arc<WpApiMiddlewarePipeline>,
+}
+
+pub trait IsRequestExecutor {
```
crazytonyli: Should this type be `IsWpApiClientDelegate`?

wp_api/src/auth.rs
```diff
@@ -0,0 +1,132 @@
...
+impl ModifiableAuthenticationProvider {
+    #[uniffi::constructor]
+    pub fn new(authentication: WpAuthentication) -> Self {
+        Self {
+            auth: Mutex::new(authentication),
+        }
+    }
+
+    pub fn set_authentication(&self, new_authentication: WpAuthentication) {
+        *self.auth.lock().unwrap() = new_authentication;
```
crazytonyli: Use `expect` instead of `unwrap`?

jetpack/src/connection.rs
oguzkocer: I wasn't quite sure about the changes to the `wpcom` and `jetpack` package changes and didn't get a chance to fully investigate. I'll take a brief look at it tomorrow and will likely follow your advice.

wp_api/src/api_client.rs
oguzkocer: I thought we might put some more request executor related methods here in the future which is why I picked this name, but I don't feel strongly about it 🤷‍♂️

wp_api/src/auth.rs
oguzkocer: Right.. I looked into this a little bit last week and it looks like unwrapping a lock is the standard practice since in most cases there isn't a clear way to resolve a poison. However, we should use `expect` here as you suggest and put this explanation into the expect message 🤔

wp_api/src/auth.rs
```diff
@@ -0,0 +1,132 @@
...
+            Self::Bearer { token } => {
+                Some(HeaderValue::from_str(&format!("Bearer {}", token)).expect("It shouldn't be possible to build WpAuthentication::Bearer with an invalid token"))
+            }
+        }
+    }
+}
+
+#[derive(uniffi::Object)]
+pub struct ModifiableAuthenticationProvider {
+    auth: Mutex<WpAuthentication>,
```
crazytonyli: The value here is going to be read way more often than written. Should we use a `RwLock` here?

crazytonyli:
It could be useful if we can support `apiClient.authProvider.as_modifiable().authentication`:

```swift
let apiClient = UniffiWpAPiClient(authProvider: .modifiable(...), ...)
apiClient.authProvider.as_modifiable().authentication = .init(username: "user", password: newPassword)
```

The app can work without this API. They just need to hold on to the reference to the `ModifiableAuthenticationProvider` instance. But I thought it'd be nice if they don't have to do that.

To support this, we'll need add two uniffi APIs:
- `UniffiWpAPiClient.authProvider() -> Arc<WpAuthenticationProvider>`
- `WpAuthenticationProvider.as_modifiable() -> Arc<ModifiableAuthenticationProvider>`

What do you think?

crazytonyli:
⬆️ I have implemented what I wanted using a dynamic auth provider. So, the comment is not a high priority issue to address.

oguzkocer:
> `WpAuthenticationProvider.as_modifiable() -> Arc<ModifiableAuthenticationProvider>`

There are a couple issues with this function signature:

* It'd have to return an `Option`, because you can't turn a dynamic authentication provider into a modifiable one.
* It breaks the design principles `WpAuthenticationProvider` is built around. Specifically, in its current state, you have to pick a single authentication provider type and stick with it. If it's allowed to be modified, then it's not any different from making the authentication itself modifiable making `WpAuthenticationProvider` mostly obsolete.

---

In the current design, I believe the correct way to achieve what you want is to use a wrapper type:

```
struct WpApiClientWithModifiableAuthentication {
    client: WpApiClient,
    authentication: ModifiableAuthenticationProvider,
}
```

jetpack/src/connection.rs
oguzkocer: Addressed in cb53bcf433ff08453681df59bf219c4fd0224004.

wp_api/src/api_client.rs
oguzkocer: Addressed in 40ce5c7411da35a500dff4d7fec3178c4dfe0612.

wp_api/src/auth.rs
oguzkocer: Addressed in b770401fe343d903ef137b2b8336b0f9d1cae8bf.

wp_api/src/api_client.rs
oguzkocer: I missed one reference in my original commit which is addressed in 991a727d08d0ce97402a6656fc2573083aab211e.

wp_api/src/auth.rs
oguzkocer: Good catch! Addressed in e7b9569b38a4b0996fafa914bbd6084ced882b7b.

oguzkocer: I am actually not 100% sure this is the right choice because if there are too many requests then the writer might not get a chance. It's especially a problem if someone uses a middleware or something to trigger a lot of retries.

Mutex is a lot better for that and the extra cost shouldn't be too significant for an async api client. 🤔 So, I'd probably go with `Mutex` if it was only up to me, just because I don't want to think about whether `RwLock` would cause issues or not depending on the future design changes 😅

jetpack/src/connection.rs
```diff
@@ -179,7 +167,7 @@ impl JetpackConnectionClient {
 
     pub async fn connect_user(
         &self,
-        wp_com_authentication: WpAuthentication,
+        auth: WpAuthentication,
```
crazytonyli: I used the argument name `wp_com_authentication` because this `authentication` is used to authenticate WP.com API calls. I'm not sure about Kotlin, this argument name becomes part of the function signature in the Swift binding, which means there is less chance that the apps pass in a self-hosted site authentication instance here. So, I think there is still value in keeping the original name, or anything that says "the `WpAuthentication` argument is for WP.com".

oguzkocer: The argument name change was unintentional. I reverted it back in d7a81d8552542219fb3e0153687eabfb97eb65f1.

wp_api/src/auth.rs
```diff
@@ -0,0 +1,138 @@
...
+    StaticAuthenticationProvider {
+        auth: WpAuthentication,
+    },
+    DynamicAuthenticationProvider {
+        inner: Arc<dyn WpDynamicAuthenticationProvider>,
+    },
+    Modifiable {
+        inner: Arc<ModifiableAuthenticationProvider>,
+    },
+}
```
jkmassel, line 88: I'm not convinced we need static/dynamic/modifiable?

The app layer should be able to handle this – if we want to fetch from some backing store constantly, that's fine. If we want to just keep a value in memory and have it returned for every request, it can do that too.

But having the complexity here IMHO isn't super valuable?

crazytonyli (approved):

wp_api/src/api_client.rs
```diff
@@ -164,6 +146,18 @@ impl WpApiClient {
     }
 }
 
+// IMPORTANT: This type may be aggressively cloned, so all of its fields must be cheap to clone!
+#[derive(Clone, uniffi::Record)]
+pub struct WpApiClientDelegate {
```
jkmassel, line 151: This is more of a data source object than a delegate – I'd expect a delegate to be more like #693 (where it says "this thing happened, and you need to do something about it") instead of a set of configuration parameters.

I'd propose changing this to  `WpApiClientDataSource` or `WpApiClientConfiguration`

jkmassel (changes requested):
I mostly like how this turned out, but am nitpicking about naming. Keen to hear your thoughts.

wp_api/src/auth.rs
oguzkocer: We need at least two of the three and having the third doesn't add any extra complexity. In fact, this setup is fairly straightforward, so I don't think we need to be worried about it.

wp_api/src/api_client.rs
oguzkocer: We picked this name together on Slack 😅 I like configuration more, because even though it's not directly a delegate, it still sort of acts like one because it contains `request_executor`.

Regardless, I don't feel strongly about any name, so if you pick the name, I am happy to open a new PR to update it. However, I'll probably do it after #693 lands to avoid unnecessary conflicts.

jkmassel: Yeah, I was thinking about #693 when we chatted, and it didn't quite go that way.

I'm happy with the shape of it though for sure

wp_api/src/auth.rs
jkmassel: > We need at least two of the three 

Sorry, why is this? IINM `DynamicAuthenticationProvider` can do anything the other two can, no?

oguzkocer: I believe this is covered in the PR description, but if that's not clear, happy to discuss it further.

PR #700, opened 2025-04-28
Improve error messaging for some connection errors

jkmassel:
Some servers drop the HTTP 2 connection, and we didn't previously give a very helpful message in the web/cli debug tools:

> stream error received: unspecific protocol error detected

This PR handles HTTP 2 connection issues differently and provides an actionable suggestion:

> Unable to connect to server: {$reason}. Please contact your server provider.

For reference, on Apple platforms:
> The network connection was lost.

In Kotlin:
> No route to host

### DIFF
--- a/wp_localization/localization/tr-TR/main.ftl
+++ b/wp_localization/localization/tr-TR/main.ftl
@@ -22,6 +22,8 @@ http_authentication_required_error = {$url} adresindeki sunucu kimlik doğrulama
 
 http_authentication_rejected_error = {$url} adresindeki sunucu giriş bilgilerinizi reddetti. Lütfen geçerli bir kullanıcı adı ve şifre girin.
 
+http_server_error = Unable to connect to server: {$reason}. Please contact your server provider.
+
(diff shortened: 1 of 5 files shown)


wp_localization/localization/tr-TR/main.ftl
```diff
@@ -22,6 +22,8 @@ http_authentication_required_error = {$url} adresindeki sunucu kimlik doğrulama
 
 http_authentication_rejected_error = {$url} adresindeki sunucu giriş bilgilerinizi reddetti. Lütfen geçerli bir kullanıcı adı ve şifre girin.
 
+http_server_error = Unable to connect to server: {$reason}. Please contact your server provider.
+
```
oguzkocer:
```suggestion
```

oguzkocer (approved):

wp_localization/localization/tr-TR/main.ftl
jkmassel: Will this cause the localization validation to fail? 

I figured we could just have a base value here for translation later?

oguzkocer: No it won't. It'll output a warning and lower the translation percentage. So, it's best to not include any untranslated strings here.

PR #705, opened 2025-04-30
Properly handle IO-related HTTP errors

jkmassel:
I noticed that HTTP/1.1 requests didn't nicely handle `ConnectionReset` errors – the root issue is that we should be calling `IOError.to_string()` instead of `Error.to_string()` – that digs into the error graph to find the most specific explanation for what went wrong.

It also adopts the new `RequestExecutionErrorReason::HttpError` instead of `RequestExecutionErrorReason::GenericError`, which directs the user to contact their server provider.

**To Test**
The easiest way is to add `.http1_only()` to the CLI tool's `ClientBuilder`, then use the same test URL as #700. Before this PR, the error isn't helpful. After, it is.

### DIFF
--- a/wp_rs_cli/src/bin/wp_rs_cli/main.rs
+++ b/wp_rs_cli/src/bin/wp_rs_cli/main.rs
@@ -182,7 +182,7 @@ impl SimplifiedDiscoveryResult {
                 println!("{}", format!("Login URL found: {login_url}").green());
             }
             Err(error) => {
-                println!("{}", format!("Error: {}", error).red());
+                println!("{}", error.to_string().red());
             }
         }
     }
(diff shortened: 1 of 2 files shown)


wp_rs_cli/src/bin/wp_rs_cli/main.rs
```diff
@@ -182,7 +182,7 @@ impl SimplifiedDiscoveryResult {
                 println!("{}", format!("Login URL found: {login_url}").green());
             }
             Err(error) => {
-                println!("{}", format!("Error: {}", error).red());
+                println!("{}", error.to_string().red());
```
jkmassel, line 185: This is just a little thing – it was weird having the string read `Error: Unable to connect to server: ⁨Connection reset by peer`

It's printed in red anyway, so I think folks will get that it's an error

oguzkocer (approved):

PR #715, opened 2025-05-08
Implement uploading media in the Swift wrapper

crazytonyli:
The majority of the diff is from `MultipartForm.swift`, which is copied (with a small modification) from [WordPressKit](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/40a984ddc1f6775052f81b82565c1fb01d8198c5/Sources/CoreAPI/MultipartForm.swift).

I have added a `request_id` to the `MediaUploadRequest` struct, to make it consistent with the `WpNetworkRequest`: all HTTP requests should have the `X-Request-Id` header. This property can be useful to track media uploading progress, which will be implemented in a separate PR.

### DIFF
--- a/native/swift/Sources/wordpress-api/Extensions.swift
+++ b/native/swift/Sources/wordpress-api/Extensions.swift
@@ -12,7 +12,7 @@ public extension MiddlewarePipeline {
 }
 
 extension WpNetworkResponse {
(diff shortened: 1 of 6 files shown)


jkmassel (approved):
I tested this on macOS and iOS, and the upload was successful (along with the metadata)

PR #716, opened 2025-05-08
Introduces `NetworkRequestAccessor` trait

oguzkocer:
The trait is implemented by both `WpNetworkRequest` and `MediaUploadRequest` to offer a way to generalize some implementation logic in the client/wrapper side. We might end up using it in Rust as well, but I haven't looked into it in detail yet as this PR is mostly to help with Swift.

### DIFF
--- a/wp_api/src/request.rs
+++ b/wp_api/src/request.rs
@@ -173,11 +173,6 @@ pub struct WpNetworkRequest {
 
 #[uniffi::export]
 impl WpNetworkRequest {
-    /// Unique identifier for this request – used for external record-keeping
-    pub fn request_id(&self) -> String {
-        self.uuid.clone()
-    }
-
     /// Clones the request and increments the retry count
     pub fn clone_with_incremented_retry_count(&self) -> Self {
         Self {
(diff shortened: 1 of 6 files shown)


wp_api/src/request.rs
```diff
@@ -279,6 +262,34 @@ impl Debug for WpNetworkRequest {
     }
 }
 
+/// Unique identifier for this request – used for external record-keeping
+#[uniffi::export(with_foreign)]
+pub trait NetworkRequestDescriptor: Send + Sync {
+    fn request_id(&self) -> String;
+    fn method(&self) -> RequestMethod;
+    fn url(&self) -> WpEndpointUrl;
+    fn header_map(&self) -> Arc<WpNetworkHeaderMap>;
```
jkmassel, line 271: These don't seem related to the purpose for the trait?

oguzkocer: @jkmassel Could you explain what you mean?

To give some context, @crazytonyli asked to add more methods to this trait so that he can generalize some things in Swift between `WpNetworkRequest` & `MediaUploadRequest`.

jkmassel: Right, so `NetworkRequestDescriptor` seems to indicate "a way to describe the request".

Using it to "generalize some things in Swift between WpNetworkRequest & MediaUploadRequest" is a different thing though?

I wonder if the name is still correct in this case?

oguzkocer: I don't love the name either, but couldn't find a suggestion to improve yet 😅 

Our goal is to describe a request type without a body. Open for suggestions?

oguzkocer: As discussed on our call today, I've renamed it as `NetworkRequestAccessor`. We both agreed that this is still not the greatest name, but until we find something better we'll go with it.

jkmassel (approved):

PR #717, opened 2025-05-08
Support tracking media upload progress in Swift

crazytonyli:
Add a new `WordPressAPI.uploadMedia(...)` function to support tracking upload progress.

A note about using `URLRequest.httpBodyStream` or `URLSession.uploadTask(from: dataOrFileURL)`: The latter produces a much more accurate progress reporting result. When using `httpBodyStream`, the progress would jump from 0.1 to 0.95. When using `uploadTask(...)`, the progress increases steadily from 0 to 1.


Package.swift
```diff
@@ -40,6 +40,7 @@ var package = Package(
             path: "native/swift/Sources/wordpress-api",
             swiftSettings: [
                 .enableExperimentalFeature("StrictConcurrency"),
+                .define("PROGRESS_REPORTING_ENABLED", .when(platforms: [.iOS, .macOS, .tvOS, .watchOS]))
```
crazytonyli, line 43: The feature is implemented using Combine, which is only available on Apple platforms.

native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -97,6 +107,59 @@ public actor WordPressAPI {
...
+            fileContentType = mimeType
+        } else {
+            fileContentType = "application/octet-stream"
+        }
+
+        let cancellable = requestExecutor.progress(forRequestWithId: requestId.uuidString())
+            .flatMap { $0.publisher(for: \.fractionCompleted, options: .new) }
+            // The apps expect progress to be updated on the main thread. We probably should look into
+            // refactoring this part (starting from the app side).
+            .receive(on: DispatchQueue.main)
```
jkmassel: Right now, the apps shouldn't be expecting progress on _any_ thread because this isn't implemented yet, right?

So can we just have this dispatch on a background thread and add `.receive(on: DispatchQueue.main)` to the apps when we implement this?

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -94,6 +101,16 @@ public final class WpRequestExecutor: SafeRequestExecutor {
         #endif
     }
 
+#if PROGRESS_REPORTING_ENABLED
+    public func progress(forRequestWithId requestId: String) -> AnyPublisher<Progress, Never> {
+        NotificationCenter.default.publisher(for: RequestExecutorDelegate.didCreateTaskNotification)
```
jkmassel, line 118: IINM, apps should call this method then use `.receive(on: DispatchQueue.main)` right?

crazytonyli: The app does not need to call this method. It's declared `public` only because the `SafeRequestExecutor` is public.

The app should call the new `WordPressAPI.uploadMedia(...)` function if they want to observe the progress. They can call the usual `WordPressAPi.media.create(...)` function if they don't,

native/swift/Sources/wordpress-api/WordPressAPI.swift
crazytonyli: > Right now, the apps shouldn't be expecting progress on any thread because this isn't implemented yet, right?

My plan is to create a new `MediaServiceRemote` implementation that uses `WordPressAPI`. Its new instance will be placed right where the current XMLRPC one is used. That means the app should not need to change anything, and all code that invokes the `WordPressAPI` will be the same one that invokes the XMLRPC calls.

Now that I think about it, we can do something at the new `MediaServiceRemote` implementation to make the progress update on the main thread. I'll push a commit to remove this `receive` call.

jkmassel (changes requested):
I tried to use this, and couldn't make it work from SwiftUI (in the example app)

Can you implement this in the example app in a manner similar to https://github.com/Automattic/wordpress-rs/pull/412/commits/118aa61dfaf2d242449543af187a0d8fcf59cec4?

native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -97,6 +107,59 @@ public actor WordPressAPI {
...
+            .sink { [weak progress] fraction in
+                guard let progress else { return }
+                progress.completedUnitCount = Int64(fraction * Double(progress.totalUnitCount))
```
jkmassel: In my testing `totalUnitCount` is always zero, so I'm not sure what this block is doing?

crazytonyli:
@jkmassel I have added a "Add Media File" feature to the Example app in https://github.com/Automattic/wordpress-rs/pull/717/commits/eba4535444bee3adb5086f7d9ba0f0b6a4888046

native/swift/Sources/wordpress-api/WordPressAPI.swift
crazytonyli: The dispatch call is removed in https://github.com/Automattic/wordpress-rs/pull/717/commits/a3c9ffb82a80629b6fd5fdb0953d55e86e3a3bde.

jkmassel:
I put some suggestions in https://github.com/Automattic/wordpress-rs/pull/732 – feel free to land it here if they're good?

crazytonyli:
@jkmassel Looks good! I have merged it into this PR.

jkmassel (approved):
Works for me!

PR #719, opened 2025-05-09
Map api discovery result to a Kotlin sealed class

oguzkocer:
Maps api discovery result to a Kotlin sealed class to make it easier to work with it without consulting the documentation. It replaces some of the helpers in the Kotlin api discovery tests with simpler alternatives that's now possible with the new result type.

Note that I am still not quite sure whether we should use a sealed class or a [Kotlin Result type](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-result/) for this. Both approaches have advantages and disadvantages, so my hope is that we'll find what we prefer as we work in clients such as WordPress-Android.


adalpari (approved):
Thank you for the changes, they look good to me! 🚢 

❓ When the changes will be released so we can import the new version?

PR #721, opened 2025-05-11
Delete `/templates` endpoint

oguzkocer:
Implements [delete & trash `/templates`](https://developer.wordpress.org/rest-api/reference/wp_templates/#delete-a-template) endpoint following the existing conventions.

### DIFF
--- a/wp_api/src/templates.rs
+++ b/wp_api/src/templates.rs
@@ -161,7 +161,12 @@ pub struct SparseTemplate {
     #[WpContext(edit, embed, view)]
     pub author: Option<UserId>,
     #[WpContext(edit, view)]
-    pub modified: Option<bool>,
+    #[WpContextualOption]
+    #[serde(
+        default,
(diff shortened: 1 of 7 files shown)


wp_api/src/templates.rs
```diff
@@ -160,7 +161,7 @@ pub struct SparseTemplate {
     #[WpContext(edit, embed, view)]
     pub author: Option<UserId>,
     #[WpContext(edit, view)]
-    pub modified: Option<bool>,
+    pub modified: Option<BoolOrWpGmtDateTime>,
```
crazytonyli: Based on [the API doc](https://developer.wordpress.org/rest-api/reference/wp_templates/), it looks like we should treat boolean values (which I assume are always `false`) as an `Option<WpGmtDateTime>.none`.

> The date the template was last modified, in the site's timezone.
> JSON data type: string,
> Format: datetime ([details](https://core.trac.wordpress.org/ticket/41032))

crazytonyli (approved):

PR #733, opened 2025-05-14
Add support for the `/oauth2/token-info` endpoint

jkmassel:
Adds the ability for the apps to check if a WP.com token is valid or not. 

Tested against the actual endpoint, but that code isn't committed because we don't want to leak the token in question.

The tests should be sufficient for this.

### DIFF
--- a/native/swift/Sources/wordpress-api/Exports.swift
+++ b/native/swift/Sources/wordpress-api/Exports.swift
@@ -19,6 +19,8 @@ public typealias MiddlewarePipeline = WordPressAPIInternal.WpApiMiddlewarePipeli
 public typealias MiddlewarePipelineBuilder = WordPressAPIInternal.WpApiMiddlewarePipelineBuilder
 public typealias ApiDiscoveryAuthenticationMiddleware = WordPressAPIInternal.ApiDiscoveryAuthenticationMiddleware
 public typealias RetryAfterMiddleware = WordPressAPIInternal.RetryAfterMiddleware
+public typealias WpApiClientDelegate = WordPressAPIInternal.WpApiClientDelegate
+public typealias WpAppNotifier = WordPressAPIInternal.WpAppNotifier
 
 // MARK: - Login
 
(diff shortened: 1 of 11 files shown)


native/swift/Sources/wordpress-api/Exports.swift
```diff
@@ -129,3 +131,7 @@ public typealias WpSiteHealthTestsRequestExecutor = WordPressAPIInternal.WpSiteH
 
 // MARK: - WordPress.org
 public typealias WordPressOrgApiClient = WordPressAPIInternal.WordPressOrgApiClient
+
+// MARK: - WordPress.com
+public typealias WPComApiClient = WordPressAPIInternal.UniffiWpComApiClient
```
crazytonyli: Sorry if I'm repeating something you are already aware of, because I roughly remember you talked about similar things.

I don't think we should expose uniffi clients as a public API. We should always wrap them in a Swift type, like how `WordPressAPI` wraps `UniffiWpApiClient`. One major pitfall I found is that the `WpApiClientDelegate` instance passed to the uniffi API clients is not retained, and we have to manually retain them in the wrapper like `WordPressAPI`.

jkmassel: Yep you're right! Addressed in https://github.com/Automattic/wordpress-rs/pull/733/commits/3b950894cee4353b60a7be5feba8bdbd0e281dd9

native/swift/Tests/wordpress-api/Support/Mocks.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+import WordPressAPI
+
+final class MockAppNotifier: WpAppNotifier {
```
crazytonyli, line 4: This class is not used.

crazytonyli (approved):

native/swift/Tests/wordpress-api/Support/Mocks.swift
jkmassel: I don't have a good method for injecting a secret WP.com token quite yet without making it public, but I made a test class to debug this endpoint with, and I used `MockAppNotifier` there.

So I'm using the test helper, just not in the repo code yet 😅

PR #734, opened 2025-05-14
Comments related improvement to the Swift wrapper

crazytonyli:
Some small changes to make comments easier to work with for the app.

### DIFF
--- a/native/swift/Sources/wordpress-api/Extensions.swift
+++ b/native/swift/Sources/wordpress-api/Extensions.swift
@@ -32,3 +32,23 @@ extension MiddlewarePipeline {
         self.init(middlewares: middlewares)
     }
 }
+
+public extension CommentStatus {
+    init(status: String) {
+        self = commentStatusFromString(value: status)
+    }
+
+    func status() -> String {
+        commentStatusToString(status: self)
+    }
+}
+
+public extension CommentType {
+    init(type: String) {
+        self = commentTypeFromString(value: type)
+    }
+
+    func type() -> String {
+        commentTypeToString(commentType: self)
+    }
+}
--- a/native/swift/Sources/wordpress-api/Extensions.swift
+++ b/native/swift/Sources/wordpress-api/Extensions.swift
@@ -43,6 +43,12 @@ public extension CommentStatus {
     }
 }
 
+extension CommentStatus: ExpressibleByStringLiteral {
+    public init(stringLiteral: String) {
+        self.init(status: stringLiteral)
+    }
+}
+
 public extension CommentType {
     init(type: String) {
         self = commentTypeFromString(value: type)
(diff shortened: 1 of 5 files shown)


native/swift/Sources/wordpress-api/Extensions.swift
```diff
@@ -32,3 +32,23 @@ extension MiddlewarePipeline {
         self.init(middlewares: middlewares)
     }
 }
+
+public extension CommentStatus {
+    init(status: String) {
+        self = commentStatusFromString(value: status)
+    }
+
+    func status() -> String {
```
jkmassel: Should this be a method or a property with a `get`?

I kinda think it should be the latter?

native/swift/Sources/wordpress-api/Extensions.swift
```diff
@@ -32,3 +32,23 @@ extension MiddlewarePipeline {
         self.init(middlewares: middlewares)
     }
 }
+
+public extension CommentStatus {
```
jkmassel, line 36: Should this conform to `ExpressibleByStringLiteral`? I wonder if that'd help with passing it around and/or instantiating new ones (for instance, for a filter)?

native/swift/Sources/wordpress-api/Extensions.swift
crazytonyli: Sounds good. Updated in https://github.com/Automattic/wordpress-rs/pull/734/commits/818599727353f51afe51a6f2824b50a002c90ec5

native/swift/Sources/wordpress-api/Extensions.swift
crazytonyli: Added in https://github.com/Automattic/wordpress-rs/pull/734/commits/7147ff98e7a52e7e2902b31d8e27cd813bff2080.

jkmassel: This gives us `init(stringLiteral: String)` and `init(status: String)`

Do we need both? It seems to me this is an enum type represented by a String, which I _think_ is basically the exact use case for `ExpressibleByStringLiteral `?

WDYT?

crazytonyli: I think they are for different use cases. `ExpressibleByStringLiteral` allows you to `let status: CommentStatus = "hello"`. But if we need to convert a `status: String` to a `status: CommentStatus`, then `CommentStatus(status: status)` is much clearer than `CommentStatus(stringLiteral: status)`.

jkmassel: WDYT about just `CommentStatus(status)` for the extension initializer (like you'd do `String(someInt)`)?

crazytonyli: Do you mean removing the labels like https://github.com/Automattic/wordpress-rs/pull/734/commits/42785042b17721c6d3bfaf5fa1c81dbeb68417c1?

jkmassel: Yeah, I think that's a decent amount nicer. Sorry for all the nitpicks 😅

jkmassel (approved):

PR #735, opened 2025-05-15
Add WP.com Subscribers Endpoints

jkmassel:
Adds support for WP.com subscribers. These endpoints were tested using the Swift implementation.

Some time after this lands, I'd like to spin up another Rust CLI tool that'll run a WP.com e2e test suite against the real endpoints.


wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -18,6 +18,24 @@ pub mod wp_utc_date_format {
         D: Deserializer<'de>,
     {
         let s = String::deserialize(deserializer)?;
+
```
jkmassel, line 21: These changes warrant some additional scrutiny – I was on the fence about whether to implement a new object for alternative time representations, but the other types I came across are trivially expressible as a UTC datetime so to me it made sense to allow a larger variety of values to be parsed. I added a test for each, but there might be more cases I haven't though of?

native/swift/Sources/wordpress-api/WPComApiClient.swift
```diff
@@ -17,4 +17,8 @@ public class WPComApiClient {
     public var jetpackConnection: JetpackConnectionRequestExecutor {
         internalClient.jetpackConnection()
     }
+
+    public var subscriptions: SubscribersRequestExecutor {
```
crazytonyli: Should this property be `subscribers`?

wp_api/src/wp_com/mod.rs
```diff
@@ -6,7 +6,7 @@ pub mod client;
 pub mod endpoint;
 pub mod jetpack_connection;
 pub mod oauth2;
-
+pub mod subscribers;
```
crazytonyli, line 9: Nitpick: add an empty line below?

wp_api/src/wp_com/subscribers.rs
```diff
@@ -0,0 +1,518 @@
...
+    date::WpGmtDateTime,
+    impl_as_query_value_for_new_type,
+    url_query::{AppendUrlQueryPairs, QueryPairs},
+};
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, ops::Not};
+use wp_serde_helper::deserialize_u64_or_string;
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct Subscriber {
+    pub user_id: u64,
```
crazytonyli: Use `UserId`, instead of `u64`?

wp_api/src/wp_com/subscribers.rs
```diff
@@ -0,0 +1,518 @@
...
+pub enum SubscriberType {
+    All,
+    WpCom,
+    Email,
+    Paid,
+    Free,
+    EmailSubscriber,
+    ReaderSubscriber,
+    UnconfirmedSubscriber,
+    BlockedSubscriber,
```
crazytonyli, line 44: I think we should add a `Other(String)` variant here. Otherwise, the parsing would fail if a new type is introduced in the backend.

wp_api/src/wp_com/subscribers.rs
```diff
@@ -0,0 +1,518 @@
...
+            SubscriberType::Paid => "paid".to_string(),
+            SubscriberType::Free => "free".to_string(),
+            SubscriberType::EmailSubscriber => "email_subscriber".to_string(),
+            SubscriberType::ReaderSubscriber => "reader_subscriber".to_string(),
+            SubscriberType::UnconfirmedSubscriber => "unconfirmed_subscriber".to_string(),
+            SubscriberType::BlockedSubscriber => "blocked_subscriber".to_string(),
+        };
+        write!(f, "{}", string)
+    }
+}
```
crazytonyli: You can use `strum` to get this implementation for free (I can't remember exactly, maybe not the `Display` trait, but you can get `to_string` for free).

wp_api/src/wp_com/subscribers.rs
jkmassel: Ah good call – added in https://github.com/Automattic/wordpress-rs/pull/735/commits/94b8f66cc744d2df7c232da9631af193f6ab048a

wp_api/src/wp_com/subscribers.rs
jkmassel: Looks like you can get both – I did that in https://github.com/Automattic/wordpress-rs/pull/735/commits/94b8f66cc744d2df7c232da9631af193f6ab048a

Thanks for the suggestion!

wp_api/src/wp_com/subscribers.rs
jkmassel: I added this in e00cb652585483e37dc95adddf8e2fced60472df – it's slightly more semantic, so it seems nice – just curious – any other reason besides it?

wp_api/src/wp_com/mod.rs
jkmassel: Good nit! f61621e0060902898c9f4851f007e06c263827a9

native/swift/Sources/wordpress-api/WPComApiClient.swift
jkmassel: Ah yep 🤦 

ddae359ffd05b2d0351be58b11fe4aa437ba897b

wp_api/src/wp_com/subscribers.rs
crazytonyli: The main one to me is that the compiler makes sure we can't pass `subscriber.user_id` as other kinds of id (also integer) like: `fn post_comment(post_id: PostId, comment: Comment)`. However, the Swift binding erases that type information, and `UserId` is just an alias to `UInt64`, so we can't actually get that compiler check on Swift.

wp_api/src/wp_com/subscribers.rs
```diff
@@ -0,0 +1,518 @@
...
+    Pending,
+    Unsubscribed,
+    Spam,
+    #[serde(untagged)]
+    #[strum(default)]
+    Custom(String),
+}
+
+#[derive(Debug, Default, PartialEq, Eq, uniffi::Record)]
+pub struct ListSubscribersParams {
```
crazytonyli: According to names of other similar types (`UserListParams`, `MediaUpdateParams`, etc.), this should be named as `SubscribersListParams`,

wp_api/src/wp_com/subscribers.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub page: u64,
+    pub per_page: u64,
+    pub subscribers: Vec<Subscriber>,
+}
+
+// MARK: - Get Subscriber
+
+#[derive(Debug, uniffi::Enum)]
+pub enum GetSubscriberQuery {
+    WpCom(u64),
```
crazytonyli, line 162: Should this be named as `WpComUser`? Or maybe just `User`? The "WP.com" part seems to be implied.

wp_api/src/wp_com/subscribers.rs
```diff
@@ -0,0 +1,518 @@
...
+    #[serde(rename = "imported")]
+    Imported,
+    // Job started.
+    #[serde(rename = "importing")]
+    Importing,
+    // Import failed.
+    #[serde(rename = "failed")]
+    Failed,
+    // Job cancelled.
+    #[serde(rename = "cancelled")]
```
crazytonyli: The individual `serde` attributes can be replaced with one `#[serde(rename_all = "snake_case")]`.

crazytonyli: `SubscriberImportJobStatus` probably needs a `Custom(String)` variant too.

wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -53,4 +71,41 @@ mod tests {
         assert_eq!(foo.wp_utc_date_time.second(), 19);
         assert_eq!(foo.wp_utc_date_time.hour(), 22);
     }
+
+    #[test]
+    fn test_deserialize_iso8601_date() {
+        let json_str = r#"
+          {
+            "wp_utc_date_time": "2023-05-12T14:30:45+00:00"
```
crazytonyli: What do you think about turning this into a test argument and adding a few cases with the exact same time?

```diff
diff --git a/wp_serde_helper/src/wp_serde_date.rs b/wp_serde_helper/src/wp_serde_date.rs
index b93e043..2fda63e 100644
--- a/wp_serde_helper/src/wp_serde_date.rs
+++ b/wp_serde_helper/src/wp_serde_date.rs
@@ -44,6 +44,7 @@ pub mod wp_utc_date_format {
 
 #[cfg(test)]
 mod tests {
+    use rstest::rstest;
     use super::*;
     use chrono::{DateTime, Datelike, Timelike, Utc};
     use serde::{Deserialize, Serialize};
@@ -72,15 +73,12 @@ mod tests {
         assert_eq!(foo.wp_utc_date_time.hour(), 22);
     }
 
-    #[test]
-    fn test_deserialize_iso8601_date() {
-        let json_str = r#"
-          {
-            "wp_utc_date_time": "2023-05-12T14:30:45+00:00"
-          }
-        "#;
-
-        let foo: Foo = serde_json::from_str(json_str).unwrap();
+    #[rstest]
+    #[case("2023-05-12T14:30:45+00:00")]
+    #[case("2023-05-12T15:30:45+01:00")]
+    fn test_deserialize_iso8601_date(#[case] date_string: &str) {
+        let json_str = format!("{{\"wp_utc_date_time\": \"{}\"}}", date_string);
+        let foo: Foo = serde_json::from_str(&json_str).unwrap();
 
         assert_eq!(foo.wp_utc_date_time.year_ce(), (true, 2023));
         assert_eq!(foo.wp_utc_date_time.month(), 5);
```

wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -18,6 +18,24 @@ pub mod wp_utc_date_format {
...
+        // Try to parse as Unix timestamp
+        if let Ok(timestamp) = s.parse::<i64>() {
+            if let Some(dt) = DateTime::<Utc>::from_timestamp(timestamp, 0) {
+                return Ok(dt);
+            }
+        }
+
         NaiveDateTime::parse_from_str(&s, WP_DATE_FORMAT)
             .map_err(serde::de::Error::custom)
             .map(|dt| DateTime::<Utc>::from_naive_utc_and_offset(dt, Utc))
```
crazytonyli: This parsing is duplicated with the first one. We can potentially change the implementation to be something like this:


```rust
    pub fn deserialize<'de, D>(deserializer: D) -> Result<DateTime<Utc>, D::Error>
    where
        D: Deserializer<'de>,
    {
        let s = String::deserialize(deserializer)?;

        // First try to parse as WordPress format
        NaiveDateTime::parse_from_str(&s, WP_DATE_FORMAT)
            .map(|dt| DateTime::<Utc>::from_naive_utc_and_offset(dt, Utc))
            .or_else(|err| {
                // If WordPress format fails, try ISO-8601 format
                DateTime::parse_from_rfc3339(&s)
                    .map(|dt| dt.with_timezone(&Utc))
                    .map_err(|_| err)
            })
            .or_else(|err| {
                // Try to parse as Unix timestamp
                s.parse::<i64>()
                    .ok()
                    .and_then(|timestamp| DateTime::<Utc>::from_timestamp(timestamp, 0))
                    .ok_or(err)
            })
            .map_err(serde::de::Error::custom)
    }
```

wp_api/src/wp_com/subscribers.rs
jkmassel: Addressed this (and some other cases) in 4fe83fc59b2bba6b4e64d40ecc86e72aeeee58da

wp_api/src/wp_com/subscribers.rs
jkmassel: No, in this case `WpCom` refers to the mechanism for notification, nothing to do with users.

I added some comments in 06fc404e96f2eb95254639520060fe078599ad96 to clarify

wp_api/src/wp_com/subscribers.rs
jkmassel: Did both of these in 29016e5dbd6ac2e3284cf3ddf3b586315a84ccca

wp_serde_helper/src/wp_serde_date.rs
jkmassel: This is an ok start – #737 has some even wilder real-world cases that don't work with this format, but I'll figure something out for that!

Adjusted in 351d76990c536043deca2d33a6e0aa233e07885b

wp_serde_helper/src/wp_serde_date.rs
jkmassel: Yeah, I noticed this in #737 – I just backported my approach from there in 351d76990c536043deca2d33a6e0aa233e07885b

crazytonyli (approved):

PR #736, opened 2025-05-15
Rename CommentStatus::Approve to CommentStatus::Approved

crazytonyli:
The "approved" comment status string is `approved`. See this comment on the test site:

```
curl -su 'test@example.com:KYMx5TAaDyoc5sF6I0ONxCFJ' http://localhost/wp-json/wp/v2/comments/1
```


jkmassel (approved):
This seems like a sensible change – it makes it an adjective (describing the comment) instead of a verb (an action on a comment)

PR #737, opened 2025-05-16
Add Support Bots Endpoint

jkmassel:
Adds the support bots endpoints for integration into the Jetpack apps.


wp_api/src/wp_com/endpoint/support_bots.rs
oguzkocer, line 1: **Nitpick:** In `wp_api`, all endpoint files end with `_endpoint` suffix. I prefer that over the naming from this PR because it makes it easier to fuzzy find in the editor.

I don't feel strongly about which naming scheme we use, but I don't like when some files use one scheme and others use a different one, so at least for this PR, I suggest naming this file as `wp_api/src/wp_com/endpoint/support_bots_endpoint.rs`.

If you feel strongly about not having `_endpoint` suffix, please feel free to open a PR to change all existing ones and I'd be happy to approve that.

wp_api/src/wp_com/endpoint/support_bots.rs
```diff
@@ -0,0 +1,88 @@
...
+    },
+};
+use serde::{Deserialize, Serialize};
+use wp_derive_request_builder::WpDerivedRequest;
+
+#[derive(WpDerivedRequest)]
+enum SupportBotsRequest {
+    #[post(url = "/odie/chat/<bot_id>", params = &CreateBotConversationParams, output = BotConversation)]
+    CreateBotConversation,
+
```
oguzkocer: **Nitpick:** All existing files omit these spaces between `enum` variants. Similar to previous suggestion, I'd like us to have a consistent style. I don't really care much about what that style is.

wp_api/src/wp_com/endpoint/support_bots.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        s.parse().map(Self)
+    }
+}
+
+impl std::fmt::Display for MessageId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
```
oguzkocer: **Nitpick:** In `wp_api` these types are always implemented in their specific modules. Similar to the previous 2 suggestions, I'd like us to have a consistent implementation.

wp_api/src/wp_com/support_bots.rs
```diff
@@ -0,0 +1,187 @@
...
+use serde::{Deserialize, Serialize};
+use serde_repr::*;
+use std::collections::HashMap;
+
+use super::WpComSiteId;
+
+#[derive(Debug, PartialEq, Eq, Serialize, uniffi::Record)]
+pub struct CreateBotConversationParams {
+    pub message: String,
+    pub user_id: UserId,
```
oguzkocer, line 16: ❓ I wonder if we should have a `WpComUserId` type or if it's OK to keep using the same `UserId` type for both `wp_api` and `wp_com`.

wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -17,15 +18,35 @@ pub mod wp_utc_date_format {
...
             return Ok(DateTime::<Utc>::from_naive_utc_and_offset(dt, Utc));
         }
 
         // If WordPress format fails, try ISO-8601 format
-        if let Ok(dt) = DateTime::parse_from_rfc3339(&s) {
+        if let Ok(dt) = DateTime::parse_from_rfc3339(s) {
+            return Ok(dt.with_timezone(&Utc));
+        }
+
+        // First, try parsing the WP REST API format
```
oguzkocer: **Nitpick:** We have a bunch of "First do this.." comments in this function. I think we can simplify all of them and keep them similar to `// Try ISO-8601 format` comment.

oguzkocer (approved):
Looks great! I only have a few styling nitpicks/suggestions and a question, but happy to merge this as is.

wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -17,15 +18,35 @@ pub mod wp_utc_date_format {
     where
         D: Deserializer<'de>,
     {
-        let s = String::deserialize(deserializer)?;
+        let value = serde_json::Value::deserialize(deserializer)?;
```
crazytonyli: I don't think `serde_json` should be used here? The original function call seems to be more correct?

oguzkocer: I assume you mean it can be deserialized as `u64` first and if it errors then we try deserializing it as `String`? That's probably a little better, but unless there is a big performance issue, I think it's fine to parse it into `serde_json::Value` first and then see if it can be cast to other types as it's implemented in this PR. 🤷‍♂️

crazytonyli: Sorry, I was not clear. `deserializer` is not always serde_json's Deserializer. Hypothetically speaking, if `WpGmtDateTime` is serialised into other forms (binary, XML, whatever), this code may break, because the data it tries to deserialize may not be in JSON string format.

oguzkocer: Ah.. right. 🤦

jkmassel: Addressed in d776c9c2e34334e8a383f94bcc8b718a0d788f63 – thanks for the suggestion @crazytonyli!

wp_serde_helper/src/wp_serde_date.rs
jkmassel: Addressed in d776c9c2e34334e8a383f94bcc8b718a0d788f63

wp_api/src/wp_com/support_bots.rs
jkmassel: They're both a u64 under the hood, so I don't see any benefit to having separate types (for the moment).

It also wouldn't be a breaking change (at least for Rust / Swift) to adjust it later if we find out we missed something

wp_api/src/wp_com/endpoint/support_bots.rs
jkmassel: Renamed in c1b742f3ceb5278d9d80b2650d67f05071507bd5

wp_api/src/wp_com/endpoint/support_bots.rs
jkmassel: I find the spaces way more readable, but I'm fine to align it – adjusted in c1b742f3ceb5278d9d80b2650d67f05071507bd5

wp_api/src/wp_com/endpoint/support_bots.rs
jkmassel: Consolidated in c1b742f3ceb5278d9d80b2650d67f05071507bd5

crazytonyli (approved):

PR #743, opened 2025-05-23
Add Support Tickets + Eligibility endpoints

jkmassel:
Adds WP.com support tickets (ie – Zendesk) and support eligibility endpoints. 

Support tickets require the ability to upload files – this PR doesn't do that because we'll need to generalize our file upload stuff. I don't think that's a blocker to this PR landing – the majority of the functionality is there.

I've also added a `wp_com_e2e` package that'll validate that the support tickets / eligibility endpoints function as expected. You can run it like `cargo run --bin wp_com_e2e test --token [my_token]`. It _must_ be a WordPress iOS token at the moment – we can consider adjusting that later though.

If you run that test (or trust that I did), I think we can consider this PR working. There are also unit tests for parsing the responses, so I think we're pretty well-covered there.


wp_api/src/wp_com/support_tickets.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub updated_at: WpGmtDateTime,
+    pub messages: Vec<SupportMessage>,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, uniffi::Record)]
+pub struct SupportMessage {
+    pub id: u64,
+    pub content: String,
+    pub author: SupportMessageAuthor,
+    pub role: String,
```
crazytonyli, line 47: At the moment how the `author` is parsed id depending on the `SupportMessageAuthor` enum variant order: it's a "User" if the JSON can be parsed as an "User"; then it's a "Agent" if the JSON can be parsed as an "Agent". Would it be more correct to parse `author` based on the `role` value, instead of how the `SupportMessageAuthor` enum type is declared?

wp_api/src/wp_com/support_tickets.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub size: u64,
+    pub url: String,
+    pub metadata: HashMap<String, AttachmentMetadataValue>,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, uniffi::Enum)]
+#[serde(untagged)]
+pub enum SupportMessageAuthor {
+    User(SupportUserIdentity),
+    SupportAgent(SupportAgentIdentity),
```
crazytonyli, line 75: Should there be an "other" variant, just in case the "role" value is some other value, like "system" messages?

wp_com_e2e/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+            let client = WpComApiClient::new(delegate);
+
+            Oauth2Test {
+                client: &client,
+                token: &token,
+            }
+            .test()
+            .await?;
+            SupportTicketsTest { client: &client }.test().await?;
+            SupportEligibilityTest { client: &client }.test().await?;
```
crazytonyli: Any reason not to use cargo to run these tests, like the existing integration tests?

crazytonyli (dismissed):

wp_com_e2e/src/main.rs
jkmassel: I'm hoping that at some point this could be run against WP.com on each commit to see if there are breaking changes, which is why I started down this path

oguzkocer: I think we'd still be able to do that with the regular cargo test setup. Is there something specific you're worried about?

oguzkocer (commented):
The implementation looks good overall, but I am not so sure about the e2e testing setup. I've brought up a question in the line comments, let's discuss where to go from there.

jkmassel:
I'm ok with whatever testing approach makes sense – if we can pre-compile a binary for Linux that uses the Rust test stuff, then I'm good with it. One other note, I guess, we'll need to be able to chain some tests together so certain actions run before others (can't be avoided for stuff like support tickets).

wp_api/src/wp_com/support_tickets.rs
oguzkocer: In order to do this, we'd need to parse it as a json value because we don't know the structure of the type.

If we know for sure that there are other types of authors, but we don't know the structure of the type, then I think this makes sense to do. Otherwise, it's probably leave it off and implement it when we need to.

wp_api/src/wp_com/support_tickets.rs
oguzkocer: Although this would be better, it's also more tricky to do. At least, I couldn't think of an immediate solution to this except for either using a custom parser or separating the `SupportMessage` type as `SupportUserMessage` and `SupportAgentMessage`. (which doesn't sound like a bad idea 🤔 )

Since `SupportUserIdentity` and `SupportAgentIdentity` don't have the same fields, I _think_ the way it's implemented is mostly safe. I say mostly, because future changes to these types may make it unsafe as in they might be incorrectly parsed to one type as described in Tony's comment. So, maybe we _should_ separate the message types if we can. (I haven't fully investigated this yet)

What do you think @crazytonyli?

wp_com_e2e/src/main.rs
oguzkocer: I can see _some_ value in making this a binary as it'll make it easier to run it from outside of this project without setting up the toolchain.

I don't have a good grasp of the full picture of what we need for this at the moment, so I don't think it's worth changing what's already implemented. Having said that, I don't think the `Testable` stuff is very useful. So, I simplified it a bit in d08e2061a2c7b5cc74dd7e2e2130cd716fa9e6bf.

wp_api/src/wp_com/support_tickets.rs
crazytonyli: My concern is that the app may fail to parse the messages if the API introduces a new "author" with a different JSON schema than "user" and "agent".

oguzkocer: Let me think about this again depending on your answer to [this](https://github.com/Automattic/wordpress-rs/pull/743#discussion_r2136596892).

My general worry is for a parsing issue to go unnoticed and introducing a fallback inherently has that problem. Ideally I'd like to have a way to not have parsing errors, but still get a notice about the unexpected fallbacks. Once we have logging implemented, it might help with that a bit.

wp_api/src/wp_com/support_tickets.rs
crazytonyli: > or separating the SupportMessage type as SupportUserMessage and SupportAgentMessage. 

This sounds good to me 👍

crazytonyli: I have created https://github.com/Automattic/wordpress-rs/issues/852 to track the potential issue discussed above.

crazytonyli (approved):

PR #745, opened 2025-05-25
List `/posts/<parent>/revisions` endpoint

oguzkocer:
Follows established patterns to implement list [`/posts/<parent>/revisions` endpoint](https://developer.wordpress.org/rest-api/reference/post-revisions/#list-post-revisions).


scripts/setup-test-site.sh
```diff
@@ -125,6 +126,16 @@ create_test_credentials () {
...
 
+  # Setup a post with post revisions for integration tests
+  REVISIONED_POST_ID="$(wp post create --post_type=post --post_title=Revisioned_POST_FOR_INTEGRATION_TESTS --porcelain)"
+  curl --user "$ADMIN_USERNAME":"$ADMIN_PASSWORD" -H "Content-Type: application/json" -d '{"content":"Add some content for first revision"}' "http://localhost/wp-json/wp/v2/posts/$REVISIONED_POST_ID/autosaves"
+  # TODO: The following doesn't create multiple revisions, but instead updates the existing revisions which is not what we need
+  # # Create some exta revisions for testing
+  # for i in {1..10};
+  # do
+  #   curl --user "$ADMIN_USERNAME":"$ADMIN_PASSWORD" -H "Content-Type: application/json" -d "{\"content\":\"content_revision_$i\"}" "http://localhost/wp-json/wp/v2/posts/$REVISIONED_POST_ID/autosaves"
+  # done
```
oguzkocer: @crazytonyli @jkmassel Any chance you know how to create multiple post revisions using the API? I only see the [`/posts/<parent_id>/autosaves` endpoint](https://developer.wordpress.org/rest-api/reference/post-revisions/#create-a-post-revision) for creating a revision in the documentation, but using that only creates 1 revision and overrides it for subsequent calls.

I tried sending a `POST` request to `/posts/<parent_id>/revisions`, but the server seems to interpret that as a `GET` request and returns the existing revisions.

---

I think I understand `/revisions` and `/autosaves` endpoints at a conceptual level, but I can't fully wrap my head around them in the documentation because there doesn't seem to be a POST request for `/revisions`. So, if either of you have extra insight about how they work, I'd appreciate it.

oguzkocer: We chatted a little bit about this with Jeremy on a Slack call and he mentioned that the revisions are most likely created on the server side when a post is updated. I'll give that a shot and update the PR accordingly.

crazytonyli (approved):

PR #750, opened 2025-05-29
Parse the media_details JSON into concrete instances

crazytonyli:
The `media_details` property is declared as `JsonValue` at the moment. This PR changes it to be a concrete `MediaDetails` type. The new type provides a function `fn parse_as_mime_type(mime_type: String)` to return a concrete instance that contains the media metadata.

| Website | Example App |
| ------- | ----------- |
|  |  |


oguzkocer (approved):
I've reviewed this PR before, but I was a little on the fence about approving it due to a few concerns:

1. As far as I can tell, there is currently no way to access the raw JSON from `MediaDetails`.
2. We can't cover all possible fields for any given mime type. If the client only needs the fields from this PR than this will work well, but if they need a specific field, client has to manually parse it.
3. The parsing errors are currently swallowed. That means when we call `parse_as_mime_type`, if we receive `None`, it's not clear why.

After some consideration, I thought this PR is an improvement over what we have and it's useful to the clients as is. So, I think it's OK to merge it and then keep improving it as we work on the integration.

PR #752, opened 2025-05-29
Allow passing WP.com base url

crazytonyli:
This is to support the app's UI tests, by allowing the app to swap the real WP.com public API URL with a mock server URL.

### DIFF
--- a/wp_api/src/wp_com/endpoint.rs
+++ b/wp_api/src/wp_com/endpoint.rs
@@ -14,16 +14,18 @@ pub mod support_bots_endpoint;
 #[derive(uniffi::Object)]
 pub struct WpComDotOrgApiUrlResolver {
     pub base_url: ParsedUrl,
-    pub site_url: String,
+    pub site_id: String,
 }
 
 #[uniffi::export]
 impl WpComDotOrgApiUrlResolver {
     #[uniffi::constructor]
-    pub fn new(site_url: String) -> Self {
+    pub fn new(base_url: Option<String>, site_id: String) -> Self {
         Self {
-            base_url: wp_com_base_url(),
-            site_url,
+            base_url: base_url
+                .and_then(|url| ParsedUrl::parse(&url).ok())
+                .unwrap_or_else(wp_com_base_url),
+            site_id,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


wp_api/src/wp_com/endpoint.rs
```diff
@@ -14,16 +14,18 @@ pub mod support_bots_endpoint;
 #[derive(uniffi::Object)]
 pub struct WpComDotOrgApiUrlResolver {
     pub base_url: ParsedUrl,
-    pub site_url: String,
+    pub site_id: String,
```
crazytonyli, line 17: Both site id and site url are acceptable. I changed this to be `site_id` to be consistent with the `$site_id` in most API docs.

wp_api/src/wp_com/endpoint.rs
```diff
@@ -14,16 +14,18 @@ pub mod support_bots_endpoint;
...
 }
 
 #[uniffi::export]
 impl WpComDotOrgApiUrlResolver {
     #[uniffi::constructor]
-    pub fn new(site_url: String) -> Self {
+    pub fn new(base_url: Option<String>, site_id: String) -> Self {
```
oguzkocer: I think using an `enum` such as the one below would be a nicer solution. The function signature with `Option<String>` doesn't make it clear what happens when `None` is passed as an argument without checking the implementation. Using an `enum` would allow us to communicate the information with type signature rather than documentation.

Having said that, the benefit is quite minimal, so I am fine if you want to go with the current implementation.

```
#[derive(Debug, Default, PartialEq, Eq, uniffi::Enum)]
pub enum WpComBaseUrl {
    #[default]
    Default,
    Custom(String),
}
```

oguzkocer (approved):

wp_api/src/wp_com/endpoint.rs
crazytonyli: Good suggestion. I have implemented it in https://github.com/Automattic/wordpress-rs/pull/752/commits/fb4e0c59a0e35c57c0be9184d3ab77b10b8310fa.

oguzkocer: Thanks for the update!

PR #765, opened 2025-06-14
[Tooling] Add localization automation

iangmaia:
Related: [AINFRA-765: Integrate new Fluent conversion tool into wordpress-rs](https://linear.app/a8c/issue/AINFRA-765/integrate-new-fluent-conversion-tool-into-wordpress-rs)

**Note**: The `fluent-tools` Gem used in the PR is currently using the Gem directly from the branch [Add Ruby Gem wrapper / CLI interface and CI setup](https://github.com/Automattic/fluent-rust-tools/pull/3). 🎗️ We need to update it once the Rust tool is fully merged.

This PR supersedes https://github.com/Automattic/wordpress-rs/pull/756.

## What does it do?
This PR integrates the new [Fluent conversion tool](https://github.com/Automattic/fluent-rust-tools) Gem into `wordpress-rs` to enable a complete translation workflow between the project's Fluent localization files (.ftl) and GlotPress, allowing for:
- **Source generation**: Converts the source English Fluent file to PO format (.pot) for uploading to GlotPress
- **Translation sync**: Downloads translated PO files from GlotPress and converts them back to Fluent format, adding them to the project
- **Locale management**: Supports MAG16 locales with mapping between GlotPress and project locale codes; this was implemented just to limit the languages we know we support, but this mapping can be easily updated in the future

## New lanes
- `generate_source_po_file` lane - Converts `en-US/main.ftl` to `en-US.pot` for GlotPress upload
- `download_translations` lane - Downloads all available translations from GlotPress following the supported locales list and updates local Fluent files
- `generate_fluent_file_from_po` lane - Converts individual PO files back to Fluent format
- `download_po_files_from_glotpress` lane - Downloads PO files for all supported locales

## Proposed localization workflow
A day-to-day workflow can look like this:

1. New features are implemented, new localization strings need to be created (or updated).
2. Developers add strings to the source language file `wp_localization/localization/en-US/main.ftl`.
3. With a certain frequency or based on a trigger (for example, new commits to `trunk` or a daily scheduled build), we run `generate_source_po_file` and commit the updated `wp_localization/glotpress/en-US.pot`.
4. A `wpcom` job uploads the updated `en-US` source file to GlotPress (see 184627-ghe-Automattic/wpcom )
5. Again with a given frequency, we fetch the translations, generate the Fluent files and commit them to the project (or create a PR for review) by running `download_translations`.

[...]

PR #770, opened 2025-06-16
[Tooling] Add Localization sync job

iangmaia:
Related to [AINFRA-573: Integrate Fluent translations import/export in the Release Process](https://linear.app/a8c/issue/AINFRA-573/integrate-fluent-translations-importexport-in-the-release-process)

Based on https://github.com/Automattic/wordpress-rs/pull/765, which needs to be merged first.

This PR adds a job to sync the translations, **reusing the nightly schedule**. It essentially:
1. Generates a new source English `.po` file based on `wp_localization/localization/en-US/main.ftl` and updates it in the repo (so it is uploaded to GlotPress with a wpcom job)
2. Downloads the translations from GlotPress and converts them to Fluent files, updating them in the repo.


.buildkite/commands/sync-localization.sh
```diff
@@ -0,0 +1,13 @@
...
+source use-bot-for-git
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :globe_with_meridians: :arrow_up: Generate the source language PO file for GlotPress based on `wp_localization/localization/en-US/main.ftl`"
+bundle exec fastlane generate_source_po_file commit_and_push_changes:true
+
+echo "--- :globe_with_meridians: :arrow_down: Download and update translations from GlotPress and update the local Fluent files"
+bundle exec fastlane download_translations commit_and_push_changes:true
```
AliSoftware: Not critical, but maybe it'd be nicer to have the lanes only `commit` without pushing, and push only at the end after both lanes have run, to push both commits at once instead of having 2 push operations? 🤷

AliSoftware (approved):

.buildkite/commands/sync-localization.sh
iangmaia: Yeah, I also realized it once I published this PR 👍  I wasn't sure how the lanes were going to work independently but having them running together makes it clearer that is better to avoid the side effect in the lane itself.

I'll write it down to tackle it in https://github.com/Automattic/wordpress-rs/pull/765 and then update it here 👍

iangmaia: Updated in 2c4547caa946d6d19c5e2a89dc7cd6e1d7fcaac2 / e022478ba7688903d662737b01c8bd692c835c6c.

PR #790, opened 2025-07-03
[Tooling] Update localization sync to create a PR

iangmaia:
Follow-up to https://github.com/Automattic/wordpress-rs/pull/770
Related to [AINFRA-573](https://linear.app/a8c/issue/AINFRA-573)

This PR changes the localization sync job to create a Pull Request when there are localization changes, instead of committing directly to `trunk` (which would fail anyway as the branch is protected and even admins are not allowed to bypass).

### Testing

I've run a couple of builds manually (with `PIPELINE=nightly.yml`) to test this PR:
- Build without changes: https://buildkite.com/automattic/wordpress-rs/builds/3194#0197d000-facf-4251-9a98-272b7c070191/312-332
- Build with one translation update: https://buildkite.com/automattic/wordpress-rs/builds/3196#0197d00b-dfef-4bef-9081-4c6b8b99e2aa/742-746 (resulting in [this PR](https://github.com/Automattic/wordpress-rs/pull/791/files#diff-444b00a617b054b7a9f666725893bcccbd84c09ffe5fa47332a26a12c56ec018R21), which contains the changes from this branch as I ran it in Buildkite directly from this branch)

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -13,6 +13,7 @@ LANE_VALUE_XCFRAMEWORK_CHECKSUM_PATH = 'WP_XCFRAMEWORK_CHECKSUM_PATH'
 
 GITHUB_REPO = 'automattic/wordpress-rs'
 GIT_REMOTE_NAME = 'origin'
+GIT_BASE_BRANCH = 'trunk'
 
 # Localization constants
 LOCALIZATION_FLUENT_FILES_DIR = File.join(PROJECT_ROOT, 'wp_localization', 'localization')
@@ -179,8 +180,9 @@ end
 # The resulting PO file is saved as the source file (.pot) for translations and is synced to GlotPress.
 #
 # @param commit_changes [Boolean] Whether to commit the generated PO file (default: false)
+# @param skip_date_only_changes [Boolean] Whether to skip commit if only date headers changed (default: true)
 #
-lane :generate_source_po_file do |commit_changes: false|
+lane :generate_source_po_file do |commit_changes: false, skip_date_only_changes: true|
   UI.header('🔄 Converting English Fluent file to PO format')
 
   FileUtils.mkdir_p(File.dirname(LOCALIZATION_PO_SOURCE_FILE))
@@ -202,6 +204,12 @@ lane :generate_source_po_file do |commit_changes: false|
   end
 
   if commit_changes
+    # Check if we should skip commit when only date headers changed
+    if skip_date_only_changes && only_date_headers_changed?(LOCALIZATION_PO_SOURCE_FILE)
+      UI.message('ℹ️ Skipping commit - only date headers changed in PO file')
+      next
+    end
+
     commit_changed_files(
       files: LOCALIZATION_PO_SOURCE_FILE,
       message: 'Update source PO file (en-US.pot) to be synced to GlotPress'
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -248,6 +256,56 @@ lane :download_translations do |commit_changes: false|
...
+
+      🔄 This is an automated localization sync.
+    BODY
+    labels: 'Localization',
+    base: GIT_BASE_BRANCH,
+    head: sync_branch,
+    api_token: github_token
+  )
+
+  UI.success("✅ Created pull request: #{pr_result}")
```
AliSoftware: Nit: do we want to `buildkite_annotate` the PR link as well while at it? 🤷

AliSoftware (approved):

fastlane/Fastfile
```diff
@@ -248,6 +256,56 @@ lane :download_translations do |commit_changes: false|
...
+    title: 'Sync localization files with GlotPress',
+    body: <<~BODY,
+      This PR synchronizes localization files with GlotPress:
+
+      - **Source PO file**: Updated from English Fluent file for GlotPress sync
+      - **Translations**: Downloaded latest translations from GlotPress and converted to Fluent format
+
+      🔄 This is an automated localization sync.
+    BODY
+    labels: 'Localization',
```
AliSoftware, line 309: Should we also add `reviewers:` or `team_reviewers:` to make sure the PR is not missed and the team have notifications to remember to review it?

iangmaia: Good idea -- added on 07de509050fb35837899b1cdfa37a8fca5713a1c.

fastlane/Fastfile
iangmaia: It felt redundant as this isn't a build part of a release (as in: there aren't many reasons to go to this build)... but at the same time, it is handy to show it. Added on 03cb0b3bf807ee7b71ecc656254017d454708d0d.

fastlane/Fastfile
```diff
@@ -248,6 +256,74 @@ lane :download_translations do |commit_changes: false|
...
+  translations_result = download_translations(commit_changes: true)
+
+  # Check if either operation resulted in changes
+  has_changes = !po_result.nil? || !translations_result.nil?
+  unless has_changes
+    UI.important('⚠️ No changes to commit')
+
+    if is_ci
+      buildkite_annotate(
+        context: 'localization-sync-no-changes',
```
AliSoftware, line 284: I think you might as well use the same context for both cases since they're mutually exclusive?

```suggestion
        context: 'localization-sync',
```

fastlane/Fastfile
```diff
@@ -248,6 +256,74 @@ lane :download_translations do |commit_changes: false|
...
+    base: GIT_BASE_BRANCH,
+    head: sync_branch,
+    api_token: github_token
+  )
+
+  UI.success("✅ Created pull request: #{pr_result}") if pr_result
+
+  if is_ci && pr_result
+    buildkite_annotate(
+      context: 'localization-sync-pr',
```
AliSoftware, line 320:
```suggestion
      context: 'localization-sync',
```

fastlane/Fastfile
```diff
@@ -248,6 +256,74 @@ lane :download_translations do |commit_changes: false|
...
+  po_result = generate_source_po_file(commit_changes: true, skip_date_only_changes: true)
+  translations_result = download_translations(commit_changes: true)
+
+  # Check if either operation resulted in changes
+  has_changes = !po_result.nil? || !translations_result.nil?
+  unless has_changes
+    UI.important('⚠️ No changes to commit')
+
+    if is_ci
+      buildkite_annotate(
```
AliSoftware, line 283: Not for this PR, but we should probably include that `if is_ci` check as part of the `def self.run` implementation of our `buildkite_*` actions from `release-toolkit` at one point…

PR #794, opened 2025-07-04
Cancelling uploadMedia task should cancel underlying HTTP request

crazytonyli:
There are two main changes in this PR:
1. Add `RequestExecutionErrorReason::CancellationError`. Currently, cancelling an HTTP API call results in `RequestExecutionErrorReason::GenericError { message: "cancelled" }`, which is not specific enough. Adding a `CancellationError` would help developers to distinguish errors due to the user cancelling operations from other unexpected errors.
2. I initially thought the `try await uploadTask.value` call would handle cancellation, but turns out [it does not](https://forums.swift.org/t/understanding-task-cancellation/75329). I have addressed this issue in this PR and added a couple of unit tests to cover it.


oguzkocer (approved):
I _think_ we'll have to update the Kotlin wrapper to handle the new error type. I am surprised it didn't cause a failure. However, we can merge this as is and I can take a look at that separately.

oguzkocer:
> I think we'll have to update the Kotlin wrapper to handle the new error type.

We are not mapping `RequestExecutionErrorReason` types in Kotlin, so this was incorrect.

PR #807, opened 2025-07-16
Add an 'extras' property to comments

crazytonyli:
@oguzkocer I tried the "serde flatten" solution you mentioned in our 1:1, and it worked out great. One minor downside is that there will be additional parsing work required for all API endpoints. I tried to use `RawValue` to avoid that, but it does not seem to work with `serde(flatten)` (see https://github.com/serde-rs/json/pull/1252).

WP.com endpoints may return extra information in the responses. We'll need to parse that information to support the app's functionalities. This PR introduces a mechanism to support that. We can add this block of code to the models if needed:

```rust
    #[serde(flatten)]
    pub extras: Option<JsonValue>,
```

### DIFF
--- a/wp_api/src/comments.rs
+++ b/wp_api/src/comments.rs
@@ -530,6 +530,7 @@ pub struct SparseComment {
     pub author_avatar_urls: Option<HashMap<UserAvatarSize, WpResponseString>>,
     #[serde(flatten)]
     #[WpContext(edit, embed, view)]
+    #[WpContextualExcludeFromFields]
     pub additional_fields: Option<Arc<AnyJson>>,
     // meta field is omitted for now: https://github.com/Automattic/wordpress-rs/issues/422
 }
--- a/wp_api/src/comments.rs
+++ b/wp_api/src/comments.rs
@@ -1,5 +1,5 @@
 use crate::{
-    UserAvatarSize, UserId, WpApiParamOrder, WpResponseString,
+    JsonValue, UserAvatarSize, UserId, WpApiParamOrder, WpResponseString,
     date::WpGmtDateTime,
     impl_as_query_value_for_new_type, impl_as_query_value_from_to_string,
     posts::PostId,
@@ -528,6 +528,9 @@ pub struct SparseComment {
     pub comment_type: Option<CommentType>,
     #[WpContext(edit, embed, view)]
     pub author_avatar_urls: Option<HashMap<UserAvatarSize, WpResponseString>>,
+    #[serde(flatten)]
+    #[WpContext(edit, embed, view)]
+    pub extras: Option<JsonValue>,
     // meta field is omitted for now: https://github.com/Automattic/wordpress-rs/issues/422
 }
 
--- a/wp_api/src/comments.rs
+++ b/wp_api/src/comments.rs
@@ -1,5 +1,5 @@
 use crate::{
-    JsonValue, UserAvatarSize, UserId, WpApiParamOrder, WpResponseString,
+    AnyJson, UserAvatarSize, UserId, WpApiParamOrder, WpResponseString,
     date::WpGmtDateTime,
     impl_as_query_value_for_new_type, impl_as_query_value_from_to_string,
     posts::PostId,
@@ -8,7 +8,7 @@ use crate::{
     },
 };
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, num::ParseIntError, str::FromStr};
+use std::{collections::HashMap, num::ParseIntError, str::FromStr, sync::Arc};
 use strum_macros::IntoStaticStr;
 use wp_contextual::WpContextual;
 
@@ -530,7 +530,7 @@ pub struct SparseComment {
     pub author_avatar_urls: Option<HashMap<UserAvatarSize, WpResponseString>>,
     #[serde(flatten)]
     #[WpContext(edit, embed, view)]
-    pub extras: Option<JsonValue>,
+    pub additional_fields: Option<Arc<AnyJson>>,
     // meta field is omitted for now: https://github.com/Automattic/wordpress-rs/issues/422
 }
 
--- a/wp_api/src/wp_com/endpoint/extensions/comments.rs
+++ b/wp_api/src/wp_com/endpoint/extensions/comments.rs
@@ -25,12 +25,12 @@ pub struct WpComCommentExtensionPostInfo {
 
 #[uniffi::export(with_foreign)]
 pub trait WpComCommentExtensionProvider: Send + Sync {
-    fn parse_extension(&self) -> Result<WpComCommentExtension, UniffiSerializationError>;
+    fn parse_wpcom_comments_extension(&self) -> Result<WpComCommentExtension, UniffiSerializationError>;
 }
 
 #[uniffi::export]
 impl WpComCommentExtensionProvider for AnyJson {
-    fn parse_extension(&self) -> Result<WpComCommentExtension, UniffiSerializationError> {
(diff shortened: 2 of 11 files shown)


oguzkocer (commented):
@crazytonyli I think this is a good start, but it looks like we have quite a few failing tests.

* We could consider marking the field with `#[WpContextualOption]` because there won't always be extra fields. I wonder if that'd also help with de-serializing speed.
* We might want to introduce a new type to hold the extras, instead of directly using `JsonValue`. This might be changed to be a `uniffi::Object` with some helpers in the future and having a type in advance could make things easier for us. It'd also look a bit more consistent when we use it in other types. Note that you probably need to use `#[serde(transparent)]` for this to work, but I don't know if `#[serde(transparent)]` & `#[serde(flatten)]` works together or not.
* The field ordering _might_ matter, as I believe it does for de-serializing `enum`s. That'd be a bit of an issue, because if we accidentally put this field above others, the ones below might always be set to `None`. If this is an issue, we could add a check in `WpContext` proc macro to ensure that if there is an `extra` field, it's at the end. Having a specific type would also help with this.
* If we have a separate type, we could consider moving the `meta` field to it. I am not so sure about this though, so just dropping it as an idea for now.

crazytonyli:
> The field ordering might matter

It does not seem so: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b62029454b6f50395fbb3dfd17a8d4b4.

BTW, I'll leave this PR open until we have an API endpoint that requires this change, like the WP.com /wp/v2/comments (see https://linear.app/a8c/issue/AINFRA-939).

crazytonyli:
@oguzkocer This PR is ready to be reviewed and merged. The `WpComCommentExtension` struct contains data that only exists in the WP.com API.

wp_api/src/comments.rs
```diff
@@ -528,6 +528,9 @@ pub struct SparseComment {
     pub comment_type: Option<CommentType>,
     #[WpContext(edit, embed, view)]
     pub author_avatar_urls: Option<HashMap<UserAvatarSize, WpResponseString>>,
+    #[serde(flatten)]
+    #[WpContext(edit, embed, view)]
+    pub additional_fields: Option<Arc<AnyJson>>,
```
crazytonyli, line 534: This field is included in the generated `SparseCommentFields`, which it should not be. I guess we'll have to implement a logic to exclude certain fields. We can either check the `#[serde(flatten)]` attribute (if that's possible), or add a new attribute like `#[WpContextualExcludeFromFields]`. What do you think? @oguzkocer

wp_api/src/wp_com/endpoint/extensions/comments.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub id: u64,
+    pub title: String,
+    #[serde(rename = "type")]
+    pub kind: String,
+    pub link: String,
+}
+
+#[uniffi::export(with_foreign)]
+pub trait WpComCommentExtensionProvider: Send + Sync {
+    fn parse_extension(&self) -> Result<WpComCommentExtension, UniffiSerializationError>;
```
oguzkocer: We probably need to rename this to `parse_wp_com_comment_extension` as otherwise it'll create a conflict if `parse_extension` is reused for another similar trait.

wp_api_integration_tests/tests/test_comments_immut.rs
```diff
@@ -207,6 +209,73 @@ async fn list_comments_with_edit_context_parse_author_avatar_urls(
...
+            assert!(map.contains_key("_links"));
+        }
+        _ => panic!("Expected extras to be an object"),
+    }
+}
+
+#[tokio::test]
+#[parallel]
+#[ignore]
+async fn wpcom_comment_extension() {
```
oguzkocer: We should keep everything related to `wp_com` contained, so that we can move them to `wp_com` crate when the time comes. Can you create a new file for `wp_com` comments tests? See `wp_api_integration_tests/tests/test_wp_com_subscribers_immut.rs` for an example.

wp_api_integration_tests/tests/test_comments_immut.rs
```diff
@@ -207,6 +209,73 @@ async fn list_comments_with_edit_context_parse_author_avatar_urls(
...
+    // You'll need to replace the site ID and comment ID with valid ones.
```
oguzkocer: Consider using `WpComTestCredentials` for this and adding the field to `wp_com_test_credentials.json-example`.

wp_api_integration_tests/tests/test_comments_immut.rs
```diff
@@ -207,6 +209,73 @@ async fn list_comments_with_edit_context_parse_author_avatar_urls(
...
+            auth_provider: Arc::new(WpAuthenticationProvider::static_with_auth(
+                WpAuthentication::Bearer {
+                    token: WpComTestCredentials::instance().bearer_token.to_string(),
+                },
+            )),
+            request_executor: Arc::new(ReqwestRequestExecutor::default()),
+            middleware_pipeline: Arc::new(WpApiMiddlewarePipeline::default()),
+            app_notifier: Arc::new(EmptyAppNotifier),
+        },
+    );
```
oguzkocer: Probably worth adding a helper for this, so we can re-use in other tests. See `wp_api_integration_tests/src/lib.rs` for various helpers. (We need to move the `wp_com` stuff out of that module, but that can be done at a later time)

wp_api_integration_tests/tests/test_comments_immut.rs
```diff
@@ -207,6 +209,73 @@ async fn list_comments_with_edit_context_parse_author_avatar_urls(
...
+        .data;
+    assert!(comment.additional_fields.parse_extension().is_ok());
+
+    let comment = client
+        .comments()
+        .retrieve_with_embed_context(&comment_id, &CommentRetrieveParams::default())
+        .await
+        .assert_response()
+        .data;
+    assert!(comment.additional_fields.parse_extension().is_ok());
```
oguzkocer: It'd be best to have a single call per test, both for efficiency and to follow existing conventions.

wp_api/src/comments.rs
oguzkocer: I was checking what needed to be done for this and figured I'd quickly implement it, so you don't have to do the same: #833

oguzkocer (changes requested):
@crazytonyli This is mostly looking good. I've left some suggestions. Let me know what you think!

wp_api/src/comments.rs
crazytonyli: Thanks!

I have got this error using the new attribute, do you mind having a look?

```
$ cargo build
   Compiling wp_api v0.1.0 (/Users/tonyli/Projects/wordpress-rs/wp_api)
error[E0599]: no variant or associated item named `AdditionalFields` found for enum `SparseCommentFieldWithEditContext` in the current scope
   --> wp_api/src/comments.rs:495:57
    |
495 | #[derive(Debug, Serialize, Deserialize, uniffi::Record, WpContextual)]
    |                                                         ^^^^^^^^^^^^
    |                                                         |
    |                                                         variant or associated item not found in `SparseCommentFieldWithEditContext`
    |                                                         variant or associated item `AdditionalFields` not found for this enum
    |
    = note: this error originates in the derive macro `WpContextual` (in Nightly builds, run with -Z macro-backtrace for more info)
```

oguzkocer: Fixed in c0cbafddf532a00296b120995df46923c098964d. (after checking with @crazytonyli that it'd be OK for me to push directly to his branch)

crazytonyli:
Hey @oguzkocer , I have addressed your comments. Do you mind having another look? Thanks.

oguzkocer (approved):

PR #810, opened 2025-07-21
Kotlin wrapper fix: use File direct access to fix the FileNotFound exception

adalpari:
**Problem**
The Kotlin wrapper was experiencing file access issues during media uploads, throwing MediaUploadRequestExecutionException.MediaFileNotFound exceptions when attempting to upload files. This prevented successful file uploads through the WordPress REST API implementation.

**Solution**
This PR resolves the file access issue by modifying the Kotlin wrapper to use direct File access instead of the previous file handling approach. The change ensures that files can be properly accessed and read during the upload process, eliminating the MediaFileNotFound exception.


oguzkocer (changes requested):
@adalpari It looks like the create media integration test is broken:

> java.lang.AssertionError: Request wasn't successful: MediaFileNotFound(filePath=/test-data/test_media.jpg)

adalpari:
> @adalpari It looks like the create media integration test is broken:
> 
> > java.lang.AssertionError: Request wasn't successful: MediaFileNotFound(filePath=/test-data/test_media.jpg)

Yes, I saw it. And actually, that's the reason why the function was not working, but the test was. The function was getting the file as a resource of the project, which is available ONLY inside the project or in test build. However, in prod build we are actually sending external files paths.

Do you happen to know the absolute path of the file? Maybe the test will work that way.

There are other two alternatives, but since I cannot build the project I am a bit limited:

1. Create a FileResolver and inject it inside `WpRequestExecutor` so I can mock it in test to get return the proper file.
2. Add the `File` as a parameter of `MediaUploadRequest` instead of the path. So the client will decide how to resolve it.

As said I am a bit limited, specially for option 2.

Wdyt?

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -168,3 +167,7 @@ private fun requestExecutionFailedWith(reason: RequestExecutionErrorReason) =
         redirects = null,
         reason = reason
     )
+
+open class FileResolver {
```
oguzkocer: Shouldn't this be an interface?

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -80,6 +81,7 @@ class WpRequestExecutor(
             }
         }
 
+    @Suppress("ComplexCondition")
```
oguzkocer: Is this due to `if (file == null || !file.exists() || !file.isFile || !file.canRead())` condition?

It's fine to suppress it, but I find it weird that it's causing a warning for this 😕

oguzkocer (changes requested):
I made a minor suggestion to use `interface` over `open class`.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
adalpari: Good point!

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
adalpari: Yes it's because of that. I can extract the file checks though.

adalpari:
I've now made the changes you suggested

oguzkocer (approved):

PR #821, opened 2025-07-29
`DynamicAuthenticationProvider` supports refreshing authentication credentials

crazytonyli:
The app has implemented automatically created application passwords, which means we can recover from "unauthorized errors" when the application password becomes invalid.

This PR updates request executors to ask `DynamicAuthenticationProvider` to refresh its credentials and retries the request with the new credentials.

### DIFF
--- a/wp_api/src/auth.rs
+++ b/wp_api/src/auth.rs
@@ -77,6 +77,10 @@ pub trait WpDynamicAuthenticationProvider: Send + Sync + Debug {
 
     /// Refresh the authentication token. The implementation should only return true
     /// if the authentication was successfully refreshed.
+    ///
+    /// **Concurrency:** This method may be called concurrently by multiple request
+    /// executors. Implementations must handle concurrent calls safely and avoid
+    /// unnecessary duplicate refresh operations.
     async fn refresh(&self) -> bool;
 }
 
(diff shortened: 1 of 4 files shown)


wp_api/src/auth.rs
```diff
@@ -70,8 +71,13 @@ impl ModifiableAuthenticationProvider {
 }
 
 #[uniffi::export(with_foreign)]
-pub trait WpDynamicAuthenticationProvider: Send + Sync {
+#[async_trait::async_trait]
+pub trait WpDynamicAuthenticationProvider: Send + Sync + Debug {
     fn auth(&self) -> WpAuthentication;
+
+    /// Refresh the authentication token. The implementation should only return true
+    /// if the authentication was successfully refreshed.
```
oguzkocer, line 79:
```suggestion
    /// Refresh the authentication token. The implementation should only return true
    /// if the authentication was successfully refreshed.
    ///
    /// **Concurrency:** This method may be called concurrently by multiple request
    /// executors. Implementations must handle concurrent calls safely and avoid
    /// unnecessary duplicate refresh operations.
```

oguzkocer (approved):
I have some reservations about adding even more complexity to the macro code, as I was hoping we can extract some of it. However, that's not an issue specific to this PR and it doesn't make things too much worse, so I think it's fine to ship it as is.

I only made one minor suggestion for improving the documentation for `refresh`.

PR #841, opened 2025-08-14
Parse videopress media details

crazytonyli:
Atomic sites (maybe all sites with Jetpack Videopress enabled) returns a `media_details` JSON object that is not compatible with the one in core. That means the `parse_as_mime_type` function does not return `VideoMediaDetails`.

We want to add a dedicated function to parse the `media_details` values provide by the VideoPress plugin.


oguzkocer (approved):

crazytonyli:
Rebased to include changes in https://github.com/Automattic/wordpress-rs/pull/842.

PR #862, opened 2025-08-29
Fix database restore SSL issues and improve test setup logging

oguzkocer:
**To Test:**
```
make test-server && make restore-test-server
```

### DIFF
--- a/wp_cli/src/lib.rs
+++ b/wp_cli/src/lib.rs
@@ -23,13 +23,23 @@ const BACKUP_PATH: &str = "/var/www/html/wp-content/dump.sql";
 
 pub fn restore_db() -> std::process::Output {
     Command::new("mariadb")
+        // Disable SSL to avoid connection errors
         .arg("--skip-ssl")
+        // Host flag
         .arg("-h")
+        // MySQL/MariaDB container hostname
         .arg("database")
+        // Username flag
         .arg("-u")
+        // Database username
         .arg("wordpress")
-        .arg("-pwordpress")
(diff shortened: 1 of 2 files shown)


wp_cli/src/lib.rs
```diff
@@ -17,7 +22,19 @@ pub use wp_cli_users::*;
...
 pub fn restore_db() -> std::process::Output {
-    run_wp_cli_command(["db", "import", BACKUP_PATH])
+    Command::new("mariadb")
+        .arg("--skip-ssl")
+        .arg("-h")
+        .arg("database")
+        .arg("-u")
+        .arg("wordpress")
+        .arg("-pwordpress")
+        .arg("wordpress")
```
crazytonyli, line 39: Is the password passed twice: `-pwordpress wordpress`?

oguzkocer: The second one is the database name. I've added comments to each argument in df9de223f2db7265f03fb0f96222c2612c9808b5.

It's not great to have these be part of the Rust file, but I really don't want to spend any more time on the setup right now and would rather finish the features I am working on.

oguzkocer: Splitting `-pwordpress` in https://github.com/Automattic/wordpress-rs/commit/df9de223f2db7265f03fb0f96222c2612c9808b5 didn't work, which should be addressed in 34892fc1fd502312543cb0db171354a77e0e9df7.

crazytonyli (approved):

PR #884, opened 2025-09-10
Add Freshly Pressed

jkmassel:
Adds the WP.com Freshly Pressed endpoints for integration into Reader.

### DIFF
--- a/wp_api/src/wp_com/endpoint/freshly_pressed.rs
+++ b/wp_api/src/wp_com/endpoint/freshly_pressed.rs
@@ -1,6 +1,9 @@
 use crate::{
     request::endpoint::{AsNamespace, DerivedRequest},
-    wp_com::{freshly_pressed::{FreshlyPressedListParams, FreshlyPressedPostList}, WpComNamespace},
+    wp_com::{
+        WpComNamespace,
+        freshly_pressed::{FreshlyPressedListParams, FreshlyPressedPostList},
+    },
 };
 use wp_derive_request_builder::WpDerivedRequest;
 
--- a/wp_api/src/wp_com/endpoint/freshly_pressed.rs
+++ b/wp_api/src/wp_com/endpoint/freshly_pressed.rs
@@ -1,0 +1,17 @@
+use crate::{
+    request::endpoint::{AsNamespace, DerivedRequest},
+    wp_com::{freshly_pressed::{FreshlyPressedListParams, FreshlyPressedPostList}, WpComNamespace},
+};
+use wp_derive_request_builder::WpDerivedRequest;
+
+#[derive(WpDerivedRequest)]
+enum FreshlyPressedRequest {
+    #[get(url = "/freshly-pressed", params = &FreshlyPressedListParams, output = FreshlyPressedPostList)]
+    List,
+}
+
+impl DerivedRequest for FreshlyPressedRequest {
+    fn namespace() -> impl AsNamespace {
+        WpComNamespace::RestV1_2
+    }
+}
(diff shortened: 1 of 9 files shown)


wp_api/src/wp_com/freshly_pressed.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub profile_url: String,
+    #[serde(alias = "site_ID", deserialize_with = "deserialize_u64_or_none")]
+    pub site_id: Option<u64>,
+    pub has_avatar: bool,
+    pub wpcom_id: u64,
+    pub wpcom_login: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct FreshlyPressedDiscussionSettings {
```
oguzkocer, line 205: If I am not mistaken, a `discussion` field needs to be added to `FreshlyPressedPost` and use this `struct`. At least, the sample json used in tests seem to include it.

wp_api/src/wp_com/freshly_pressed.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub comments_open: bool,
+    pub comment_status: String,
+    pub pings_open: bool,
+    pub ping_status: String,
+    pub comment_count: u32,
+}
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct FreshlyPressedEditorialSettings {
```
oguzkocer, line 214: This seems to be unused.

wp_api/src/wp_com/endpoint/freshly_pressed.rs
oguzkocer, line 1: Currently, all `endpoint` files have an `_endpoint` suffix, which this file should also follow.

wp_api/src/wp_com/freshly_pressed.rs
```diff
@@ -0,0 +1,287 @@
...
+pub struct FreshlyPressedDateRange {
+    pub before: WpGmtDateTime,
+    pub after: WpGmtDateTime,
+}
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct FreshlyPressedPost {
+    #[serde(rename = "ID")]
+    pub id: PostId,
+    #[serde(alias = "site_ID")]
```
oguzkocer, line 36: Unless we specifically need to `alias`, we should be consistent and use `rename` everywhere.

wp_api/src/wp_com/freshly_pressed.rs
```diff
@@ -0,0 +1,287 @@
...
+        Some(Self {
+            number: query_pairs.get(FreshlyPressedListParamsField::Number),
+            page: query_pairs.get(FreshlyPressedListParamsField::Page),
+        })
+    }
+
+    fn supports_pagination() -> bool {
+        true
+    }
+}
```
oguzkocer, line 271: These should no longer be necessary and should be replaced with the new macros you've reviewed yesterday.

oguzkocer (changes requested):
Thanks for the PR @jkmassel! I've left a few comments.

wp_api/src/wp_com/freshly_pressed.rs
```diff
@@ -0,0 +1,287 @@
...
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct FreshlyPressedDateRange {
+    pub before: WpGmtDateTime,
+    pub after: WpGmtDateTime,
+}
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct FreshlyPressedPost {
```
kean, line 33: "Freshly Pressed" seemingly uses the same response format as all other response streams in Reader: site, tags, feeds, discover. I suggest generalizing it from the start even if we don't end up using it everywhere originally.

jkmassel:
Closing this, because as @kean noted we should use the newer WP.com APIs if we're going to do this.

PR #885, opened 2025-09-11
`/pages` endpoint

oguzkocer:
Follows established patterns to implement [`/pages` endpoint.](https://developer.wordpress.org/rest-api/reference/pages/)


jkmassel (approved):
I tested this in the example app after adding Swift support, and it worked great.

PR #893, opened 2025-09-15
Add Feature Detection

jkmassel:
Adds the ability to detect namespace and endpoint registration. This makes it possible for applications to answer questions like:

1. Does this site support the Gutenberg Editor Manifest?
2. Does this site support using Application Passwords for XML-RPC, or to fetch post previews?

I debated whether to implement something like `ApiDetails::supports(WpSiteFeature::ApplicationPasswords)` but this seems like an application-level concern, not a library one. That means _some_ code duplication, but I think it's preferable, because an application will have a better idea of what features it cares about.

This PR spawned https://github.com/Automattic/jetpack/pull/45186

### DIFF
--- a/wp_api/src/login.rs
+++ b/wp_api/src/login.rs
@@ -1,5 +1,6 @@
 use crate::{
-    login::url_discovery::is_local_dev_environment_url, parsed_url::ParsedUrl, uuid::WpUuid,
+    JsonValue, login::url_discovery::is_local_dev_environment_url, parsed_url::ParsedUrl,
+    uuid::WpUuid,
 };
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, str, sync::Arc};
@@ -7,6 +8,7 @@ use wp_localization::{MessageBundle, WpMessages, WpSupportsLocalization};
 use wp_localization_macro::WpDeriveLocalizable;
 use wp_serde_helper::{
     deserialize_empty_array_or_hashmap, deserialize_false_or_string, deserialize_offset,
+    deserialize_string_vec_or_string_as_option,
 };
 
 const KEY_APPLICATION_PASSWORDS: &str = "application-passwords";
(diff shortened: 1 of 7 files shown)


wp_serde_helper/src/lib.rs
```diff
@@ -388,4 +450,45 @@ mod tests {
...
+    ) {
+        let string_vec_or_string: StringVecOrString =
+            serde_json::from_str(test_case).expect("Test case should be a valid JSON");
+        assert_eq!(expected_result, string_vec_or_string.string);
+    }
+
+    #[derive(Debug, Deserialize)]
+    pub struct OptionStringVecOrString {
+        #[serde(deserialize_with = "deserialize_string_vec_or_string_as_option")]
+        #[serde(default)]
```
jkmassel, line 476: I don't really like that we need this line, but I couldn't think of an easy way to avoid it.

wp_api/src/login.rs
```diff
@@ -66,6 +68,35 @@ pub struct WpApiDetails {
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, uniffi::Record)]
+pub struct WpEndpointArg {
+    pub required: bool,
+
+    pub default: Option<JsonValue>,
+    pub description: Option<String>,
+    #[serde(deserialize_with = "deserialize_string_vec_or_string_as_option")]
+    #[serde(default)]
+    pub r#type: Option<Vec<String>>,
+    pub r#enum: Option<Vec<String>>,
```
crazytonyli, line 97: What does the `r#` mean?

jkmassel: `type` and `enum` are reserved terms, so you need to do this to use them as field names

crazytonyli: TIL! https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html

crazytonyli (approved):

PR #897, opened 2025-09-17
Fix SwiftPM Dependency Issues

jkmassel:
I kept getting weird SwiftPM-related changes in my diffs, and finally tracked it down.

I'm not sure why, but Swift only changes `Package.resolved` after a clean build. This PR adds a CI step to validate that `Package.swift` is in step with `Package.resolved` to avoid changes not being properly propagated.


crazytonyli:
The `Package.resolved` would only get updated by the SPM commands, not by Xcode. I typically just use Xcode and rarely run `swift build|test`.

crazytonyli (approved):

jkmassel:
> The Package.resolved would only get updated by the SPM commands

Yeah, it was changed whenever I run `make lint-swift`, which was unexpected. It shouldn't matter if we don't have too many Swift dependencies, but I figure might as well address it while I'm seeing it.

PR #900, opened 2025-09-18
Fix Android Example App

jkmassel:
The Android App's build file was broken, and it wasn't using the same JDK as the rest of the project.

I've fixed these, and ensured it builds in CI.

This is mostly to test dependabot PRs, but we can also use the demo app to validate support for 16K page sizes.


adalpari (approved):
Building the example app works for me!

Image

PR #903, opened 2025-09-21
Add sqlite cache

jkmassel:
Adds an initial SQLite cache implementation that works on Android and iOS, implements row-level change updates, and has a migration mechanism.

We'll need the ability to store and fetch objects, but a lot of this should just be generated code instead of being hand-written – that's for a future PR.


crazytonyli:
If the database is designed to be used as a cache, do we even need to migrate the data? By definition, users should not expect _cached_ data always to be there, right?

jkmassel:
> If the database is designed to be used as a cache, do we even need to migrate the data

This is a great question. I see this as a replacement for Core Data / FluxC – so while we could just throw away the data every time we need to change something, it seems wasteful of system resources. For a large site, destroying the cache could mean re-downloading thousands of posts/media items/comments/etc. That's a lot of CPU time and battery that we don't _need_ to spend if we're a little bit careful about our data model.

I also like this model of migrations because each one can be small and easy to reason about – if something weird happens you'll know where it happened in the chain and can pick up where you left off.

Though this _is_ still a cache – if something goes wrong with it, the move is to destroy it and re-create it, so we don't need to be quite as careful as user data.

I think this PR represents a good balance between flexibility and being careful about user resources, but happy to discuss further if anyone disagrees!

crazytonyli:
> For a large site, destroying the cache could mean re-downloading thousands of posts/media items/comments/etc.

I think the "re-downloading" is going to happen regardless of whether the posts have been fetched and saved in the local SQLite database. When users go to view the post list, we'd have to re-download every post again. That's the only way we can guarantee that the most recent versions of the posts are displayed on the list.

wp_api/src/cache/wp_api_cache.rs
```diff
@@ -0,0 +1,226 @@
...
+    GLOBAL_QUERY_MONITOR.get_delegate()
+}
+
+#[uniffi::export]
+fn set_global_delegate(delegate: Arc<dyn DatabaseDelegate>) {
+    GLOBAL_QUERY_MONITOR.set_delegate(delegate);
+}
+
+struct DBManager {
+    connection: Connection,
```
oguzkocer: We should wrap this in a `Mutex` to make the connection thread-safe and eliminate the need for the unsafe `Send + Sync` implementations above. SQLite connections aren't thread-safe by default, so `Mutex<Connection>` would provide the necessary synchronization while automatically implementing the required traits for UniFFI.

wp_api/src/cache/wp_api_cache.rs
```diff
@@ -0,0 +1,226 @@
...
+        let next_migration_id = self.get_next_migration_index()?;
+        for (index, migration) in MIGRATION_QUERIES[next_migration_id..].iter().enumerate() {
+            for query in migration
+                .split(";")
+                .filter(|query| !query.trim().is_empty())
+            {
+                self.connection.execute(query, ())?;
+            }
+
+            self.insert_migration((index + 1) as u64)?;
```
oguzkocer:
```suggestion
            self.insert_migration((next_migration_id + index + 1) as u64)?;
```

The `enumerate()` call will start the index from `0` which means the current code will re-run the migrations from the start.

Here is a test that showcases the issue which is probably worth adding to our suite:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_migration_numbering_should_be_sequential() {
        let connection = Connection::open_in_memory().unwrap();
        let mut migration_manager = MigrationManager::new(&connection).unwrap();

        // Create migrations table and run first migration manually
        migration_manager.create_migrations_table().unwrap();
        migration_manager.insert_migration(1).unwrap();

        // Try to run remaining migrations
        // BUG: This currently fails because line 112 tries to insert migration ID 1 again
        // instead of migration ID 2. The fix should be:
        // self.insert_migration((next_migration_id + index + 1) as u64)?;
        let result = migration_manager.perform_migrations();

        // This test FAILS with the current bug - it should succeed after fixing line 112
        result.expect("Migration should succeed after fixing the numbering bug on line 112");

        // After the bug is fixed, verify migration IDs are sequential
        let mut stmt = connection
            .prepare("SELECT migration_id FROM _migrations ORDER BY migration_id")
            .unwrap();
        let migration_ids: Vec<u64> = stmt
            .query_map([], |row| row.get::<_, u64>(0))
            .unwrap()
            .collect::<Result<Vec<_>, _>>()
            .unwrap();

        // Verify migration IDs are sequential and complete
        let expected_ids: Vec<u64> = (1..=MIGRATION_QUERIES.len() as u64).collect();
        assert_eq!(
            migration_ids, expected_ids,
            "Migration IDs should be sequential from 1 to {}", MIGRATION_QUERIES.len()
        );
    }
}
```

wp_api/src/cache/wp_api_cache.rs
```diff
@@ -0,0 +1,226 @@
...
+
+#[uniffi::export]
+fn get_global_query_monitor() -> Option<Arc<dyn DatabaseDelegate>> {
+    GLOBAL_QUERY_MONITOR.get_delegate()
+}
+
+#[uniffi::export]
+fn set_global_delegate(delegate: Arc<dyn DatabaseDelegate>) {
+    GLOBAL_QUERY_MONITOR.set_delegate(delegate);
+}
```
oguzkocer: In my opinion, we should avoid using a global state like this. I imagine it's done to simplify the observer pattern, but using a global state comes at a cost. It makes tests difficult to isolate meaning the tests can be run in parallel. If we need a singleton type behavior, the correct place for that would be the application layer, not the library. The global state introduces ownership/cleanup concerns (not in Rust terms) and locks all clients to a specific pattern.

I suggest moving this to the `WpApiCache` to make it instance based.

wp_api/src/cache/wp_api_cache.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    pub fn perform_migrations(&self) -> Result<u64, SqliteDbError> {
+        Ok(MigrationManager::new(&self.inner.connection)?.perform_migrations()?)
+    }
+
+    pub fn flush(&self) -> Result<(), SqliteDbError> {
+        self.inner.connection.execute("commit", ())?;
+        Ok(())
+    }
```
oguzkocer: I'm unsure about the intended usage of this function. SQLite auto-commits by default unless you're in an explicit transaction, so calling `commit` here seems incorrect. Could we remove this function until it's actually needed? That way we'll have proper context about the intended behavior and can implement it correctly.

native/kotlin/api/android/build.gradle.kts
```diff
@@ -1,3 +1,6 @@
+import org.jetbrains.kotlin.konan.target.linker
+import kotlin.system.exitProcess
+
```
oguzkocer:
```suggestion
```
These don't seem to be used in this PR.

native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
```diff
@@ -0,0 +1,16 @@
+package rs.wordpress.api.cache.kotlin
+
```
oguzkocer:
```suggestion
```

Android Studio complains: `Package directive does not match the file location`

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/cache/kotlin/WordPressApiCache.kt
```diff
@@ -0,0 +1,47 @@
...
+    // Creates a new cache at the specified path
+    constructor(string: String) {
+        this.cache = WpApiCache(string)
+    }
+
+    fun performMigrations(): Int {
+        internalDispatcher.run {
+            return this@WordPressApiCache.cache.performMigrations().toInt()
+        }
+    }
```
oguzkocer:
```suggestion
    suspend fun performMigrations(): Int = withContext(internalDispatcher) {
        cache.performMigrations().toInt()
    }
```

native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+import rs.wordpress.cache.kotlin.WordPressApiCache
+import kotlin.test.assertEquals
+
+@Execution(ExecutionMode.CONCURRENT)
+class WordPressApiCacheTest {
+
+    @Test
+    fun testThatMigrationsWork() {
```
oguzkocer:
```suggestion
    fun testThatMigrationsWork() = runTest {
```

This will be necessary when `performMigrations` is updated to be a suspended function.

native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
```diff
@@ -0,0 +1,16 @@
+package rs.wordpress.api.cache.kotlin
+
+import org.junit.jupiter.api.Test
```
oguzkocer, line 2:
```suggestion
import kotlinx.coroutines.test.runTest
import org.junit.jupiter.api.Test
```

This will be necessary when `performMigrations` is updated to be a suspended function.

oguzkocer (changes requested):
@jkmassel Thanks for the POC! I've left some suggestions.

wp_api/src/cache/wp_api_cache.rs
jkmassel: I'd love if we could avoid this – I just couldn't figure out how to call `self` inside of `connection.update_hook` without massive ownership issues so I'm not sure how to narrow the scope. 

Any suggestions?

jkmassel: I'm curious your thoughts on [e7563df](https://github.com/Automattic/wordpress-rs/pull/903/commits/e7563dfbe77e5f83d3a3da58fdc105b7cca63e63).

We can inject the delegate into the `startListeningForUpdates` method, which in the presence of a wrapper (like in Swift) makes it pretty much transparent.

native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
jkmassel: [9acf4ef](https://github.com/Automattic/wordpress-rs/pull/903/commits/9acf4efa620475ebe9757b3ffb835e45c23821d7)

native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
jkmassel: [9acf4ef](https://github.com/Automattic/wordpress-rs/pull/903/commits/9acf4efa620475ebe9757b3ffb835e45c23821d7)

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/cache/kotlin/WordPressApiCache.kt
jkmassel: Added in [60409f3](https://github.com/Automattic/wordpress-rs/pull/903/commits/60409f379b7cc8ee235711f3706c399d65e764a1) – I didn't use the suggestion because I needed to also import `withContext`

wp_api/src/cache/wp_api_cache.rs
oguzkocer: Looks much better to me. Currently, I can't think of any issues with that approach.

oguzkocer:
I've mentioned this on Slack as well, but just leaving a note for myself. I've re-reviewed up to [0bedb9c](https://github.com/Automattic/wordpress-rs/pull/903/commits/0bedb9cd63237c528a19272b1e2cf56480958184) and it's looking great. Just a couple more unaddressed comments and I think we are good to merge this.

wp_api/src/cache/wp_api_cache.rs
oguzkocer: Removed in 5b3b34060716b68234394b7d4481a6087e3f6077.

wp_api/src/cache/wp_api_cache.rs
oguzkocer: Addressed in 5b3b34060716b68234394b7d4481a6087e3f6077.

oguzkocer (approved):
I believe all the issues in Rust are addressed. I am approving it so we can merge it once the Swift issues are resolved.

PR #904, opened 2025-09-22
Cache XCFramework build

jkmassel:


jkmassel:
Closing this because the performance improvements didn't materialize (again)

PR #922, opened 2025-09-26
Add page revisions and autosaves support

oguzkocer:
Unifies revisions and autosaves endpoints to work with both posts and pages using `PostEndpointType`.

- Updated `post_revisions` and `autosaves` types to use `AnyPostRevision` for both posts and pages
- Added `PostEndpointType` parameter to revision and autosave endpoints
- Enhanced setup script with page revision and autosave test data generation
- Added comprehensive integration tests for page revisions and autosaves


jkmassel (approved):
I tested with the Swift sample app and this worked great.

Because it's hierarchical (revisions belong to a specific post), I didn't include permanent UI in the sample app because that would've been a bunch of extra unrelated work.

PR #924, opened 2025-09-26
`/statuses` endpoint

oguzkocer:
Follows established patterns to implement create [`/statuses`](https://developer.wordpress.org/rest-api/reference/post-statuses/) endpoint.


jkmassel (approved):
Tested in the Swift example app; works great

PR #926, opened 2025-09-27
`/menus` endpoint

oguzkocer:
Follows established patterns to implement create [`/menus`](https://developer.wordpress.org/rest-api/reference/nav_menus/) endpoint.


jkmassel (approved):
Tested with the Swift sample app, works great

PR #929, opened 2025-09-29
`/menu-locations` endpoint

oguzkocer:
Follows established patterns to implement create [`/menu-locations`](https://developer.wordpress.org/rest-api/reference/menu-locations/) endpoint.

Note that I didn't even check if the `_fields` filtering works for this endpoint, because there are only 3 fields and filtering doesn't seem very useful. Adding filtering is super easy, but it adds a bunch of extra methods that'll not be used, so I decided to skip them.


jkmassel (approved):
As usual, I tested using the Swift Example app, and this works for me.

PR #939, opened 2025-09-30
`/menu-items` endpoint

oguzkocer:
Follows established patterns to implement create [`/menu-items`](https://developer.wordpress.org/rest-api/reference/nav_menu_items/) endpoint.


jkmassel (approved):
Tested with the Swift Example app and it works for me

PR #940, opened 2025-10-01
`/menu-items/<nav_menu_item_id>/autosaves` endpoint

oguzkocer:
Follows established patterns to implement create [`/menu-items/<nav_menu_item_id>/autosaves`](https://developer.wordpress.org/rest-api/reference/nav_menu_item-revisions/) endpoint.


jkmassel (approved):
This one isn't especially feasible to test in-depth because it's a child record type. It seems to work, so I just added Swift support and am approving.

PR #941, opened 2025-10-03
`/navigation` endpoint

oguzkocer:
Follows established patterns to implement [`/navigation`](https://developer.wordpress.org/rest-api/reference/wp_navigations/) endpoint.

I initially implemented it as a separate endpoint, but then @jkmassel and I discussed it and decided to make it a `PostEndpointType`. We do lose a bit static type information with this approach, so I kept the original implementation as well, so we can decide in this PR whether to keep the `PostEndpointType` approach or use a separate endpoint. Either way, I think it's good to have both approaches in the git history to be able to revert our decision if we regret it.

I haven't implemented Kotlin integration tests yet as I'd like to make this ^ decision first.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,26 +97,21 @@ WordPress REST API returns different fields depending on the `context` parameter
 
 3. **Define parameter types** for list/create/update operations:
    ```rust
-   #[derive(Debug, Default, PartialEq, Eq, uniffi::Record)]
+   #[derive(Debug, Default, PartialEq, Eq, uniffi::Record, WpDeriveParamsField)]
+   #[supports_pagination(true)]  // or false if endpoint doesn't support pagination
    pub struct UserListParams {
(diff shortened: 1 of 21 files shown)


jkmassel (approved):
I tested this and it works fine.

The enum wrappers for `title` and `content` are a bit weird, but it's workable. I think having these post params be optional make a lot of sense in the context of custom post types, so I think this approach works.

I'm good to move forward with it if you are.

oguzkocer:
> The enum wrappers for title and content are a bit weird, but it's workable. I think having these post params be optional make a lot of sense in the context of custom post types, so I think this approach works.

@jkmassel As mentioned in the PR description, we won't actually use the types in `navigations` module if we are going with the `PostEndpointType` implementation. They are removed in 5b887e47df5653fed9b007fa96305fcf1a48df79.

PR #942, opened 2025-10-03
`/navigation/<parent>/revisions` & `/navigation/<parent>/autosaves` endpoints

oguzkocer:
Follows established patterns to implement [`/navigation/<parent>/revisions` & `/navigation/<parent>/autosaves`](https://developer.wordpress.org/rest-api/reference/wp_navigation-revisions/) endpoints.


jkmassel (approved):
I added Swift type support – because this is a nested type it doesn't play nicely with the example app, but I'm confident it works well based on the code.

PR #948, opened 2025-10-06
Add WP.com `/rest/v1.1/me` endpoint

jkmassel:
Title explains it pretty well


wp_serde_helper/src/lib.rs
```diff
@@ -401,6 +538,73 @@ where
...
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    where
+        A: de::SeqAccess<'de>,
+    {
+        let mut vec = Vec::new();
+        while let Some(elem) = seq.next_element::<T>()? {
+            vec.push(elem);
+        }
+        Ok(vec)
+    }
```
jkmassel: @oguzkocer – this doesn't feel like the best way to do this – can we just pass the entire thing back without iterating it?

crazytonyli: This version without Visitor works too. I don't think it's worse performance-wise.

```rust

pub fn deserialize_null_as_empty_vec<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
where
    D: Deserializer<'de>,
    T: Deserialize<'de>,
{
    Option::<Vec<T>>::deserialize(deserializer)
        .map(|opt| opt.unwrap_or_default())
}
```

jkmassel: Ah this is nice! I adopted it in [6702cba](https://github.com/Automattic/wordpress-rs/pull/948/commits/6702cba6b2fbb21757fe98a32ac7a2e59dbc96f1) with a slight modification to make the error messaging match.

Thanks!!

wp_api/src/wp_com/me.rs
```diff
@@ -0,0 +1,167 @@
...
+    deserialize_empty_string_as_none, deserialize_null_as_empty_vec,
+    deserialize_string_vec_or_string, deserialize_u64_or_none,
+    deserialize_u64_or_none_with_negative_as_none, deserialize_u64_or_none_with_zero_as_none,
+};
+
+#[derive(Debug, Serialize, Deserialize, uniffi::Record)]
+pub struct WPComUserInfo {
+    /// The user's WP.com ID.
+    #[serde(rename = "ID")]
+    pub id: u64,
```
crazytonyli, line 14: Nitpick: use `UserId` instead of `u64`?

crazytonyli (approved):

wp_api/src/wp_com/me.rs
jkmassel: This is already in the next PR 😁

PR #953, opened 2025-10-09
Don't generate uniffi bindings for Request Builders

oguzkocer:
We are hitting `MethodTooLargeException` issues in Kotlin bindings due to the number of methods we are generating, so we are going to stop exporting some of the ones we are not actively using at the moment.


jkmassel (approved):

oguzkocer:
We are admin merging this one because the CI errors are unrelated and we need this as a separate fix - possibly to revert in the future.

PR #966, opened 2025-10-16
Add xcframework code signing

jkmassel:
Adds code signing to the release and debug XCFrameworks. This allows consumer projects to ensure the binary dependency hasn't been tampered with.

### DIFF
--- a/.buildkite/download-xcframework.sh
+++ b/.buildkite/download-xcframework.sh
@@ -4,6 +4,14 @@ set -euo pipefail
 
 echo "--- :arrow_down: Downloading xcframework"
 buildkite-agent artifact download libwordpressFFI.xcframework.zip . --step "xcframework"
+unzip libwordpressFFI.xcframework.zip -d .
+rm libwordpressFFI.xcframework.zip
+
 buildkite-agent artifact download 'native/swift/Sources/wordpress-api-wrapper/*.swift' . --step "xcframework"
+
+# Temporarily also download to target folder because some parts still expect it there
+mkdir -p ./target
+pushd target
+buildkite-agent artifact download libwordpressFFI.xcframework.zip . --step "xcframework"
 unzip libwordpressFFI.xcframework.zip -d .
 rm libwordpressFFI.xcframework.zip
--- a/.buildkite/download-xcframework.sh
+++ b/.buildkite/download-xcframework.sh
@@ -3,6 +3,7 @@
 set -euo pipefail
 
 echo "--- :arrow_down: Downloading XCFramework"
+mkdir -p target
 pushd target
 buildkite-agent artifact download libwordpressFFI.xcframework.zip . --step "xcframework"
 unzip libwordpressFFI.xcframework.zip -d .
--- a/.buildkite/download-xcframework.sh
+++ b/.buildkite/download-xcframework.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 echo "--- :arrow_down: Downloading xcframework"
-buildkite-agent artifact download target/libwordpressFFI.xcframework.zip . --step "xcframework"
+buildkite-agent artifact download libwordpressFFI.xcframework.zip . --step "xcframework"
 buildkite-agent artifact download 'native/swift/Sources/wordpress-api-wrapper/*.swift' . --step "xcframework"
-unzip target/libwordpressFFI.xcframework.zip -d .
-rm target/libwordpressFFI.xcframework.zip
+unzip libwordpressFFI.xcframework.zip -d .
+rm libwordpressFFI.xcframework.zip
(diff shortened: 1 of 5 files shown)


fastlane/Fastfile
```diff
@@ -396,6 +411,27 @@ lane :generate_fluent_file_from_po do |file_path:|
...
+lane :set_up_signing do |readonly: true|
+  require_env_vars!(*ASC_API_KEY_ENV_VARS, *CODE_SIGNING_STORAGE_ENV_VARS)
+
+  sync_code_signing(
+    platform: 'macos',
+    app_identifier: APPLE_BUNDLE_IDENTIFIER,
+    team_id: APPLE_TEAM_ID,
+    api_key: app_store_connect_api_key,
+    type: 'development',
+    certificate_id: 'Apple Development: Created via API (886NX39KP6)',
```
crazytonyli: What do you think about using the "Apple Distribution" one instead? I feel like it's appropriate to use binaries that are signed with "Apple Distribution" on all builds, but "Apple Development" builds may not be suitable for release builds.

fastlane/Fastfile
```diff
@@ -24,6 +24,21 @@ PROJECT_NAME = 'wordpress-rs'
 # GlotPress configuration
 GLOTPRESS_PROJECT_BASE_URL = 'https://translate.wordpress.com/projects/mobile/wordpress-rs'
 
+# Code Signing
+APPLE_TEAM_ID = 'PZYM8XX95Q'
+APPLE_BUNDLE_IDENTIFIER = 'com.automattic.hostmgr'
```
crazytonyli: If I understand it correctly, this bundle id is only for fastlane to download the certificates and things. If that's the case, can we move it into the `set_up_signing` lane, to avoid the potential confusion that `com.automattic.hostmgr` is the "bundle id" of the wordpress-rs xcframework?

mokagio:
Sorry, I meant to create a dedicated branch for the distribution certificate (tracked in https://linear.app/a8c/issue/AINFRA-1477/make-a-release-mode-code-signing-certificate) but I pushed without realizing I was on this branch...

mokagio:
I know how to fix the build failures caused by the change in the artifact path... but my Docker is bricked and that is blocking my shell from entering into the this repo's folder -.-'

Makefile
```diff
@@ -137,6 +139,9 @@ xcframework-package: xcframework-all
 xcframework-package-checksum:
 	swift package compute-checksum libwordpressFFI.xcframework.zip | tee libwordpressFFI.xcframework.zip.checksum.txt
 
+xcframework-sign:
+	codesign --timestamp -v --sign "${certificate_name_release}" target/libwordpressFFI.xcframework
```
mokagio, line 143: Working good. See https://buildkite.com/automattic/wordpress-rs/builds/4086/steps/canvas?sid=019a538a-c172-4ce7-9a98-a586299292e5#019a538a-c1c1-4c09-9d8d-36437a2435fb/704-2883

image

.buildkite/download-xcframework.sh
```diff
@@ -3,7 +3,15 @@
...
 buildkite-agent artifact download 'native/swift/Sources/wordpress-api-wrapper/*.swift' . --step "xcframework"
-unzip target/libwordpressFFI.xcframework.zip -d .
-rm target/libwordpressFFI.xcframework.zip
+
+# Temporarily also download to target folder because some parts still expect it there
+mkdir -p ./target
+pushd target
+buildkite-agent artifact download libwordpressFFI.xcframework.zip . --step "xcframework"
+unzip libwordpressFFI.xcframework.zip -d .
+rm libwordpressFFI.xcframework.zip
```
mokagio: @jkmassel @crazytonyli This PR changed the ZIP destination from `target/` to `./` in https://github.com/Automattic/wordpress-rs/pull/966/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53L74

I thought it would be enough to update the download call at the start of the script (see above) but there evidently are other parts of the automation that expect the file to be in `target/`, because to make the build pass I had to add this additional ZIP expansion in the `target/` folder.

I'm happy to help tidy this up. Alternatively, we could restore the ZIP-to-target-folder behavior and leave it at that. Up to you.

crazytonyli: Looks like `make xcframework-package` was introduced to replace the `zip` command. Maybe we can update `xcframework-package` to put the zip file at `target/`, and you don't need the new script here to unzip twice?

mokagio: I did that and now there's other stuff that fails...

I gotta go offline for a bit but I'll get to the bottom of this later/tomorrow

crazytonyli: Was the issue around the unzipped content not in the `target` directory? It's probably because the folder structures of the `zip`-ed and `ditto`-ed zip files are different. 😥

crazytonyli (approved):

PR #976, opened 2025-10-22
Caching support

oguzkocer:
## Summary

Implements SQLite-based caching infrastructure for WordPress API data with multi-site support and normalized term relationships.

### Repository Pattern with Type-Safe Transactions

The implementation uses three distinct traits to model database operations:
- `QueryExecutor` - Base trait for executing queries (implemented by both `Connection` and `Transaction`)
- `TransactionManager` - Trait for types that can create transactions (only `Connection`)
- `InTransaction` - Marker trait for transaction contexts (only `Transaction`)

Methods requiring atomicity (like `sync_terms_for_object`) accept `&impl InTransaction`, enforcing at compile-time that they must run within a transaction while maintaining the trait abstraction layer.

### Multi-Site Architecture

Introduced `DbSite` wrapper around `RowId` to enforce site isolation at the type level. All queries are scoped to a specific site via foreign key constraints, enabling support for multiple WordPress sites in a single database without data leakage.

### Normalized Term Relationships

Categories and tags are stored in a separate `term_relationships` table using WordPress object IDs (not SQLite rowids) as the foreign key. This normalization:
- Supports efficient batch loading of terms for multiple posts
- Minimizes database observer events by only syncing actual changes
- Maintains referential integrity through cascading deletes

### Cache Freshness Tracking

Posts include a `last_fetched_at` timestamp (ISO 8601 UTC) automatically set on insert/update via SQLite's `strftime()`. This enables cache invalidation strategies without requiring application-level timestamp management.

### Atomic Upsert with Term Sync

The `PostRepository::upsert` method handles insert-or-update and term relationship sync as a single atomic operation. The implementation uses SQLite's `UPSERT` with conflict resolution on `(db_site_id, id)` to ensure consistency.

## Implementation Details

- **Type wrappers**: `RowId` for SQLite rowids, `DbSite` for site scoping
- **Mapping layer**: Bidirectional conversions between API types and database representations
- **Builder pattern**: `PostBuilder` test fixture with ergonomic defaults
- **Comprehensive test coverage**: Constraint validation, multi-site isolation, transaction rollback scenarios

## Migration Path

Three migrations establish the schema:
1. `0001-create-sites-table.sql` - Site registry
2. `0002-create-posts-table.sql` - Posts with edit context
3. `0003-create-term-relationships.sql` - Normalized term associations


wp_mobile_cache/migrations/0002-create-posts-table.sql
```diff
@@ -0,0 +1,66 @@
...
+
+  -- Optional enums (stored as TEXT)
+  `comment_status` TEXT,
+  `ping_status` TEXT,
+  `format` TEXT,
+
+  -- Complex optional fields (JSON)
+  `meta` TEXT,
+  `categories` TEXT,
+  `tags` TEXT,
```
oguzkocer: These are now stored in the relationship table.

wp_mobile_cache/src/mappings/posts.rs
```diff
@@ -0,0 +1,333 @@
...
+        // Basic format check: YYYY-MM-DDTHH:MM:SS.fffZ
+        assert!(
+            timestamp.len() >= 20,
+            "Timestamp should be at least 20 chars: {}",
+            timestamp
+        );
+    }
+
+    #[rstest]
+    fn test_round_trip_with_minimal_fields(mut test_db: Connection, test_site: DbSite) {
```
oguzkocer: We can combine these two tests into cases with minimal and full as they test the same thing.

wp_mobile_cache/src/term_relationships.rs
```diff
@@ -0,0 +1,25 @@
...
+
+/// Represents a term relationship in the database.
+///
+/// This associates an object (post, page, nav item, etc.) with a WordPress term
+/// for a specific taxonomy (category, tag, custom taxonomy).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DbTermRelationship {
+    /// SQLite rowid of this relationship
+    pub row_id: RowId,
+
```
oguzkocer: Remove the extra spaces.

wp_mobile_cache/src/test_helpers.rs
```diff
@@ -0,0 +1,106 @@
...
+/// #[rstest]
+/// fn test_something(test_ctx: TestContext) {
+///     let post = PostBuilder::new().build();
+///     test_ctx.repo.insert(&test_ctx.conn, &post, &test_ctx.site).unwrap();
+/// }
+/// ```
+pub struct TestContext {
+    pub conn: Connection,
+    pub site: DbSite,
+    pub repo: PostRepository,
```
oguzkocer: Should be post_repo.

wp_mobile_cache/src/test_helpers.rs
```diff
@@ -0,0 +1,106 @@
...
+        .execute("INSERT INTO sites (id) VALUES (2)", [])
+        .expect("Failed to insert second test site");
+    (test_db, DbSite { row_id: RowId(2) })
+}
+
+/// Fixture: PostRepository instance.
+///
+/// Zero-cost abstraction since PostRepository is zero-sized.
+#[fixture]
+pub fn post_repo() -> PostRepository {
```
oguzkocer: Should we always use the `TestContext` and add the second site there as well?

wp_mobile_cache/src/repository/posts_transaction_tests.rs
```diff
@@ -0,0 +1,186 @@
+//! Transaction rollback tests for PostRepository.
+//!
+//! These tests verify that transaction failures properly rollback database state
+//! without leaving partial writes or corrupted data.
```
oguzkocer: This comment doesn't seem entirely correct as there are successful test cases as well.

wp_mobile_cache/src/repository/term_relationships.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    /// Delete specific terms for an object.
+    fn delete_terms(
+        &self,
+        executor: &impl QueryExecutor,
+        site: &DbSite,
+        object_id: RowId,
+        taxonomy_type: &TaxonomyType,
+        term_ids: &[TermId],
```
oguzkocer: We probably don't ever need to delete certain term_ids.

wp_mobile_cache/src/repository/term_relationships.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    /// Insert new terms for an object.
+    fn insert_terms(
+        &self,
+        executor: &impl QueryExecutor,
+        site: &DbSite,
+        object_id: RowId,
+        taxonomy_type: &TaxonomyType,
+        term_ids: &[TermId],
```
oguzkocer: We probably don't need to insert term_ids directly and should use the sync one instead.

wp_mobile_cache/src/repository/term_relationships.rs
```diff
@@ -0,0 +1,483 @@
...
+        let sql = "SELECT taxonomy_type, term_id FROM term_relationships WHERE db_site_id = ? AND object_id = ?";
+        let mut stmt = executor.prepare(sql)?;
+        let rows = stmt.query_map(rusqlite::params![site.row_id, object_id], |row| {
+            let taxonomy_str: String = row.get(0)?;
+            let term_id: i64 = row.get(1)?;
+            Ok((taxonomy_str, term_id))
+        })?;
+
+        let mut result: HashMap<TaxonomyType, Vec<TermId>> = HashMap::new();
+        for row_result in rows {
```
oguzkocer: Would be good to use functional programming if we can. (there are other for statements in this PR)

wp_mobile_cache/src/repository/posts.rs
```diff
@@ -0,0 +1,898 @@
...
+        self.populate_terms(executor, site, &mut db_post)?;
+
+        Ok(db_post)
+    }
+
+    /// Select posts by author user ID for a given site (returns wrappers with rowids).
+    ///
+    /// Returns an empty vector if no posts by the given author exist for the site.
+    /// Automatically populates categories and tags from term_relationships table.
+    pub fn select_by_author(
```
oguzkocer: We should probably drop this and select_by_status for now.

wp_mobile_cache/src/repository/posts.rs
```diff
@@ -0,0 +1,898 @@
...
+    /// Term relationships are synced using a diff approach - only changes generate DB events.
+    ///
+    /// Returns the rowid of the inserted or updated row.
+    pub fn upsert(
+        &self,
+        transaction_manager: &mut impl TransactionManager,
+        site: &DbSite,
+        post: &AnyPostWithEditContext,
+    ) -> Result<RowId, SqliteDbError> {
+        use crate::mappings::helpers::{bool_to_integer, serialize_value_to_json};
```
oguzkocer: Move this to the top of the file.

wp_mobile_cache/src/repository/posts.rs
```diff
@@ -0,0 +1,898 @@
...
+        // Insert post
+        let rowid = repo.upsert(&mut test_db, &test_site, &post).unwrap();
+
+        // Retrieve and validate last_fetched_at
+        let retrieved = repo.select_by_rowid(&test_db, &test_site, rowid).unwrap();
+
+        // Validate ISO 8601 UTC format
+        assert!(retrieved.last_fetched_at.ends_with('Z'));
+        assert!(retrieved.last_fetched_at.contains('T'));
+        assert!(retrieved.last_fetched_at.len() >= 20);
```
oguzkocer: We should use the common helper for this.

wp_mobile_cache/migrations/0002-create-posts-table.sql
```diff
@@ -0,0 +1,64 @@
...
+  `parent` INTEGER,
+  `menu_order` INTEGER,
+
+  -- Optional enums (stored as TEXT)
+  `comment_status` TEXT,
+  `ping_status` TEXT,
+  `format` TEXT,
+
+  -- Complex optional fields (JSON)
+  `meta` TEXT,
```
oguzkocer, line 36: This will be handled in a relationship table in a subsequent PR.

wp_mobile_cache/src/repository/posts.rs
```diff
@@ -0,0 +1,951 @@
...
+    taxonomies::TaxonomyType,
+};
+
+/// Repository for managing posts in the database.
+///
+/// Provides CRUD operations and post-specific query methods.
+pub struct PostRepository;
+
+impl PostRepository {
+    const TABLE_NAME: &'static str = "posts_edit_context";
```
jkmassel, line 22: If we need a non-edit context, will we need to modify this?

Should we just have `EDIT_CONTEXT_TABLE_NAME` from the start?

Or should this be `PostEditContextRepository`?

wp_mobile_cache/src/repository/posts.rs
```diff
@@ -0,0 +1,951 @@
...
+        Ok(posts)
+    }
+
+    /// Select a post by its WordPress post ID for a given site (returns wrapper with rowid).
+    ///
+    /// This is different from `select_by_rowid` which uses the SQLite rowid.
+    /// The post_id is the WordPress post ID from the REST API.
+    ///
+    /// Returns an error if no post with the given WordPress post ID exists for this site.
+    /// Automatically populates categories and tags from term_relationships table.
```
jkmassel, line 109: I'm a little bit surprised by this – I'd expect `Result<Option<DbAnyPostWithEditContext>, SqliteDbError>` where `SqliteDbError` is reserved for something going wrong with the database and a "not found" is a `.None` optional.

WDYT?

jkmassel (changes requested):

wp_mobile_cache/src/repository/posts.rs
oguzkocer: In our initial discussion, we said we might only support the `edit` context for most types, because the applications should work with that. We mentioned that in some cases it might be `view` context, but even then, we'd only be supporting 1 context type to make it easier for clients to choose which context they want.

With that in mind, I kept the name of the repository and the constant simple, but we can rename them both at any time. The only one that might be tricky is the table name `posts_edit_context`, but I think that one is accurate?

Having said all that, I am totally happy to rename anything you like.

wp_mobile_cache/src/repository/posts.rs
oguzkocer: To be honest, I haven’t given too much attention to the error handling part of the design yet, as I thought that part is fairly easy to modify. Using `Result<Option<DbAnyPostWithEditContext>, SqliteDbError>` makes sense to differentiate the "not found" state. Having said that, I don't think `SqliteDbError` can only be about the DB errors - not in the current design anyway. That's because there are other error types that needs to be represented such as issues with mapping. We might want something like:

```
enum WpDbError {
    SqliteDbError {
        inner: SqliteDbError,
    },
    MappingError {
        ..
    },
}
```

I _think_ I have a tendency to leave the error design to a later stage as I like to gather more information about different states. I've done the same thing with `wp_api`'s error types as well, but I am not sure if that's a good thing or a bad one. I think committing early can sometimes make it harder to figure out where some error types will belong, so I keep it simple at first and then organize it as things get fleshed out more. However, maybe that's not necessary in this case as it's a bit easier to imagine the types of errors we'll deal with in a caching layer.

jkmassel: Yeah, we probably want a more robust error layer. I still think that separating "error" from "not found" is valuable in the DB layer because "not found" is a very normal thing whereas "error" should be very rare.

We can change this later if needed though.

wp_mobile_cache/src/repository/posts.rs
jkmassel: > The only one that might be tricky is the table name posts_edit_context, but I think that one is accurate

Yep, the table name was actually what made me think of this.

I think we're fine to have a breaking change later if needed on this.

oguzkocer:
We had a discussion with @jkmassel on Slack. We'll land this as is and then I'll address his 2 comments and a generic way to deal with contexts in a subsequent PR.

jkmassel (approved):

PR #977, opened 2025-10-23
Generalize the file uploading implementation

crazytonyli:
At the moment, the file upload API is hand-written and hard-coded for the `wp/v2/media` endpoint. This PR implements a general solution for sending multipart/form-data POST requests.

The existing core media upload endpoint has been migrated to the new approach. To summarize, this is how to implement an endpoint that accepts file uploading via multipart/form-data requests:

```rust

#[derive(WpDerivedRequest)]
enum MediaRequest {
    // 1. Use `multipart = true` to mark the endpoint as supporting uploading files.
    #[post(url = "/media", params = &MediaCreateParams, output = MediaWithEditContext, multipart = true)]
    Create,
}

pub struct MediaCreateParams {
    // ...

    // 2. skip serializing the file-related properties.
    #[serde(skip)]
    pub file_path: String,
}

// 3. Implement the `RequiresMultipartForm` trait to provide the files to be uploaded.
impl RequiresMultipartForm for MediaCreateParams {
    fn multipart_form_files(&self) -> HashMap<String, MultipartFormFile> {
        // Key: the parameter name of the form field.
        // Value: file path, mime type, etc.
    }
}
```

### DIFF
--- a/native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
+++ b/native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
@@ -14,13 +14,12 @@ import okhttp3.RequestBody
 import okhttp3.RequestBody.Companion.asRequestBody
 import okhttp3.RequestBody.Companion.toRequestBody
 import uniffi.wp_api.InvalidSslErrorReason
-import uniffi.wp_api.MediaUploadRequest
-import uniffi.wp_api.MediaUploadRequestExecutionException
 import uniffi.wp_api.RequestContext
 import uniffi.wp_api.RequestExecutionErrorReason
 import uniffi.wp_api.RequestExecutionException
 import uniffi.wp_api.RequestExecutor
 import uniffi.wp_api.RequestMethod
+import uniffi.wp_api.WpMultipartFormRequest
 import uniffi.wp_api.WpNetworkHeaderMap
 import uniffi.wp_api.WpNetworkRequest
 import uniffi.wp_api.WpNetworkResponse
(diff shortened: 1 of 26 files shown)


native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
```diff
@@ -85,54 +84,57 @@ class WpRequestExecutor(
...
-            val file = fileResolver.getFile(mediaUploadRequest.filePath())
-            if (file == null || !file.canBeUploaded()) {
-                throw MediaUploadRequestExecutionException.MediaFileNotFound(mediaUploadRequest.filePath())
+            request.files().forEach { (name, fileInfo) ->
+                val file = fileResolver.getFile(fileInfo.filePath)
+                if (file == null || !file.canBeUploaded()) {
+                    throw RequestExecutionException.MediaFileNotFound(filePath = fileInfo.filePath)
+                }
+                val mimeType = fileInfo.mimeType ?: "application/octet-stream"
+                val requestBody = getRequestBody(file, mimeType, uploadListener)
```
crazytonyli, line 101: @oguzkocer I'm not sure if the progress reporting is still working as expected. The previous implementation only supported uploading one file, but the new implementation supports uploading many files. I'm not sure if the `uploadListener` still works.

wp_api/src/wp_com/support_tickets.rs
```diff
@@ -15,9 +19,30 @@ pub struct CreateSupportTicketParams {
...
 
+impl RequiresMultipartForm for CreateSupportTicketParams {
+    fn multipart_form_files(&self) -> HashMap<String, MultipartFormFile> {
+        self.attachments
+            .iter()
+            .enumerate()
+            .map(|(i, file_path)| {
+                (
+                    // TODO: The backend is not ready yet. This name may need to be changed.
+                    format!("attachment_{i}"),
```
crazytonyli, line 34: @jkmassel FYI, I presumed the attachment parameter is named "attachment_x" (and the app can control how many attachments are allowed). Still, we can always change this implementation after you have implemented the server side.

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
oguzkocer: This should be OK for this PR because `/media` only supports uploading a single file.

I am working on adding support for multiple files. However, since I also have a few other fixes/improvements in the Kotlin wrapper, so I want to split it into its own PR.

oguzkocer (approved):
Looks good :shipit:

native/kotlin/api/kotlin/src/main/kotlin/rs/wordpress/api/kotlin/WpRequestExecutor.kt
crazytonyli: > I am working on adding support for multiple files.

Just to double check, by "support", do you mean support tracking the progress of uploading multiple files?

oguzkocer: Kind of, it's tracking cumulative progress. https://github.com/Automattic/wordpress-rs/pull/979.

PR #983, opened 2025-10-27
Make support attachment dimensions strongly typed

jkmassel:
Adds a test that validates attachment parsing and typed lookup for metadata.

### DIFF
--- a/wp_api/src/wp_com/support_tickets.rs
+++ b/wp_api/src/wp_com/support_tickets.rs
@@ -76,14 +76,24 @@ pub enum SupportMessageAuthor {
 }
 
 #[derive(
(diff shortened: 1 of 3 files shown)


wp_api/src/wp_com/support_tickets.rs
crazytonyli: 🤔 This variant should be marked as `#[serde(untagged)]`, I think?

jkmassel: You're right! Addressed (with tests) in 94fedd148bd4e4b6d0dc2c0055d613dddda6be84

crazytonyli (approved):

PR #985, opened 2025-10-28
Export taxonomy details serialize and deserialize functions

crazytonyli:
What says in the title.

### DIFF
--- a/native/swift/Sources/wordpress-api/Exports.swift
+++ b/native/swift/Sources/wordpress-api/Exports.swift
@@ -192,6 +192,9 @@ public typealias TermsRequestListWithEmbedContextResponse = WordPressAPIInternal
 
 // MARK: - Taxonomies
 public typealias TaxonomyListParams = WordPressAPIInternal.TaxonomyListParams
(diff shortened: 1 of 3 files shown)


jkmassel (commented):
We can have uniffi export an auto-derived `Codable` implementation too – would that be preferable to this?

Why do we need it for these types specifically?

crazytonyli:
> We can have uniffi export an auto-derived Codable implementation too

My concern is that you may have two different JSON structures: one expressed via the default `Codable` implementation in Swift, and the other via the serde attributes in the Rust layer. That means if we somehow pass JSON encoded by Swift to Rust, Rust would fail to decode it, and vice versa. Here is an example:
```
// In Rust
// The expected JSON schema is `{ "first_name": "Adam" }`
struct Person {
  first_name: String
}

// In Swift
// The expected JSON schema is `{ "firstName": "Adam" }`
struct Person {
  var firstName: String
}
```

> Why do we need it for these types specifically?

Sorry, I should have provided more context. I need to store the taxonomy details on the app side.

crazytonyli:
@jkmassel I have tried your suggestion of copying the properties to another type, instead of storing the original full JSON. That worked out well. So, I'll close this PR.

PR #987, opened 2025-10-28
Expand progress tracking to all API calls

crazytonyli:
The current progress tracking is only implemented for the wp/v2/media endpoint. This PR expands the progress tracking to all API calls.

The existing `uploadMedia` API is kept. Its implementation also showcases how the new, more general API `fulfill(progress:withApiCall:)` is used.


jkmassel (approved):
I initially wasn't seeing progress in the example app when uploading images. This seems to be a `URLSession` issue – it just doesn't emit progress events very frequently for smaller files. When I tried with a 20MB images, I got the expected behaviour – more frequent updates over time.

PR #989, opened 2025-10-28
Support uploading files in the support tickets endpoints

crazytonyli:
The `CreateSupportTicket` and `AddMessageToSupportConversation` endpoints now support uploading files via their `attachments` parameter.


adalpari:
I tested it and it seems to be working on the Android side. [I found a problem with tags](https://a8c.slack.com/archives/C06S1QS55K9/p1761727172320149?thread_ts=1761678140.975589&cid=C06S1QS55K9), but it seems to be a server issue.

I'm not approving because I'm not familiar with the codebase, but as said, tests look good! 👍

oguzkocer (approved):

PR #991, opened 2025-10-29
Implement generic contextual repository pattern for posts

oguzkocer:
## Summary

Implements generic contextual repository pattern for WordPress REST API caching and applies it to posts. WordPress returns different field sets based on the `context` parameter (`edit`, `view`, `embed`), so we need separate database tables for each context.

This PR establishes the generic infrastructure and implements it for posts. Future entities (comments, users, etc.) will follow the same pattern.

## Implementation

Added three context tables for posts:

- **Edit context** (36 columns) - Full field set including raw content, passwords, permalink templates
- **View context** (29 columns) - Public-facing fields without sensitive data or raw content
- **Embed context** (14 columns) - Minimal fields for embedding posts in other content

## Key Technical Decisions

**Lazy term loading**: The `PostContext` trait uses `FnOnce()` closures for optional data loading. This allows EmbedContext (which doesn't include categories/tags) to avoid unnecessary database queries by simply not calling the closure.

**RETURNING optimization**: All upsert operations use SQLite's `RETURNING rowid` clause instead of separate SELECT queries, eliminating 3 queries per upsert across the three contexts.

**Column index verification**: PRAGMA-based integration tests verify that column enum ordering matches actual SQL schema, preventing bugs from migration/enum mismatches.

## Architecture

- `context.rs` - Generic `IsContext` trait and marker types (EditContext, ViewContext, EmbedContext)
- `db_types/posts/` - Column enums and wrapper structs (types only, no logic)
- `repository/posts.rs` - Entity-specific `PostContext` trait and all database operations
- `docs/ADDING_NEW_ENTITIES.md` - Guide for adding future entity types (comments, users, etc.)


wp_mobile_cache/docs/ADDING_NEW_ENTITIES.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Key Design Decisions
+
+**Column enum ordering**: Column indexes in the enum must match the exact order in SQL `CREATE TABLE` statements and all `SELECT *` queries. Add PRAGMA-based integration tests to verify this (see `test_post_edit_context_column_enum_matches_schema` in `repository/posts.rs` for reference).
+
+**Lazy data loading**: If your entity requires optional related data (e.g., posts use term relationships for categories/tags), use `FnOnce()` closures in the trait method. This allows contexts that don't need the data to avoid unnecessary database queries by simply not calling the closure. See `PostContext::from_row_with_terms` for an example.
+
+**Repository pattern**: All database operations belong in `repository/` - this includes both SQL execution AND row-to-type mapping logic. The `db_types/` module contains only type definitions.
+
+**RETURNING optimization**: Use SQLite's `RETURNING rowid` clause in upsert statements to eliminate separate SELECT queries. Our bundled SQLite version (3.50.2) fully supports this.
```
jkmassel, line 71: Probably worth mentioning the minimum iOS version supported (3.36.0), which also supports this.

wp_mobile_cache/src/repository/posts.rs
```diff
@@ -391,7 +1020,8 @@ mod tests {
         let retrieved = test_ctx
             .post_repo
             .select_by_rowid(&test_ctx.conn, &test_ctx.site, rowid)
-            .unwrap();
+            .unwrap()
```
jkmassel: In some of these we have `unwrap().expect()` and in others we have `expect().expect()`

The latter is probably better in tests but WDYT?

wp_mobile_cache/src/lib.rs
```diff
@@ -320,4 +324,28 @@ mod tests {
             MIGRATION_QUERIES.len()
         );
     }
+
+    #[test]
+    fn test_bundled_sqlite_version_supports_returning() {
```
jkmassel, line 394: I wonder if we can or should expose a function like this within the library layer, then expose it via FFI so that we can run the same check on iOS (where we use the system sqlite)

jkmassel (changes requested):
Left a few suggestions, but this generally looks solid.

wp_mobile_cache/docs/ADDING_NEW_ENTITIES.md
oguzkocer: Should I mention that the minimum supported iOS version for this library will be iOS 15 [which bundles `3.36.0`](https://github.com/yapstudios/YapDatabase/wiki/SQLite-version-(bundled-with-OS))?

wp_mobile_cache/src/repository/posts.rs
oguzkocer: Addressed in c1bf4f08166b7f010ded0ec89fea889d61ef4541.

wp_mobile_cache/src/lib.rs
oguzkocer: Added in 465f1bf5d174b23dc3b519fd8bc74d2b7ff0fe68. This is not exposed through FFI as `wp_mobile_cache` doesn't have `uniffi` as a dependency, but we can add a function to `wp_mobile` crate.

wp_mobile_cache/docs/ADDING_NEW_ENTITIES.md
jkmassel: I think that's great

jkmassel (approved):
Looks great now, thanks!!

wp_mobile_cache/docs/ADDING_NEW_ENTITIES.md
oguzkocer: Added in 65a2f18b2d39f9a3ddc9312188cff7ee2e195782.

PR #995, opened 2025-10-30
Fix multipart vec types

jkmassel:
When submitting multiple values for the same key using x-www-form-encoded you use `field[]` as the name – the multipart layer didn't properly encode this for `Vec` types. 

This PR should fix it.


adalpari (approved):
Tested it and it works. Kotlin code looks good, and the general code looks consistent as well.

Image

PR #996, opened 2025-10-30
Stop building for Intel Mac

crazytonyli:
Save us some building time, and we can add them back easily if we really need them.


jkmassel (changes requested):
I could probably live with this for development, but we shouldn't drop it from the published `xcframework`.

jkmassel:
@crazytonyli I think we should close this for now – #1205 got us some build speed improvements and I don't think we want to write off cross-platform support for the moment.

WDYT?

crazytonyli:
Sounds good 👍

PR #997, opened 2025-10-30
Put multipart form data into `Vec<WpMultipartFormField>`

crazytonyli:
This PR introduces a new `WpMultipartFormField` to include both the existing `fields` (plain text) and `files` (file binary). It also moves the key-value encoding part to the Rust library, so that the wrappers don't have to have code to encode `key=[foo, bar]` as `key[]=foo&key[]=bar`.

The encoding logic also handles nested arrays and objects. See the new unit tests to see how they are encoded. I believe the result follows the PHP convention.

It's easier to review this PR [with "Hiding whitepsaces" option one](https://github.com/Automattic/wordpress-rs/pull/997/files?w=1).


jkmassel (approved):
Nice improvement – thanks for the tests!

PR #998, opened 2025-10-31
Allow fetching first message of a bot conversation

jkmassel:
This is a slightly breaking change to accommodate the fact that the field names are different.


wp_api/src/wp_com/endpoint/support_bots_endpoint.rs
```diff
@@ -18,7 +19,7 @@ use wp_derive_request_builder::WpDerivedRequest;
 enum SupportBotsRequest {
     #[post(url = "/odie/chat/<bot_id>", params = &CreateBotConversationParams, output = BotConversation)]
     CreateBotConversation,
-    #[get(url = "/odie/conversations/<bot_id>", output = Vec<BotConversationSummary>)]
+    #[get(url = "/odie/conversations/<bot_id>", params = &ListBotConversationsParams, output = Vec<BotConversationSummary>)]
     GetBotConverationList,
```
oguzkocer: Looks like we have a spelling issue here, but this is unrelated to this PR.

oguzkocer (approved):

wp_api/src/wp_com/endpoint/support_bots_endpoint.rs
jkmassel: Ah yep thanks!

Fixed in 92cb83d726faaebc85102b47495131b92bf26572

wp_api/src/wp_com/support_bots.rs
```diff
@@ -107,7 +144,9 @@ pub struct UserPaidSupportPlan {
 pub struct BotConversationSummary {
     pub chat_id: u64,
     pub created_at: WpGmtDateTime,
-    pub last_message: BotMessageSummary,
+    #[serde(alias = "first_message")]
+    #[serde(alias = "last_message")]
+    pub summary_message: BotMessageSummary,
```
adalpari, line 149: ❓ Is this a description of the conversation, so we can use it as a conversation subject in the conversations list?

PR #1020, opened 2025-11-10
Improve cookies nonce api

crazytonyli:
This PR adds a new `CookiesNonceAuthenticationProvider` constructor that accepts a site URL parameter. When this constructor is used, it performs the URL discovery process internally.

### DIFF
--- a/native/swift/Sources/wordpress-api/WordPressAPI.swift
+++ b/native/swift/Sources/wordpress-api/WordPressAPI.swift
@@ -68,6 +68,55 @@ public actor WordPressAPI {
         )
     }
 
+    public init(
+        urlSession: URLSession,
+        siteUrl: String,
+        apiRootUrl: ParsedUrl,
+        username: String,
+        password: String,
+        middlewarePipeline: MiddlewarePipeline = .default,
+        appNotifier: WpAppNotifier? = nil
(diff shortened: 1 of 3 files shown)


oguzkocer (approved):
The PR looks good. If you would like to further improve the implementation (possibly in a separate PR), I do have a suggestion; which is to try to synchronize the `create_nonce_retrieval_if_needed` between multiple threads. At the moment, it's possible for multiple threads to start their own discovery process because the `nonce_retrieval` only represents whether there is a `nonce` or not, but it doesn't represent whether a discovery is already happening.

To get around this, one possibility would be to use a state like so:
```
enum NonceRetrievalState {
  NotStarted,
  InProgress,
  Success(Arc<WpRestNonceRetrieval>),
  Failed(AutoDiscoveryAttemptFailure), // allows retry logic
}
```

This ^ is probably the most flexible option, but we also have a simpler alternative which might be more than good enough for this situation and that's to use [`tokio::sync::OnceCell`](https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html) which handles the state tracking internally.

crazytonyli:
@oguzkocer Good point. I'll merge this PR for now. That definitely could be a future improvement.

PR #1030, opened 2025-11-19
Add Support Ticket encrypted log IDs

jkmassel:
Allows encrypted logs to be attached to support requests

### DIFF
--- a/native/swift/Tests/api-compatibility/wpcom/SupportTickets.swift
+++ b/native/swift/Tests/api-compatibility/wpcom/SupportTickets.swift
@@ -1,0 +1,27 @@
+import Foundation
+import Testing
+import WordPressAPI
+
+// This test ensures that existing code compiles – it's not meant to run anything
+struct SupportTicketsCompatTests {
+
+    @Test func `test ticket creation params`() async throws {
+        _ = CreateSupportTicketParams(
+            subject: "Hello World",
+            message: "Test Message",
+            application: "Test Suite",
+            wpcomSiteId: 1234,
+            tags: ["tag1", "tag2"],
+            encryptedLogIds: [UUID().uuidString]
+        )
+    }
+
+    @Test func `test ticket reply params`() async throws {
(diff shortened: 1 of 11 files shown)


native/swift/Tests/api-compatibility/wpcom/SupportTickets.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+import Testing
+import WordPressAPI
+
+// This test ensures that existing code compiles – it's not meant to run anything
+struct SupportTicketsCompatTests {
+
+    @Test func `test ticket creation params`() async throws {
```
crazytonyli, line 8: I have no idea you can use ``test ticket creation params`` as a function name. But it looks a bit strange, maybe just use regular function name convention?

crazytonyli (approved):

native/swift/Tests/api-compatibility/wpcom/SupportTickets.swift
jkmassel: Only for tests :) 

It makes the test names easy to read faster than `@Test("test ticket creation params") func testTicketCreationParams()`:

Screenshot 2025-11-19 at 4 14 47 PM

Do you still feel this way if this is only for use with Swift Testing?

crazytonyli: No, I don't really mind it. Seems like a neat trick.

crazytonyli (approved):

native/swift/Tests/api-compatibility/wpcom/SupportTickets.swift
adalpari: Just a note: this is pretty common in Kotlin :)

adalpari:
❓ Shouldn't we add the logs also to the `addMessageToSupportConversation` function as well?
I see, we added it `CreateSupportTicketParams` to but not to `AddMessageToSupportConversationParams`

PR #1042, opened 2025-11-26
Add `URLSessionTaskDelegate` parameter to WordPressAPI

crazytonyli:
The consumer can now pass a `URLSessionTaskDelegate` instance to get notified of URLSession tasks that the Rust library sends.

At the moment, only a subset of `URLSessionTaskDelegate` and `URLSessionDataDelegate` functions will be called on the new parameter.

In the implementation, I had to stop using the async functions like `await data(for: request, delegate: ...)`, and manually implement an async function using the `dataTask(...)` API. That's because the URLSession does not call the `delegate`'s task result functions (`didReceiveData`, `didReceiveResponse`, `didCompleteWithError`), when the caller uses an API that returns the HTTP response. To make sure all the functions in the new `notifyingDelegate: URLSessionTaskDelegate` parameter get called, we had to manually workaround the URLSession constraints. Additionally, the `async data(for: request, delegate: ...)` API is not available API. If we want to support the new feature across all platforms, we have to implement the async API ourselves.


Package.swift
```diff
@@ -18,7 +18,7 @@ var package = Package(
     name: "WordPressAPI",
     platforms: [
         .iOS(.v16),
-        .macOS(.v12),
+        .macOS(.v13),
```
crazytonyli, line 21: Some `URLSessionTaskDelegate` API is available on macOS 13. Bump this minimal macOS version to reduce a couple of `if #available` checks.

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -315,6 +323,95 @@ private final class RequestExecutorDelegate: NSObject, URLSessionTaskDelegate, @
...
+            )
+        #else
+            delegate?.urlSession?(
+                session,
+                task: task,
+                didSendBodyData: bytesSent,
+                totalBytesSent: totalBytesSent,
+                totalBytesExpectedToSend: totalBytesExpectedToSend
+            )
+        #endif
```
crazytonyli, line 396: You'll see many similar code below, where we have two branches that call functions on Linux and Apple platforms. The difference is the `.urlSession?(...)` part. `URLSessionTaskDelegate` API is different on Linux and Apple platforms, because having optional functions in protocols is only available on Objective-C.

jkmassel (approved):

PR #1046, opened 2025-11-27
[Bug Report] Using URLSessionTask with a delegate fails tests on Linux

crazytonyli:
This PR demonstrates a potential bug in the `URLSession` implementation on Linux.


crazytonyli:
I have reported the issue to the Swift project: https://github.com/swiftlang/swift-corelibs-foundation/issues/5294.

PR #1049, opened 2025-11-27
Support additional SSL certificate alternative names

crazytonyli:
Implement https://github.com/Automattic/wordpress-rs/pull/1043 in the Swift wrapper.

### DIFF
--- a/native/swift/Tests/wordpress-api/LoginTests.swift
+++ b/native/swift/Tests/wordpress-api/LoginTests.swift
@@ -300,7 +300,6 @@ class LoginTests {
         _ = try await self.client.findLoginUrl(forSite: "https://vanilla1.wpmt.co")
     }
 
-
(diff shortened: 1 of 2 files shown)


native/swift/Tests/wordpress-api/LoginTests.swift
```diff
@@ -11,7 +11,7 @@ import FoundationNetworking
 @Suite("Login Tests", .enabled(if: !isLinux()))
 class LoginTests {
 
-    let client = WordPressLoginClient(urlSession: .shared)
+    let client = WordPressLoginClient(urlSession: .init(configuration: .ephemeral))
```
crazytonyli, line 14: When you manually trust a certificate, it's trusted for the entire lifetime of the URLSession, which means that using the `.shared` URLSession would cause unit tests to interfere with each other.

jkmassel (approved):

PR #1050, opened 2025-11-28
Use JsonValue for post type supports field

oguzkocer:
Addresses a breaking change discovered in #1047: WordPress 6.9 returns post type `supports` as either `bool` or array, so we use `JsonValue`.


jkmassel (approved):
Thanks for taking care of this – can we validate that we have a `post_type_supports("editor")` method somewhere that just checks for that key to be present? It seems like the actual data is pretty much a black box, so we're really just checking for the key to be part of the dictionary or not.

oguzkocer:
> can we validate that we have a post_type_supports("editor") method somewhere that just checks for that key to be present? It seems like the actual data is pretty much a black box, so we're really just checking for the key to be part of the dictionary or not.

@jkmassel This should be addressed in bb2ccd9005afb99c9015e13e93f049fbb0f4177c. Let me know if this works for you or not.

Note that there is some weirdness to this design because we are basically bypassing the possible `bool` value, so I've added a comment for the assumption.

PR #1053, opened 2025-12-04
Merge the WordPressApiCache module into WordPressAPI

crazytonyli:
For context: pbArwn-7LE-p2


jkmassel (commented):
One of the goals for this was to make it so that folks could use the API layer without the caching stuff. 

Is there another way to fix the linker issues where we're consuming the library?

crazytonyli:
This PR is no longer needed, because `WordPressAPI` on the trunk branch was updated to depend on `WordPressApiCache`.

PR #1054, opened 2025-12-04
Export ParsedUrl's Eq & Hash implementations

crazytonyli:
The app can benefits from the exported APIs.


oguzkocer (approved):

jkmassel (approved):
This will make `ParsedURL` `Hashable, Comparable, Equatable`, yeah?

crazytonyli:
I don't think the Swift type will get `Comparable`, but it will get `Hashable & Equatable`.

PR #1024, opened 2025-11-16
Add `wp_mobile` service layer with observable entity pattern

oguzkocer:
This PR introduces `wp_mobile`, a mobile-friendly service layer that provides observable entities and collections with automatic cache management. The design enables reactive UIs on iOS/Android by bridging Rust database updates to platform-native observers.

## 1. `wp_mobile_cache`: Repository Pattern & Change Notifications

**Key Changes:**
- **EntityId system**: Universal identifier for cached entities across language boundaries (`DbSite + table_name + rowid`)
- **FullEntity wrapper**: Combines `EntityId` with entity data for consistent API returns
- **Repository pattern**: All repositories now return `EntityId` from upsert operations
- **Database change hooks**: `DatabaseDelegate` trait allows platforms to observe all database updates via `UpdateHook`
- **DbTable enum**: Single source of truth for table names, prevents string typos

**Design rationale**: `EntityId` provides a stable, UniFFI-compatible reference to cached data. The delegate pattern enables platform-specific reactive patterns without coupling the cache to any particular UI framework.

## 2. `wp_mobile`: Service Layer & Observable Entities

**Architecture:**

```
┌─────────────────────────────────────────────────────┐
│  Platform (iOS/Android)                             │
│  ┌──────────────────────────────────────────────┐   │
│  │  ViewModel observes Entity/Collection        │   │
│  └──────────────────────────────────────────────┘   │
└─────────────────────────────────────────────────────┘
                      ▲
                      │ UniFFI bindings
                      ▼
┌─────────────────────────────────────────────────────┐
│  wp_mobile (Rust)                                   │
│  ┌──────────────────┐  ┌──────────────────────────┐ │
│  │ Entity<T>        │  │ PostCollection<T>        │ │
│  │ - load_data()    │  │ - filter()               │ │
│  │ - is_relevant()  │  │ - fetch_page()           │ │
│  └──────────────────┘  │ - load_data()            │ │
│                        └──────────────────────────┘ │
│  ┌──────────────────────────────────────────────┐   │
│  │ PostService / SiteService                    │   │
│  │ - Create entities/collections                │   │
│  │ - Coordinate network + cache                 │   │
│  └──────────────────────────────────────────────┘   │
└─────────────────────────────────────────────────────┘
                      ▲
                      │ Uses repositories
                      ▼
┌─────────────────────────────────────────────────────┐
│  wp_mobile_cache                                    │
│  - PostRepository, SiteRepository                   │
│  - SQLite storage                                   │
[...]

PR #1055, opened 2025-12-04
Change WordPressAPI and WordPressApiCache from actor to class

crazytonyli:
Considering `WordPressAPI` and `WordPressApiCache` do not have mutable states, and all their properties conform to `Sendable`, declaring them as a `Sendable` class should have no practical impact.

The main motivation is that I don't think we need/should use `await` when accessing their properties:

```swift
let api: WordPressAPI = ...
let delegate = /* ~await~ */ api.apiClientDelegate

let apiCache: WordPressApiCache = ...
let cache = /* ~await~ */ apiCache.cache

use(delegate)
```

BTW, the publicized properties help integrate https://github.com/Automattic/wordpress-rs/pull/1024 into the app.


native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -9,18 +9,18 @@ import FoundationNetworking
...
+public final class WordPressAPI: Sendable {
 
     enum Errors: Error {
         case unableToParseResponse
     }
 
     private let apiUrlResolver: ApiUrlResolver
     let requestExecutor: SafeRequestExecutor
-    private let apiClientDelegate: WpApiClientDelegate
+    public let apiClientDelegate: WpApiClientDelegate
```
jkmassel: What's the reason for this change? I think I'd rather define whatever methods we need on `WordPressAPI` than allow others to reach into this objects implementation?

crazytonyli: This is to provide a convenient way to pass the delegate instance to `WpSelfHostedService`. I have reverted this change. There are other ways, though less convenient.

jkmassel (approved):
I _think_ this is ok – Rust seems to guarantee the thread-safety we need for the cache layer instead of needing Swift to take care of it. 

The only thing I'm a bit concerned about is that `WpApiCache` should never be accessed from the main thread – an actor guarantees this but a class doesn't.

crazytonyli:
> The only thing I'm a bit concerned about is that WpApiCache should never be accessed from the main thread

Is that why the `performMigrations` was declared as an `async` function, even though there is no `await` call in it?

jkmassel:
> Is that why the performMigrations was declared as an async function, even though there is no await call in it?

IIRC it was because I wasn't sure how the async story was going to work for sqlite yet. We ended up going with mutex-based thread safety, so I don't think we need it anymore.

It also occurred to me that we could use runtime debug assertions to enforce "don't call this from the main thread" as needed.

crazytonyli:
There are some other considerations when using the cache in the cpt prototype in the app. I'll include when/how to call `performMigrations` on the app side. I'll leave the main thread's check assertions out for now, until the prototype is completed, so we have concrete code to discuss on how the cache instance is used in the app.

PR #1056, opened 2025-12-04
Export a function to return the API root URL

crazytonyli:
The iOS apps can benefit from this change.

### DIFF
--- a/wp_api/src/request/endpoint.rs
+++ b/wp_api/src/request/endpoint.rs
@@ -132,6 +132,10 @@ impl WpOrgSiteApiUrlResolver {
     pub fn new(api_root_url: Arc<ParsedUrl>) -> Self {
         Self { api_root_url }
     }
+
+    fn api_root_url(&self) -> Arc<ParsedUrl> {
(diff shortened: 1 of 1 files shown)


oguzkocer (approved):

wp_api/src/request/endpoint.rs
```diff
@@ -132,6 +132,10 @@ impl WpOrgSiteApiUrlResolver {
     pub fn new(api_root_url: Arc<ParsedUrl>) -> Self {
         Self { api_root_url }
     }
+
```
jkmassel, line 135: I think this is fine, but can you give some context on why it's needed? :)

crazytonyli: Just providing a convenient way to get the API root URL, to omit an `apiRootURL: URL` argument from a function in the iOS app: https://github.com/wordpress-mobile/WordPress-iOS/blob/5b46dda65f16b78c72a7c4e468b6972ecad61378/Modules/Sources/WordPressCore/WordPressClient.swift#L11-L18

PR #1057, opened 2025-12-04
Swift support for wp-mobile

jkmassel:


native/swift/Sources/wordpress-api-cache/DatabaseChangeNotifier.swift
```diff
@@ -0,0 +1,294 @@
...
+            object: nil,
+            queue: nil
+        ) { notification in
+            guard let updateHook = notification.userInfo?["updateHook"] as? UpdateHook else { return }
+            callback(updateHook)
+        }
+
+        tokenStore.add(token, for: object.entityId) {
+            NotificationCenter.default.removeObserver($0)
+        }
```
crazytonyli: I can see this hand-writen observer implementation is necessary for Linux, where Combine is not available. But for Apple platforms, maybe we should use Combine.Publisher, which can turn into AsyncSequence (thus will work well with the async code)?

native/swift/Sources/wordpress-api-cache/ObservableCollection.swift
```diff
@@ -0,0 +1,29 @@
...
+    var handle: UInt64 { get }
+
+    /// The method used to load data from the cache. Usually not implemented directly – if `Element` is correct this can be mapped automatically
+    func loadData() async throws -> [Element]
+}
+
+extension AllAnyPostWithEditContextCollection: ObservableCollection {
+    public typealias Element = FullEntityAnyPostWithEditContext
+
+    public var tableName: String { "postsEditContext" }
```
crazytonyli, line 21: I believe we should use the `DbTable` type here, instead of String?

native/swift/Sources/wordpress-api-cache/DatabaseChangeNotifier.swift
```diff
@@ -0,0 +1,294 @@
...
+                try await callback(updateHook)
+            }
+        }
+
+        tokenStore.add(token, for: collection.handle) {
+            NotificationCenter.default.removeObserver($0)
+        }
+    }
+
+    @available(macOS 15.0, *)
```
crazytonyli, line 124: I believe we should prefer to use the Combine solution on all platforms, instead of just macOS.

native/swift/Sources/wordpress-api-cache/DatabaseChangeNotifier.swift
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+}
+
+public struct UpdateHookSequence: AsyncSequence, Sendable {
+
+    let name: Notification.Name
+
+    public struct AsyncIterator : AsyncIteratorProtocol {
+        var publisher: AsyncCompactMapSequence<AsyncPublisher<NotificationCenter.Publisher>, UpdateHook>.Iterator
```
crazytonyli, line 278: I don't think you need `UpdateHookSequence`, it's just a wrapper around this async publisher?

If you want to erase the type, you can use `any AsyncSequence<UpdateHook, Never>` as the return value.

If [the `Failure` associated type is introduced on iOS 18 and macOS 15](https://developer.apple.com/documentation/swift/asyncsequence/failure) causes some compilation issues, maybe we simply return `AnyPublisher<UpdateHook, Never>`, and let the client call the `.values`?

native/swift/Sources/wordpress-api-cache/DatabaseChangeNotifier.swift
```diff
@@ -0,0 +1,294 @@
...
+///
+/// - Note: Observers are automatically replaced if you start observing the same entity/collection multiple times.
+///         The previous observer will be cleaned up automatically.
+public final class DatabaseChangeNotifier: DatabaseDelegate, Sendable {
+
+    /// The shared singleton instance of the database change notifier.
+    public static let shared = DatabaseChangeNotifier()
+
+    /// Internal storage for notification observer tokens.
+    private let tokenStore = TokenStore()
```
crazytonyli, line 40: I don't think we _need_ to implement the observer pattern ourselves? On Apple platforms, we can use Combine. On Linux, _I think_ we can pass the `callback` closure to `NSNotificationCenter`?

native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -138,6 +143,22 @@ public actor WordPressAPI {
         self.requestExecutor = executor
     }
 
+    public func asSelfHostedService() throws -> WpSelfHostedService {
+        let cache = try WpApiCache(path: nil)
+        _ = try cache.performMigrations()
+        cache.startListeningForUpdates(delegate: DatabaseChangeNotifier.shared)
```
crazytonyli, line 149: _I think_ there is a potential isuse here. When you call `startListeningForUpdates` multiple times, Rust gets multiple instances of "update delegate", which is one single instance: `DatabaseChangeNotifier.shared`. That means, potentially, the singleton receives multiple calls upon one database change.

```swift
let api: WordPressAPI = ...
// There are 5 `WpApiCache` instances created, most importantly, 5 delegates are added to the Rust
let services = (1...5).map { api.asSelfHostedService() }

let observer = Task {
  for await _ in DatabaseChangeNotifier.shared.startObsever(...) {
    print("Update received")
  }
}

// Later, a post is added to the database. How many "Update received" are expected to be printed?
services[0].fetch_post(1)
```

native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -138,6 +143,22 @@ public actor WordPressAPI {
         self.requestExecutor = executor
     }
 
+    public func asSelfHostedService() throws -> WpSelfHostedService {
+        let cache = try WpApiCache(path: nil)
+        _ = try cache.performMigrations()
```
crazytonyli, line 148: Also, based on a hunch (we can probably write some unit tests to verify this), we should not call `performMigrations` on the same database file. I don't think that the migration implementation is supposed to handle multiple `performMigrations` calls at the same time.

I'm aware that this exact code uses an in-memory database, so the above won't be an issue. But we probably should add TODO or FIXME comment to say we are not supposed to have multiple `WpApiCache` instances and call `performMigrations` multiple times.

crazytonyli (approved):
Approve to unblock, but we probably should discuss and potentially address some of the comments.

PR #1072, opened 2025-12-19
`WpEndpointArgs` now accepts all JSON types

crazytonyli:
I found a plugin ([The Event Tickets plugin](https://wordpress.org/plugins/event-tickets/)) has an unexpected `args` in its `wp-json` response, which causes the `WpApiDetails` parsing failure. That means any site with that plugin won't be able to sign in with application passwords from the app.

### DIFF
--- a/test-data/api-details/test-case-05.json
+++ b/test-data/api-details/test-case-05.json
@@ -1,0 +1,78 @@
+{
+  "name": "Example WordPress Site",
+  "description": "",
+  "url": "https://example-site.wordpress.test",
+  "home": "https://example-site.wordpress.test",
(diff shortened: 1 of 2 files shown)


jkmassel (approved):
Nice fix, and thanks for the test.

I wonder if `WpEndpoint.args` should just be `JsonValue` and consumers can work with it however they want?

crazytonyli:
I don't know how to answer that 😄 , as I'm not sure how the `WpEndpointArg` type is intended to be used, which appears to be the only thing affected here. The current PR maintains the existing API, with minimal changes to how JSON is parsed.

PR #1085, opened 2026-01-07
Fix a few parsing issues when parsing custom post types

crazytonyli:
> [!Note]
> It should be easier to review this PR commit by commit.

The `PostMeta` can't be parsed when the meta does not contain "footnotes".

The `title` and `password` properties in `SparseAnyPost` are changed from non-optional to optional. Some custom post types (such as Jetpack Social Notes) do not include those properties.

To change the `title` declaration, the `WpContextualField` macro is updated to accept a new `option` marker attribute. When present, the `title` in the generated `AnyPostWithXxxContext` structs is declared as `Option`. Here is a snippet of the actual diff of macro-expanded code between the trunk and this PR branch:

```diff
--- wp_api_expanded_trunk.rs	2026-01-07 15:28:39
+++ wp_api_expanded_new.rs	2026-01-07 15:28:49
@@ -241994,6 +241994,16 @@
     #[automatically_derived]
     impl ::core::fmt::Debug for ParsedUrl {
         #[inline]
@@ -297777,6 +297777,7 @@
         #[WpContext(edit, embed, view)]
         pub post_type: Option<String>,
         #[WpContext(edit)]
+        #[WpContextualOption]
         pub password: Option<String>,
         #[WpContext(edit)]
         #[WpContextualOption]
@@ -297785,7 +297786,7 @@
         #[WpContextualOption]
         pub generated_slug: Option<String>,
         #[WpContext(edit, embed, view)]
-        #[WpContextualField]
+        #[WpContextualField(option)]
         pub title: Option<SparsePostTitle>,
         #[WpContext(edit, view)]
         #[WpContextualField]
@@ -299849,10 +299850,10 @@
         pub status: PostStatus,
         #[serde(rename = "type")]
         pub post_type: String,
-        pub password: String,
+        pub password: Option<String>,
         pub permalink_template: Option<String>,
         pub generated_slug: Option<String>,
-        pub title: PostTitleWithEditContext,
+        pub title: Option<PostTitleWithEditContext>,
         pub content: PostContentWithEditContext,
         pub author: Option<UserId>,
         pub excerpt: Option<SparsePostExcerpt>,
```

### DIFF
--- a/wp_mobile_cache/src/repository/posts.rs
+++ b/wp_mobile_cache/src/repository/posts.rs
@@ -476,10 +476,10 @@ impl PostContext for EditContext {
             password: row.get_column(Password)?,
             permalink_template: row.get_column(PermalinkTemplate)?,
             generated_slug: row.get_column(GeneratedSlug)?,
-            title: PostTitleWithEditContext {
+            title: Some(PostTitleWithEditContext {
                 raw: row.get_column(TitleRaw)?,
                 rendered: row.get_column(TitleRendered)?,
-            },
+            }),
             content: PostContentWithEditContext {
                 raw: row.get_column(ContentRaw)?,
                 rendered: row.get_column(ContentRendered)?,
@@ -563,9 +563,9 @@ impl PostContext for ViewContext {
             slug: row.get_column(Slug)?,
             status: parse_enum(row, Status)?,
             post_type: row.get_column(PostType)?,
-            title: PostTitleWithViewContext {
+            title: Some(PostTitleWithViewContext {
                 rendered: row.get_column(TitleRendered)?,
-            },
+            }),
             content: PostContentWithViewContext {
                 rendered: row.get_column(ContentRendered)?,
                 protected: row.get_column(ContentProtected)?,
@@ -639,9 +639,9 @@ impl PostContext for EmbedContext {
             link: row.get_column(Link)?,
             slug: row.get_column(Slug)?,
             post_type: row.get_column(PostType)?,
-            title: PostTitleWithEmbedContext {
+            title: Some(PostTitleWithEmbedContext {
                 rendered: row.get_column(TitleRendered)?,
-            },
+            }),
             author: get_optional_id(row, Author)?,
             excerpt: {
                 let excerpt_rendered: Option<String> = row.get_column(ExcerptRendered)?;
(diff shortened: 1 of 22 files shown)


wp_mobile_cache/src/test_fixtures/posts.rs
```diff
@@ -177,7 +177,7 @@ fn create_minimal_post() -> AnyPostWithEditContext {
         slug: "minimal-post".to_string(),
         status: PostStatus::Publish,
         post_type: "post".to_string(),
-        password: "".to_string(),
+        password: Some("".to_string()),
```
jkmassel, line 192: We should probably have a test fixture for `None`?

jkmassel (approved):
Generally fine, but one testing concern.

wp_mobile_cache/src/repository/posts.rs
```diff
@@ -1191,7 +1192,25 @@ mod tests {
...
+        }
+        if original_post.title.is_none() {
+            assert_eq!(
+                retrieved_post.title.map(|t| t.rendered),
+                Some("".to_string())
+            );
+
+            retrieved_post.title = None;
+        }
+        assert_eq!(retrieved_post, original_post);
```
crazytonyli: There is a mismatch of the `password` and `title_rendered` declarations between the Rust type and the database schema. In Rust, they are `Option`, but in the database, they are `NOT NULL`. When the Rust property value is None, an empty string is stored in the database (see the `:title_rendered` and `:password` changes above).

That means the `retrieved_post` and `original_post` do not match exactly. The `original_post.password` is nil, whereas `retrieved_post.password` is `""`. But they can be treated as the "not set". That's why the if statements above exist: they manually verify the two `Option` Rust properties, but `NON NULL` in the database.

I'm aware that this is a hacky solution. A proper solution would be to change the database schema to match the Rust code's declaration. To do that, we'll need to add an SQL migration file. However, since SQLite does not support altering table column constraints directly, we'll need to recreate the tables entirely.

The migration process could involve creating a new temporary table, copying the existing table's data into it, deleting the existing table, and finally renaming the temporary table to match the existing table. This process needs to be repeated on three post tables, one for each context. IMO, that feels too much.

Alternatively, we can recreate the tables without migrating the existing data. The post data will be re-fetched the next time the user views the post list. This would be a simplified solution, with the cost of losing, though harmless, the already fetched post data.

@jkmassel what do you think?

crazytonyli: I have created a follow-up PR which implements a database migration: https://github.com/Automattic/wordpress-rs/pull/1086

jkmassel: Let's just change the code – this doesn't run anywhere in production, so I'd prefer not to make a DB migration that benefits nobody?

PR #1086, opened 2026-01-07
Remove the 'NOT NULL' constraint on password and title columns

crazytonyli:
> [!Note]
> This PR builds on top of https://github.com/Automattic/wordpress-rs/pull/1085. If the approach looks good, I can merge this PR so that all the changes can be reviewed in one single PR.

This PR is a solution for https://github.com/Automattic/wordpress-rs/pull/1085#discussion_r2667706289.

This PR removes the `NOT NULL` constraint on `password` and `title_rendered` columns, so that the database schema matches the Rust property declaration.

The new migration SQL file does:
1. Copy the existing `password` and `title_rendered` data to a temporary table.
2. Re-create the `password` and `title_rendered` columns without the `NOT NULL` constraint. This operation removes table data, which is why we copied the existing data in the step above.
3. Update the new columns with the copied data.
4. Delete the temporary table.

These same migration steps are performed on all three `posts` tables.


wp_mobile_cache/docs/ADDING_NEW_ENTITIES.md
```diff
@@ -62,7 +62,7 @@ When adding a new entity type (e.g., implementing cache support for comments):
 
 ## Key Design Decisions
 
-**Column enum ordering**: Column indexes in the enum must match the exact order in SQL `CREATE TABLE` statements and all `SELECT *` queries. Add PRAGMA-based integration tests to verify this (see `test_post_edit_context_column_enum_matches_schema` in `repository/posts.rs` for reference).
```
crazytonyli, line 65: Step 2, which deletes and re-creates the columns, changes the column order, so the hard-coded column indexes in `PostEditContextColumn` are no longer valid. We can change the hard-coded values to match the new order. But I thought it might be more flexible to retrieve row data by column name rather than column index, which is why I refactored the code to remove `ColumnIndex`. I'm open to keeping the `ColumnIndex` though. Let me know what you think. @jkmassel

crazytonyli: > We can change the hard-coded values to match the new order.

I have implemented this in https://github.com/Automattic/wordpress-rs/pull/1089, which has a much smaller diff. We can compare these two and decide which one to merge.

PR #1092, opened 2026-01-12
Add WP.com OAuth support

jkmassel:
- Add WordPress.com OAuth2 authentication support in Rust
- Updates the Swift demo app to support WP.com sites

To test: Try the demo app


native/swift/Sources/wordpress-api/WPComApiClient.swift
```diff
@@ -1,7 +1,32 @@
 import Foundation
 import WordPressAPIInternal
 
-public class WPComApiClient {
+public final class WPComApiClient: Sendable {
+
+    public struct OAuth2 {
+        public static func build_token_request_url(
```
crazytonyli: Change to `buildTokenRequestURL`?

wp_api/src/wp_com/oauth2.rs
```diff
@@ -25,10 +46,136 @@ pub struct TokenValidationResponse {
...
+/// * `blog` - Optional blog ID to request access to a specific blog
+///
+/// # Returns
+///
+/// A `ParsedUrl` containing the authorization URL to open in the user's browser.
+#[uniffi::export]
+pub fn build_token_request_url(
+    client_id: &str,
+    redirect_uri: &str,
+    response_type: &str,
```
crazytonyli: Considering that the `extract_code_from_authorization_url` function is hard-coded to parse the 'code', should we remove the `responseType` argument and hard-code the authorize URL to use `code`?

wp_api/src/wp_com/oauth2.rs
```diff
@@ -25,10 +46,136 @@ pub struct TokenValidationResponse {
...
+/// * `client_secret` - Your application's client secret
+/// * `code` - The authorization code received from the OAuth2 callback
+/// * `grant_type` - Must be `"authorization_code"` for the Authorization Code flow
+/// * `redirect_uri` - Must match the redirect URI used in the authorization request
+#[derive(Debug, PartialEq, Eq, Serialize, uniffi::Record)]
+pub struct TokenRequestParameters {
+    pub client_id: u64,
+    pub client_secret: String,
+    pub code: String,
+    pub grant_type: String,
```
crazytonyli, line 76: Similarly, should this be hard-coded as `authorization_code`?

native/swift/Sources/wordpress-api/WPComApiClient.swift
```diff
@@ -1,7 +1,32 @@
...
+            clientId: UInt64,
+            redirectUri: URL,
+            responseType: String = "code",
+            scope: [String],
+            state: String = UUID().uuidString,
+            blog: UInt64? = nil
+        ) -> URL {
+            buildTokenRequestUrl(
+                clientId: String(clientId),
```
crazytonyli: The "client id" is declared as a string in Rust, but as an integer in Swift. Maybe we should keep them consistent?

wp_api/src/wp_com/oauth2.rs
```diff
@@ -1,7 +1,28 @@
...
+use url::Url;
+use wp_serde_helper::deserialize_u64_or_none_with_zero_as_none_from_string;
 use wp_serde_helper::deserialize_u64_or_string;
 
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum AuthorizationCodeExtractionError {
+    #[error("Invalid URL: {reason}")]
+    InvalidUrl { reason: String },
+    #[error("Missing 'code' parameter in authorization URL")]
+    MissingCode,
```
crazytonyli, line 15: The callback URL may contain failure messages, like `error=login_denied`. What do you think about adding a case here to include the returned error?

native/swift/Sources/wordpress-api/WPComApiClient.swift
jkmassel: Got Rust all up in my Swift...fixed in 285206ef76c56b0123b33fe57a580bf903e9eaf0

native/swift/Sources/wordpress-api/WPComApiClient.swift
jkmassel: For WP.com it should always be an integer, so even though the spec says it's a string we can be a little more specific here, I think.

I've aligned it around u64 – it'd be pretty easy to switch it back without it being a breaking change though.

wp_api/src/wp_com/oauth2.rs
jkmassel: Done in 206f2b8

wp_api/src/wp_com/oauth2.rs
```diff
@@ -161,14 +169,33 @@ pub fn build_token_request_url(
...
-) -> Result<String, AuthorizationCodeExtractionError> {
+) -> Result<AuthorizationCodeExtractionResult, AuthorizationCodeExtractionError> {
     let url = Url::parse(&response)?;
+
+    if let Some(error_code) = value_from_query_pairs("error", &url) {
+        if let Some(error_description) = value_from_query_pairs("error_description", &url) {
+            return Err(AuthorizationCodeExtractionError::Error {
+                message: error_description.clone(),
+                code: error_code.clone(),
+            });
```
crazytonyli: I think it'd be okay if `error_description` is nil. In which case, the parsing will be more relaxed.

crazytonyli (approved):

PR #1097, opened 2026-01-13
Fix `perform_migration` race condition

crazytonyli:
It's possible to have multiple `WpApiCache` instances that use the same database file. Thus, it's possible that the `WpApiCache.perform_migration` gets called multiple times concurrently on the same database file. The expected result in that scenario is:
1. All `perform_migration` should return a successful result.
2. The actual database migration SQL should only run once, and the end result of the database should be as if only one `perform_migration` is called.

### DIFF
--- a/wp_mobile_cache/src/lib.rs
+++ b/wp_mobile_cache/src/lib.rs
@@ -613,4 +613,89 @@ mod tests {
         assert!(!does_sqlite_support_returning("invalid"));
         assert!(!does_sqlite_support_returning(""));
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 10)]
+    async fn test_concurrent_migrations_race_condition() {
+        let db_path = "/tmp/wp_cache_race_test.db";
+        let _ = std::fs::remove_file(db_path);
+
+        let num_tasks = 10;
+        let mut tasks = vec![];
+
+        for task_id in 0..num_tasks {
+            let task = tokio::spawn(async move {
+                let cache = WpApiCache::new(Some(db_path.to_string())).unwrap();
+                let result = cache.perform_migrations();
+
+                (task_id, result)
+            });
+
+            tasks.push(task);
+        }
+
+        let results: Vec<(usize, Result<i64, SqliteDbError>)> = futures::future::join_all(tasks)
+            .await
+            .into_iter()
+            .map(|r| r.unwrap())
+            .collect();
+
+        let mut error_count = 0;
+
+        for (task_id, result) in &results {
+            match result {
+                Ok(migrations_run) => {
+                    println!(
+                        "Task {} succeeded, ran {} migrations",
+                        task_id, migrations_run
+                    );
+                }
+                Err(e) => {
+                    println!("Task {} failed with error: {}", task_id, e);
+                    error_count += 1;
+                }
+            }
+        }
+
+        let migrations_run_counts: Vec<i64> = results
+            .iter()
+            .filter_map(|(_, result)| result.as_ref().ok())
+            .copied()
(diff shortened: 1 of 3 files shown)


wp_mobile_cache/src/lib.rs
```diff
@@ -440,6 +440,21 @@ impl<'a> MigrationManager<'a> {
     }
 
     pub fn perform_migrations(&mut self) -> SqliteResult<i64> {
+        self.connection.execute("BEGIN EXCLUSIVE", ())?;
```
crazytonyli, line 443: There is [a dedicated transaction API](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.transaction_with_behavior). But using that API changes lots of the `MigrationManager` code (mainly switching from using `Connection` to `Transaction`). Maybe we can do that later.

wp_mobile_cache/src/lib.rs
```diff
@@ -613,4 +628,89 @@ mod tests {
         assert!(!does_sqlite_support_returning("invalid"));
         assert!(!does_sqlite_support_returning(""));
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 10)]
+    async fn test_concurrent_migrations_race_condition() {
```
jkmassel, line 633: Would this test fail prior to the change?

jkmassel (commented):
What are the circumstances under which this might happen?

It seems like a programmer error to end up in this state and I'm not certain about the fix.

In other areas, we've used a mutex to protect concurrent access to shared object – could we use that instead of a transaction?

wp_mobile_cache/src/lib.rs
crazytonyli: Yes, it would. You can checkout the first commit of this PR branch and run the test locally.

crazytonyli:
> What are the circumstances under which this might happen?
> It seems like a programmer error

I was concerned about a programming error. I don't think anyone would intentionally create multiple `WpApiCache` instances that use the same database file and then call the `perform_migration` function. However, that does not mean the scenario won't happen at runtime.

If the API design, at the Rust layer, is that the user is allowed to create multiple `WpApiCache` instances that use the same database file, we'd have to deal with the potential issue of the `perform_migration` function being called at the same time.

> In other areas, we've used a mutex to protect concurrent access to shared object

In this particular case, a global mutex would be needed since we are preventing a race condition between multiple `WpApiCache` instances here. However, that's what this PR does, though. It's just that the mutex is not in Rust, it's the `BEGIN EXCLUSIVE` transaction from the sqlite library.

crazytonyli:
> However, that does not mean the scenario won't happen at runtime.

To expand on this. Let's say the app is coded in a way that only one instance of `WordPressAPI`, `WpSelfHostedService`, and `WpApiCache` is created per site.

```swift
func createSiteApiThings(blogId: ...) {
  if hasNotCreated {
    // create WordPressAPI
    // create WpSelfHostedService
    // create WpApiCache that uses `/path/to/app.sqlite` and call performMigration
  }
  ...
}
```

It's not possible to prevent this `createSiteApiThings` being called by various places of the app, with _different_ blogId parameter.

jkmassel (approved):
> In this particular case, a global mutex would be needed since we are preventing a race condition between multiple WpApiCache instances here. However, that's what this PR does, though. It's just that the mutex is not in Rust, it's the BEGIN EXCLUSIVE transaction from the sqlite library.

I'm aware of this, but a little nervous delegating this to another system (which can have different behaviour based on WAL mode, among other things). The fact that we have a test should catch if there's a regression though, so I think this is good to go.

PR #1099, opened 2026-01-14
Re-implement observing WpApiCache updates

crazytonyli:
I have raised a few concerns about the `DatabaseChangeNotifier` in https://github.com/Automattic/wordpress-rs/pull/1057. This PR removes `DatabaseChangeNotifier` and re-implements a way to observe database changes.

There are three new APIs:
1. `addDatabaseUpdatesObserver`, which uses a closure. This is similar to the `DatabaseChangeNotifier.startObserving(callback:)` API, but `DatabaseChangeNotifier` almost implemented its own version of `NotificationCenter`. The `addDatabaseUpdatesObserver` just reuses `NotificationCenter`.
2. `databaseUpdates`, which returns an `AsyncSequence` instance. However, this API is only available on iOS 18, due to syntax issues in using the `Failure` generic in `AsyncSequence`.
3. `databaseUpdatesPublisher`, which returns a Combine publisher. The app can easily call `.values` to convert it to an `AsyncSequence` and use it in an async context, which is exactly what it will do.

Also, internally, there is no longer a singleton like `DatabaseChangeNotifier.shared`, which prevents observers from receiving updates that were sent by other `WordPressApiCache` instances. There are new comprehensive unit tests that verify this behaviour.


jkmassel (approved):
Conceptually I think this is fine, pending test fixes

crazytonyli:
All the changes above fix the test setup in unit tests: using `Task` is too unreliable, so the tests now use `Combine` publishers.

PR #1100, opened 2026-01-15
Add integration tests for wp_mobile_cache

crazytonyli:
Add a couple of simple integration tests as a start.


jkmassel (approved):
This works fine if I run `cargo test -p wp_mobile_cache_integration_tests` but I'd expect that I could invoke it via `make` and that it would run in CI as well.

crazytonyli:
Good catch! I missed that. It's now part of CI.

PR #1101, opened 2026-01-15
Reproduce wp mobile cache issue

crazytonyli:
> [!Note]
> This PR will be merged after https://github.com/Automattic/wordpress-rs/pull/1100.

Relates to p1768466053646389-slack-C06S1QS55K9.

I think the long code comment in the new test should be enough to explain the issue this PR tries to fix. You can checkout the second commit (the one that adds the new integration tests) and run it locally to see the result. The third commit fixes the issue.


jkmassel (approved):
This looks ok – a few things before merging:

1. Test failures
2. We can modify the existing migration instead of making a new one (because this isn't used in production yet)

PR #1102, opened 2026-01-15
Make the WordPressLoginClient Swift wrapper cancellable

crazytonyli:
The `WordPressLoginClient` APIs now support native Swift cancellation.


jkmassel (approved):
I finally had a chance to dig into this – looks good, thanks!!

PR #1111, opened 2026-01-20
Swift: generate mutable records

crazytonyli:
Change struct properties from `let` to `var`. Sometimes it's easier, syntactically, to mutate an existing `struct` instance rather than use `.init()` to copy all the properties.

### DIFF
--- a/wp_api/uniffi.toml
+++ b/wp_api/uniffi.toml
@@ -20,6 +20,6 @@ from_custom = "Int64({}.timeIntervalSince1970)"
 ffi_module_name = "libwordpressFFI"
 ffi_module_filename = "wp_api_uniffi"
 generate_module_map = false
-generate_immutable_records = true
+generate_immutable_records = false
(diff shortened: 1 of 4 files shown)


jkmassel:
> Change struct properties from let to var. Sometimes it's easier, syntactically, to mutate an existing struct instance rather than use .init() to copy all the properties.

Where are we doing this? In general, the philosophy for this project is that external users should copy the data out of these types as soon as possible and they should remain immutable.

crazytonyli:
I'm using `PostListFilter` in the custom post list to apply status filters. The struct is stored as a `@State` property in SwiftUI view. The post list currently only supports filtering by status, but we may want to support sorting, or potentially other things. I thought it'd be easier to store `PostListFilter` directory, rather than individual properties (`status`, `sortBy`, `sortOrder`, etc).

crazytonyli:
Closing this PR, as I have changed the app to work without this PR.

PR #1117, opened 2026-01-22
Fix parsing Newspack Newsletter post type

crazytonyli:
The `has_archive` property is supposed to be a boolean, but the plugin returns a string (see the unit test file).

I think the solution here (replacing `bool` with `BoolOrString`) is not great. But since we don't use the `has_archive` property, it has no practical impact.

### DIFF
--- a/wp_api/src/lib.rs
+++ b/wp_api/src/lib.rs
@@ -190,7 +190,7 @@ impl TryFrom<BoolOrString> for WpResponseString {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, uniffi::Enum)]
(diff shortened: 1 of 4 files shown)


jkmassel (approved):
This is actually correct behaviour – from `register_post_type` [documentation](https://developer.wordpress.org/reference/functions/register_post_type/):

> has_archive bool|string
Whether there should be post type archives, or if a string, the archive slug to use. Will generate the proper rewrite rules if $rewrite is enabled. Default false.

It might be worth using something more descriptive than `BoolOrString` (ie – something that communicates "yes this does have an archive" regardless of whether it's `true` or a slug is provided (which is also true-ish).

crazytonyli:
Hmm, I did not know that. In that case, `BoolOrString` sounds better than just `bool`. 😄 

I'll keep the current code for now, since the property is not used. We can improve it when we need to use it.

PR #1119, opened 2026-01-26
CMM-1177/CMM-1178 create the top post and referrers endpoints

adalpari:
## Summary                                                                                                                                                                      
  - Adds new WordPress.com REST API endpoints for fetching site statistics:                                                                                                       
    - **Top Posts** (`/rest/v1.1/sites/<site_id>/stats/top-posts`) - Returns most viewed posts/pages                                                                              
    - **Referrers** (`/rest/v1.1/sites/<site_id>/stats/referrers`) - Returns traffic referrer sources                                                                             
                                                                                                                                                                                  
  ## Implementation Details                                                                                                                                                       
                                                                                                                                                                                  
  ### New Files                                                                                                                                                                   
                                                                                                                                                                                  
  **Top Posts Endpoint:**                                                                                                                                                         
  - `wp_api/src/wp_com/stats_top_posts.rs` - Types, parameters, response structures, and helper functions                                                                         
  - `wp_api/src/wp_com/endpoint/stats_top_posts_endpoint.rs` - Endpoint definition                                                                                                
  - `wp_api/tests/wpcom/stats_top_posts/top-posts-01.json` - Test data fixture                                                                                                    
                                                                                                                                                                                  
  **Referrers Endpoint:**                                                                                                                                                         
  - `wp_api/src/wp_com/stats_referrers.rs` - Types, parameters, response structures, and helper functions                                                                         
  - `wp_api/src/wp_com/endpoint/stats_referrers_endpoint.rs` - Endpoint definition                                                                                                
  - `wp_api/tests/wpcom/stats_referrers/referrers-01.json` - Test data fixture                                                                                                    
                                                                                                                                                                                  
  ### Modified Files                                                                                                                                                              
  - `wp_api/src/wp_com/mod.rs` - Module registration                                                                                                                              
  - `wp_api/src/wp_com/endpoint.rs` - Endpoint module registration                                                                                                                
  - `wp_api/src/wp_com/client.rs` - Request builder and executor integration                                                                                                      
                                                                                                                                                                                  
  ### API Parameters                                                                                                                                                              
                                                                                                                                                                                  
  **Top Posts:**                                                                                                                                                                  
                                                                                                                                                                                  
  | Parameter    | Type                  | Description                                   |                                                                                        
  | ------------ | --------------------- | --------------------------------------------- |                                                                                        
  | `period`     | `StatsTopPostsPeriod` | Time period grouping (day, week, month, year) |                                                                                        
  | `start_date` | `String`              | Start date in YYYY-MM-DD format               |                                                                                        
  | `date`       | `String`              | Query date in YYYY-MM-DD format               |                                                                                        
  | `max`        | `u32`                 | Maximum number of top posts to return         |                                                                                        
                                                                                                                                                                                  
  **Referrers:**                                                                                                                                                                  
                                                                                                                                                                                  
  | Parameter    | Type                   | Description                                   |                                                                                       
  | ------------ | ---------------------- | --------------------------------------------- |                                                                                       
  | `period`     | `StatsReferrersPeriod` | Time period grouping (day, week, month, year) |                                                                                       
  | `date`       | `String`               | Query date in YYYY-MM-DD format               |                                                                                       
  | `start_date` | `String`               | Start date in YYYY-MM-DD format               |                                                                                       
  | `max`        | `u32`                  | Maximum number of referrers to return         |                                                                                       
                                                                                                                                                                                  
  ### Helper Functions                                                                                                                                                            
                                                                                                                                                                                  
  **Top Posts:**                                                                                                                                                                  
  - `get_stats_top_posts_all_post_views()` - Returns all post views from all days                                                                                                 
  - `get_stats_top_posts_for_date()` - Returns post views for a specific date                                                                                                     
  - `get_stats_top_posts_total_views()` - Returns total views across all days                                                                                                     
                                                                                                                                                                                  
  **Referrers:**                                                                                                                                                                  
  - `get_stats_referrers_all_groups()` - Returns all referrer groups from all days                                                                                                
  - `get_stats_referrers_for_date()` - Returns referrer data for a specific date                                                                                                  
  - `get_stats_referrers_total_views()` - Returns total views across all days                                                                                                     
                                                                                                                                                                                  
  ### Notable Implementation Details                                                                                                                                              
                                                                                                                                                                                  
  The referrers endpoint has a polymorphic `results` field that can be either:                                                                                                    
  - A simple views object: `{"views": 12}`                                                                                                                                        
  - A detailed referrer array (e.g., for Search Engines): `[{"name": "Google Search", ...}]`                                                                                      
                                                                                                                                                                                  
  This is handled using `#[serde(untagged)]` enum (`StatsReferrersResults`).


wp_api/src/wp_com/stats_referrers.rs
```diff
@@ -0,0 +1,377 @@
...
+impl AppendUrlQueryPairs for StatsReferrersParams {
+    fn append_query_pairs(&self, query_pairs_mut: &mut QueryPairs) {
+        query_pairs_mut
+            .append_option_query_value_pair("period", self.period.as_ref())
+            .append_option_query_value_pair("date", self.date.as_ref())
+            .append_option_query_value_pair("start_date", self.start_date.as_ref())
+            .append_option_query_value_pair("max", self.max.as_ref())
+            .append_option_query_value_pair("num", self.num.as_ref())
+            .append_query_value_pair("summarize", &1u32)
+            .append_query_value_pair("skip_archives", &1u32);
```
jkmassel: What's the story with these?

wp_api/src/wp_com/stats_referrers.rs
```diff
@@ -0,0 +1,377 @@
...
+    Week,
+    Month,
+    Year,
+}
+
+impl_as_query_value_from_to_string!(StatsReferrersPeriod);
+
+/// Parameters for the stats referrers endpoint.
+#[derive(Debug, Default, PartialEq, Eq, uniffi::Record)]
+pub struct StatsReferrersParams {
```
jkmassel, line 40: This needs to take a `locale` param – that's used to translate the response's strings (for example "Search Engines").

It should probably be an enum that matches WP.com locales - @crazytonyli wdyt about using `WPComLanguage` from #1011?

wp_api/src/wp_com/stats_top_posts.rs
```diff
@@ -0,0 +1,282 @@
...
+        params.append_query_pairs(&mut query_pairs);
+
+        assert_eq!(
+            query_pairs.finish().as_str(),
+            "https://public-api.wordpress.com/rest/v1.1/sites/1234/stats/top-posts?period=week&date=2026-01-19&summarize=1&skip_archives=1"
+        );
+    }
+
+    #[test]
+    fn test_stats_top_posts_response_deserialization() {
```
jkmassel: Anything that uses the JSON tests should be parameterized so that we can easily add more test files to it. This comment applies to all cases like it in this PR.

wp_api/src/wp_com/stats_referrers.rs
```diff
@@ -0,0 +1,377 @@
...
+    pub site_id: Option<u64>,
+    /// The blog title.
+    pub blog_title: Option<String>,
+    /// Whether currently following.
+    pub is_following: Option<bool>,
+}
+
+/// Returns all referrer groups from the summary.
+#[uniffi::export]
+pub fn get_stats_referrers_all_groups(
```
jkmassel: These methods can now be associated with the parent object across Uniffi boundaries – see #1120

jkmassel (changes requested):
Good start – this properly parsed both `https://public-api.wordpress.com/rest/v1.1/sites/en.blog.wordpress.com/stats/referrers` and `https://public-api.wordpress.com/rest/v1.1/sites/en.blog.wordpress.com/stats/top-posts` (with `summarize=1`).

Left some inline suggestions.

wp_api/src/wp_com/stats_referrers.rs
adalpari: Good catch! Added [here](https://github.com/Automattic/wordpress-rs/pull/1119/changes/e08f1954daca71edcfed173ba08ef2890780acda)

wp_api/src/wp_com/stats_referrers.rs
adalpari: I guess you are referring to `summarize` and `skip_archives` params being always _1_?

1. Summarize: it changes the output JSON format from a daily result to an aggregate one (in a non very standard way). And, since the cases I found are always using summaries, I opted to hardcode the value to be sure we are always getting those summaries. 
2. skip_archives: similar answer, but with archive-type referrers

Can RS gracefully handle two different JSON formats depending on the query params?

I guess, that if we can make it more generic, we can accept both params

wp_api/src/wp_com/stats_referrers.rs
adalpari: I'm thinking that we don0t really need these functions sonce they are just accessing to plain variables inside the objects. So, I assume the client can just follow `response.summary.total_views`

So, I might be removing them. Wdyt?

wp_api/src/wp_com/stats_top_posts.rs
adalpari: Done [here](https://github.com/Automattic/wordpress-rs/pull/1119/changes/ad07516d368d79b3c2a7046f62174937a44512ac)

Note: I'm not sure this is what you expected since I used Claude. If this is not the way of handling parameterized tests, and you could provide an example. It would be great.

wp_api/src/wp_com/stats_referrers.rs
adalpari: I've now made the arguments configurable by the client, supported both responses, and added tests for them. Let me know if there's anything I'm missing.

jkmassel: Let's:

1. Use opinionated defaults as in your initial implementation.
2. Allow overriding as needed.
3. Support both formats the API can emit.
4. Add inline and docblock documentation for how it works.

wp_api/src/wp_com/stats_referrers.rs
crazytonyli: > wdyt about using WPComLanguage?

That makes sense to me. I presume the `local` argument in this request is treated as the same as `WPComLanguage.slug` in the backend.

wp_api/src/wp_com/stats_referrers.rs
jkmassel: >  I might be removing them

Removing code is pretty much always a 👍 from me :D

wp_api/src/wp_com/stats_referrers.rs
```diff
@@ -0,0 +1,631 @@
...
+        assert_eq!(
+            query_pairs.finish().as_str(),
+            "https://public-api.wordpress.com/rest/v1.1/sites/1234/stats/referrers?period=day&summarize=0&skip_archives=0"
+        );
+    }
+
+    #[rstest]
+    #[case("tests/wpcom/stats_referrers/referrers-01.json")]
+    #[case("tests/wpcom/stats_referrers/referrers-03-follow-data-false.json")]
+    #[case("tests/wpcom/stats_referrers/referrers-04-real-response.json")]
```
jkmassel: This is a good parameterized test, but should include all of the test cases you've defined. Even if they're used for a more specific test, running every valid `json` test case you have through the generic deserialization test is more likely to catch issues.

Said another way – you have the data anyway, might as well use it :)

adalpari: You are right. The problem is that because of the endpoint specifications, some JSON files are different.
For instance, `referrers-02-days.json` has no period field. So, it cannot be included in this general test.
However, there's a specific test for it.

Note: I've added `referrers-05-with-nulls.json`, though, and an extra check.

Done here: https://github.com/Automattic/wordpress-rs/pull/1119/changes/2c589f43e55660f808d4a9a17db8c008b7ae552c

jkmassel: > For instance, `referrers-02-days.json` has no period field

This seems odd – `https://public-api.wordpress.com/rest/v1.1/sites/en.blog.wordpress.com/stats/referrers?summarize=1&period=day` has a `period` field, but `https://public-api.wordpress.com/rest/v1.1/sites/en.blog.wordpress.com/stats/referrers?period=day` does not. Should we be using the former?

jkmassel:
To help land this quicker, I've:

1. Rebased this PR atop `trunk` now that #1011 and #1126 have landed.
2. Opened #1133 with an implementation of the WP.com e2e tests covering these new endpoints.

Still to do – if we really need both `summarize=1` and `summarize=0`, we should support both. The fact that we can't run the parser tests over all of our JSON files means something's not quite right – we should be able to (if they have different expectations – like one should always have `Some` summary, for instance, that can be baked into the test definition and doesn't need to be part of the test function body.

adalpari:
> To help land this quicker, I've:
> 
> 1. Rebased this PR atop `trunk` now that [Add language endpoint and database #1011](https://github.com/Automattic/wordpress-rs/pull/1011) and [Rework WP.com e2e tests to remove custom test harness and run in CI #1126](https://github.com/Automattic/wordpress-rs/pull/1126) have landed.
> 2. Opened [Cmm 1177 suggestions #1133](https://github.com/Automattic/wordpress-rs/pull/1133) with an implementation of the WP.com e2e tests covering these new endpoints.

Thank you!!

> Still to do – if we really need both `summarize=1` and `summarize=0`, we should support both. The fact that we can't run the parser tests over all of our JSON files means something's not quite right – we should be able to (if they have different expectations – like one should always have `Some` summary, for instance, that can be baked into the test definition and doesn't need to be part of the test function body.


We are already supporting both outputs. In tests, we can parse both as well (but in different tests). 
However, given that each test checks different output fields, they cannot be mixed. One test checks `period` and `summary.groups`, while the other one checks `summary.is_none()` and `days.is_some()`. 

I've tried to follow your suggestion about "baking it into the test definition". So, let me know if you think this is a good approach: https://github.com/Automattic/wordpress-rs/pull/1119/changes/fde295ef2392280590f3888f3a5d21c35deeee08

adalpari:
Tests renaming https://github.com/Automattic/wordpress-rs/pull/1119/changes/6a1917fbc4389541a3fa94444a6b2b44e3a09953

jkmassel (approved):

PR #1130, opened 2026-01-29
Localized FetchError and CollectionError

crazytonyli:
Localize these errors because they will be used to show error messages on UI.

### DIFF
--- a/wp_localization/localization/en-US/main.ftl
+++ b/wp_localization/localization/en-US/main.ftl
@@ -73,4 +73,4 @@ parse_api_root_failure_reason_wordfence_blocking_access = Wordfence is blocking 
 
 already_logged_in = You are already logged in as {$username}.
 
-database_generic_message = There was a problem loading your data.
+database_generic_message = There was a problem loading your data: {$reason}.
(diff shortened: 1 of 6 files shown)


wp_localization/localization/en-US/main.ftl
```diff
@@ -72,3 +72,5 @@ parse_api_root_failure_reason_server_fatal_error = Your server encountered an un
 parse_api_root_failure_reason_wordfence_blocking_access = Wordfence is blocking access to the site's API. Please check your Wordfence configuration.
 
 already_logged_in = You are already logged in as {$username}.
+
+database_generic_message = There was a problem loading your data.
```
jkmassel: Could we ensure that this message always contains the underlying SQLite message? Otherwise it's pretty hard to tell what's going on if a user reports it.

jkmassel (changes requested):

wp_localization/localization/en-US/main.ftl
crazytonyli: Addressed in https://github.com/Automattic/wordpress-rs/pull/1130/commits/f91bd8b5dd32b525cc7cf0a47c99e52aac88061a

jkmassel (approved):

PR #1131, opened 2026-01-29
CMM-1202 create the stats country endpoint

adalpari:
## Summary                                                                                                                                                                      
                                                                                                                                                                                  
  - Add new `stats/country-views` endpoint for WordPress.com REST API v1.1                                                                                                        
  - Returns location/country-based view statistics with support for both summary and per-day breakdown modes                                                                      
  - All response fields support null values for robust API handling                                                                                                               
                                                                                                                                                                                  
  ## Changes                                                                                                                                                                      
                                                                                                                                                                                  
  **New files:**                                                                                                                                                                  
  - `wp_api/src/wp_com/stats_country_views.rs` - Type definitions and parameters                                                                                                  
  - `wp_api/src/wp_com/endpoint/stats_country_views_endpoint.rs` - Endpoint implementation                                                                                        
  - `wp_api/tests/wpcom/stats_country_views/*.json` - Test fixtures for parameterized tests                                                                                       
                                                                                                                                                                                  
  **Types added:**                                                                                                                                                                
  - `StatsCountryViewsPeriod` - Period enum (day, week, month, year)                                                                                                              
  - `StatsCountryViewsParams` - Query parameters (period, date, start_date, max, num, days, locale, summarize)                                                                    
  - `StatsCountryViewsResponse` - Response with `summary` or `days` depending on `summarize` parameter                                                                            
  - `StatsCountryInfo` - Country metadata (flag icons, country name, map region)                                                                                                  
  - `StatsCountryView` - Individual country view entry (location, views, country_code)                                                                                            
                                                                                                                                                                                  
  **Response modes:**                                                                                                                                                             
  - `summarize=0` (default): Returns `days` field with per-day breakdown keyed by date                                                                                            
  - `summarize=1`: Returns `summary` field with aggregated country views                                                                                                          
                                                                                                                                                                                  
  ## Test plan                                                                                                                                                                    
                                                                                                                                                                                  
  - [x] Unit tests for parameter serialization                                                                                                                                    
  - [x] Parameterized tests for JSON deserialization covering summary, days, null values, and empty responses                                                                     
  - [x] Lint passes (`cargo fmt`, `cargo clippy`)


adalpari:
@jkmassel I added the remaining stuff to this endpoint:
- WPComLanguage
- Renaming and parametrizing json deserialization tests
- e2e test

wp_api/src/wp_com/stats_country_views.rs
```diff
@@ -0,0 +1,482 @@
...
+    #[uniffi(default = None)]
+    pub locale: Option<WPComLanguage>,
+    /// Whether to return a summary of the data.
+    ///
+    /// - `Some(true)` (summarize=1): Response contains `summary` field with aggregated country
+    ///   views across all requested periods. The `days` field will be absent.
+    /// - `Some(false)` (summarize=0, default): Response contains `days` field with a per-day
+    ///   breakdown of country views, where each day is keyed by its date string. The `summary`
+    ///   field will be absent.
+    #[uniffi(default = Some(false))]
```
jkmassel: We should match the endpoints from the last PR – if we think this should be `true` by default we should have it be `default=true` and remove the optionality.

wp_api/src/wp_com/stats_country_views.rs
```diff
@@ -0,0 +1,482 @@
...
+            let days = response
+                .days
+                .as_ref()
+                .expect("Days should be present for non-summarized response");
+            assert!(!days.is_empty());
+        }
+    }
+
+    #[test]
+    fn test_stats_country_views_response_deserialization_summary_01() {
```
jkmassel, line 307: Does this need to be its own test?

It seems to be mostly covered by the test above – should we add more parameters to it?

wp_api/src/wp_com/stats_country_views.rs
adalpari: Ahh, I forgot about this 🤦 
[Done](https://github.com/Automattic/wordpress-rs/pull/1131/changes/4aa8913182c6390caf4bbbfe58af45ba05dd4fd7)

jkmassel (changes requested):

wp_api/src/wp_com/stats_country_views.rs
adalpari: There's a small difference between these tests. The above one is a lightweight sanity check test for general purposes (parameterised), while this one is more about checking the detailed values. Si, it's not only checking the deserialisation, but checking that the expected values are present.
Hence, I think that both tests add value in their way. Plus, I think that having extra tests is harmless.

Does it make sense?

jkmassel: I agree it's not hurting anything, so I'm fine to leave it :)

jkmassel (approved):

PR #1133, opened 2026-01-29
Cmm 1177 suggestions

jkmassel:


wp_api/src/wp_com/stats_referrers.rs
```diff
@@ -54,14 +55,13 @@ pub struct StatsReferrersParams {
...
     ///
-    /// - `Some(true)` (default): Response contains `summary` field with aggregated data
-    /// - `Some(false)`: Response contains `days` field with per-day breakdown
-    /// - `None`: Parameter is not sent to the API
-    #[uniffi(default = Some(true))]
-    pub summarize: Option<bool>,
+    /// - `true` (default): Response contains `summary` field with aggregated data
+    /// - `false`: Response contains `days` field with per-day breakdown
+    #[uniffi(default = true)]
+    pub summarize: bool,
```
adalpari, line 64: ❓ Any reason why not accepting null?

wp_api/src/url_query.rs
```diff
@@ -26,6 +26,21 @@ pub trait QueryPairsExtension {
     ) -> &mut Self
     where
         T: AsQueryValue;
+    /// Conditionally appends a boolean parameter as `key=1` only when true.
+    ///
+    /// This is useful for API parameters where:
+    /// - `true` means "enable this option" and should be sent as `key=1`
+    /// - `false` means "use server default" and should not be sent at all
```
adalpari, line 33: ⛏️ I'm not sure if using false to "use server as default" is clear. I mean, what if the client wants to specifically set the value to false? 
Which can actually be a scenario if we are not looking for aggregates using `summary`

adalpari: Extra note: by allowing `null`, the client can set it to `true`, `false`, or `server default`

adalpari: After chatting about this, this sounds like a plan!

adalpari (approved):
LGTM!

PR #1134, opened 2026-01-30
[DO NOT MERGE] Internal: merge endpoints prs for testings

adalpari:
This PR hasbeen created to deploy the two new stats endpoints and be able to referentiate it from the client.


adalpari:
We don't need it for testing anymore.

PR #1135, opened 2026-02-02
Add a few more plugins that blocks XML-RPC

crazytonyli:
Relates to https://github.com/wordpress-mobile/WordPress-iOS/pull/25183

The new plugins and new `xmlrpc_blocking_plugins` API will be used in the app to show what plugins may blocked the site's XML-RPC endpoint.

### DIFF
--- a/wp_api/src/login.rs
+++ b/wp_api/src/login.rs
@@ -1,6 +1,6 @@
 use crate::{
     JsonValue, login::url_discovery::is_local_dev_environment_url, parsed_url::ParsedUrl,
-    uuid::WpUuid,
+    plugins::PluginSlug, uuid::WpUuid,
 };
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, str, sync::Arc};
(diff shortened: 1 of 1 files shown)


jkmassel (changes requested):
I'm not sure how this should work – the application blocking plugins can be detected by fetching the JSON API root, but these XMLRPC ones would only be detectable _after_ the user is authenticated, right?

crazytonyli:
> but these XMLRPC ones would only be detectable after the user is authenticated, right?

Yes, I plan to show them in error messages, which will be built on top of https://github.com/wordpress-mobile/WordPress-iOS/pull/25183

jkmassel:
> Yes, I plan to show them in error messages

This is a separate type of error though – it's not a `KnownAuthenticationBlockingPlugin`. Can we introduce a different type and list for this to avoid making `KnownAuthenticationBlockingPlugin` more complex?

crazytonyli:
Sure. I saw there is an `xmlrpc` function there, I thought I'd build on that.

But after a second look in the error message UI in the app, I think I'll stick with a general message for now.

PR #1136, opened 2026-02-02
Update get_csv implementation to support URL encoded by WordPress core

crazytonyli:
Relates to https://github.com/wordpress-mobile/GutenbergKit/pull/299#discussion_r2737781423.

`wp/v2/posts?status=any` response's "next link" is not parsed correctly at the moment. The "next link" contains "status[0]=any", whereas the library (the `get_csv` function specifically) looks for the `status` key. This results in the `status` query being lost.

I have added an integration test to this PR, and Claude Code added some unit test to verify the fix.

### DIFF
--- a/wp_api/src/url_query.rs
+++ b/wp_api/src/url_query.rs
@@ -145,28 +145,30 @@ impl<'a> UrlQueryPairsMap<'a> {
     /// When an array parameter like `status=["draft", "publish"]` is serialized,
     /// PHP produces `status[0]=draft&status[1]=publish` instead of `status=draft,publish`.
     ///
-    /// This function handles that format by looking up `key[0]`, `key[1]`, etc.
-    /// in sequence, stopping when a gap is encountered.
+    /// This function handles that format by scanning all keys matching `key[$int]`,
+    /// where `$int` is any valid integer. Non-sequential indices are supported.
     ///
     /// See:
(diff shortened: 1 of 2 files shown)


wp_api/src/url_query.rs
```diff
@@ -255,4 +290,75 @@ mod tests {
...
+        pairs.insert(Cow::Borrowed("status[1]"), Cow::Borrowed("publish"));
+        pairs.insert(Cow::Borrowed("page"), Cow::Borrowed("1"));
+
+        let map = UrlQueryPairsMap::new(pairs);
+        let statuses: Vec<String> = map.get_csv("status");
+        assert_eq!(statuses, vec!["draft".to_string(), "publish".to_string()]);
+    }
+
+    #[test]
+    fn test_get_csv_with_php_array_stops_at_gap() {
```
jkmassel: I don't think it should work this way – PHP makes it really easy to end up with a valid state that looks like this (by using `array_unique` for instance). If the server sends back `status[0]=draft&status[4]=publish&status[4234]=pending`, we know what it means, and I think we should handle it properly?

jkmassel (changes requested):

wp_api/src/url_query.rs
crazytonyli: Addressed in https://github.com/Automattic/wordpress-rs/pull/1136/commits/d8631629af8e146989e1b20368e9ca71dc89dd19.

jkmassel (approved):

PR #1147, opened 2026-02-04
Add a couple APIs to return site home url and timezone info

crazytonyli:
Relates to https://github.com/wordpress-mobile/WordPress-iOS/pull/25183. These info will be used to populate the `Blog.options` in the app.

### DIFF
--- a/wp_api/src/login.rs
+++ b/wp_api/src/login.rs
@@ -192,6 +192,16 @@ impl WpApiDetails {
             .is_ok_and(|parsed_url| is_local_dev_environment_url(&parsed_url))
     }
 
+    /// Returns the site home URL as a string.
+    pub fn home_url_string(&self) -> String {
+        self.home.clone()
+    }
+
+    /// Returns the site GMT offset.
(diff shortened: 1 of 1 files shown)


jkmassel (approved):

wp_api/src/login.rs
```diff
@@ -192,6 +192,16 @@ impl WpApiDetails {
             .is_ok_and(|parsed_url| is_local_dev_environment_url(&parsed_url))
     }
 
+    /// Returns the site home URL as a string.
+    pub fn home_url_string(&self) -> String {
```
jkmassel, line 196: I wonder if it would be worthwhile to have a `ParsedUrl` variant of this that's guaranteed to be a valid URL?

Not a blocker, just for your consideration.

crazytonyli: I think it's okay both ways. For the specific case of how I plan to use the new API in the app, I don't plan to parse it: the original value is stored and they will be parsed at different places where it's used.

PR #1164, opened 2026-02-13
Rename `WpSelfHostedService` and add WP.com sites support

crazytonyli:
Context: p1770902854097129-slack-C06S1QS55K9

Adds a new `WpService.wordPressCom` constructor, which accepts a site id.

@nbradbury The CI job will publish an artifact, which you can try and see if there is any issue.

### DIFF
--- a/native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
+++ b/native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
@@ -10,6 +10,6 @@ class WordPressApiCacheTest {
 
     @Test
     fun testThatMigrationsWork() = runTest {
-        assertEquals(9, WordPressApiCache().performMigrations())
+        assertEquals(10, WordPressApiCache().performMigrations())
     }
 }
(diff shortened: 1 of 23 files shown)


nbradbury:
@crazytonyli I'll leave the code review to @jkmassel, but I verified that this works correctly in [this WPAndroid branch](https://github.com/wordpress-mobile/WordPress-Android/tree/nick/rs-test) which includes the rename.

native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
```diff
@@ -10,6 +10,6 @@ class WordPressApiCacheTest {
 
     @Test
     fun testThatMigrationsWork() = runTest {
-        assertEquals(9, WordPressApiCache().performMigrations())
+        assertEquals(10, WordPressApiCache().performMigrations())
```
jkmassel, line 13: We should probably rework this to just care about success instead of the specific number. WDYT?

jkmassel (approved):

native/kotlin/api/kotlin/src/integrationTest/kotlin/WordPressApiCacheTest.kt
crazytonyli: Absolutely. IIRC, we talked about these unit tests a while ago. I can have a look at it later.

PR #1175, opened 2026-02-18
Default interceptors to empty list in Android constructors

jkmassel:
## Description

The `interceptors` parameter in the secondary constructors of `WpLoginClient` and `WpRequestExecutor` was the only parameter without a default value, requiring callers to always pass it explicitly even when no interceptors are needed.

## Changes

- Add `= listOf()` default value for `interceptors` in `WpLoginClient` secondary constructor
- Add `= listOf()` default value for `interceptors` in `WpRequestExecutor` secondary constructor

## Test plan

- Verify existing callers that pass interceptors explicitly still compile and work
- Verify callers can now omit the `interceptors` parameter


adalpari (approved):
LGTM!

PR #1177, opened 2026-02-20
Parse custom taxonomy terms in SparseAnyPost

crazytonyli:
## Description

We can't explicitly parse the custom taxonomy terms because the taxonomy keys are dynamic. This PR adds a new `SparseAnyPost.additional_fields` property, which captures all keys that are not listed in the `SparseAnyPost`.

The cache table is also updated to store the new properties.

These new API will be used in the app to support updating taxonomy terms from custom post's Post Settings.


scripts/swift-bindings.sh
```diff
@@ -47,6 +47,15 @@ function patch_wp_api {
...
 #endif/' "$swift_binding"
+
+    cat >> "$swift_binding" <<'PATCH'
+
+// AnyPostWith*Context types contain `AnyJson` (a reference type) which
+// prevents automatic Hashable synthesis. Add the conformance manually.
+extension AnyPostWithEditContext: Hashable {}
+extension AnyPostWithEmbedContext: Hashable {}
+extension AnyPostWithViewContext: Hashable {}
+PATCH
```
jkmassel, line 58: Do you also want these?

```swift
extension SparseAnyPostWithEditContext: Hashable {}
extension SparseAnyPostWithEmbedContext: Hashable {}                                                                                                                          
extension SparseAnyPostWithViewContext: Hashable {}
```

wp_api/src/lib.rs
```diff
@@ -140,6 +140,39 @@ pub struct AnyJson {
     pub raw: Value,
 }
 
+#[uniffi::export]
+impl AnyJson {
```
jkmassel, line 144: I wonder if we'll want `AnyPostJson` or something later on – I could see this `impl` getting very full over time?

Maybe an issue to address later?

jkmassel (approved):

scripts/swift-bindings.sh
crazytonyli: I'll leave these out for now, because they are not used by the app atm.

wp_api/src/lib.rs
crazytonyli: Good point. Probably for later, though. I'm not sure if there will be more helper methods in AnyJson.

PR #1178, opened 2026-02-20
Fix a crash after using a higher versioned database

crazytonyli:
## Description

The changes are for the purpose of helping with local development. During development, we may use a version of this library with a higher database schema version, then switch to an older version that uses a lower version of the database schema. The `MIGRATION_QUERIES[next_migration_id..]` would crash in that case, due to being out of bounds.


jkmassel (approved):
Presumably we can catch this error, examine it, realize the DB is corrupt, then throw it away and start from scratch?

crazytonyli:
@jkmassel Yes, that's the existing api cache initialization logic in the iOS app.

PR #1180, opened 2026-02-23
Update metadata to bring in post changes

crazytonyli:
## Description

Relates to https://linear.app/a8c/issue/CMM-1264.

The `PostMetadataCollectionWithEditContext` internal state is driven by the pagination result (total pages & total items) and the metadata for each posts that belong to the pagination result. So far, we have been using it in a read-only environment, where the posts do not change.

When users change posts, for example, publishing a draft, creating a new draft, trashing a post, etc, the post details may be saved into the database, but they are ignored by `PostMetadataCollectionWithEditContext`. Because the metadata collection does not really care about the post details. However, `PostMetadataCollectionWithEditContext` should be automatically updated to reflect the post changes. For example, for a draft list collection, the draft post should be automatically removed from the list when published.

The correct, in terms of maintaining data consistency, way to implement it would be to call the core REST API to refresh the pagination results, which in turn will trigger metadata updates and post updates; finally, everything will be correct eventually. However, that means a series of HTTP requests is required for the post changes to be reflected on `PostMetadataCollectionWithEditContext`. That's not a very good UX from the app's pov.

This PR went another direction: the `PostMetadataCollectionWithEditContext` will be called internally to bring in the post changes: by updating the metadata based on the post details in the database. The whole thing happens locally, which means an instant UI update.

Please note: this PR adds a few updating posts functions to `PostService`, and they need to be used for all these to work, instead of the stateless pure REST API calls (`WpApiClient.posts.update(...)`). And the `PostService` that's used to update post need to be the same one that's used to create `PostMetadataCollectionWithEditContext` instances. Just a note to you, @nbradbury, this paragraph is relevant to you. You'll need to make sure `PostService` is used correctly to present the post list and update the posts.

## Tests

I have added a few integration tests to test scenarios: publish draft posts and create a new draft.


nbradbury:
> Just a note to you, @nbradbury, this paragraph is relevant to you. You'll need to make sure `PostService` is used correctly to present the post list and update the posts.

@crazytonyli Just so I'm clear, should I entirely drop using `WpApiClient` in the new Android post list in favor of `PostService` or just use `PostService` when updating posts?

crazytonyli:
@nbradbury I don't think you need to use WpApiClient much in the post list. You should be using the `PostMetadataCollectionWithEditContext` to present the post list. With the changes in this PR, you should use `PostService` to update posts. And I think that's all you needed in the post list.

jkmassel (approved):
It took me a bit to understand this, but it's a good optimization. 

The tl;dr for me was

> The PR just stops throwing away a perfectly good API response and instead feeds it into the local cache and collection logic. The server is the source of truth for the post's actual state; this PR just uses that truth more efficiently.

It seems like `loosely_matches_post` could be match more closely once #1177 lands? I think it's fine to leave that for later though – this is pretty isolated so tightening that up later shouldn't affect users of this.

crazytonyli:
> It seems like loosely_matches_post could be match more closely once https://github.com/Automattic/wordpress-rs/pull/1177 lands?

#1177 adds custom taxonomy terms to AnyPost, which is not supported by the PostListFilter. So, I don't think we need to change the function specificy about #1177? Let me know if I missed anything.

However, I noticed that categories and tags were not checked in PostListFilter, which is now added in https://github.com/Automattic/wordpress-rs/pull/1180/commits/224b41560e2bc20a8d75490a8e8ed39d264a9d3e.

PR #1185, opened 2026-02-24
Delete some unused files

crazytonyli:
I don't think these files are needed. Please let me know if that's not the case.


jkmassel:
Did you happen to check if the issues referenced are a problem for us still? I don't know off the top of my head, but if we're going to delete them I'd like some confidence that the issue was resolved (or at least moved into a GitHub issue)

crazytonyli:
Sorry, I didn't know they are used (like a bug report?). I just happened to see them in the root level and thought they were some leftovers from previous work. I'll close this PR.

PR #1189, opened 2026-02-24
Add PasswordTransformer trait with platform implementations

jkmassel:
## Description

Add encrypted credential storage for persisting site authentication across app launches.

This PR introduces a `PasswordTransformer` trait and platform-specific implementations that encrypt credentials at rest, enabling `AccountRepository` (also introduced here) to persist sites and tokens securely.

This is the foundation for credential storage in:
- This project's example apps (#1190)
- The GutenbergKit sample apps
- Eventually the WordPress/Jetpack iOS and Android apps

## Changes

1. **`PasswordTransformer` trait + AES-256-GCM implementation** (`wp_mobile`)
   - UniFFI-exported trait with `encrypt`/`decrypt` methods
   - Uses `#[uniffi::export(with_foreign)]` so platforms can provide hardware-backed implementations
   - Pure-Rust `AesGcmPasswordTransformer` serves as a cross-platform fallback (particularly for Linux where platform keystores are unavailable)
   - `AccountRepository` for storing and retrieving encrypted site credentials

2. **Swift bindings for persistence types**
   - Re-exports `AccountRepository`, `PasswordTransformer`, etc. to the public Swift API
   - Conditionally exposes `AesGcmPasswordTransformer` on Linux via `#if os(Linux)`
   - On Apple platforms, `SecureEnclavePasswordTransformer` fills this role instead

3. **`SecureEnclavePasswordTransformer`** (Apple)
   - ECIES encryption (P-256 ECDH + HKDF-SHA256 + AES-256-GCM) backed by the Secure Enclave
   - Falls back to software keys on simulators
   - Keys persisted via Keychain with the application name as `kSecAttrService` for visibility in Keychain Access
   - Tests run under `@Suite(.serialized)` to prevent cooperative thread pool deadlocks during SE key creation

4. **`KeystorePasswordTransformer`** (Android)
   - AES-256-GCM encryption backed by the Android Keystore
   - Prefers StrongBox (API 28+), falls back to TEE, works in software on emulators
   - Exposes `isHardwareBacked` property so callers can check the security level
   - Explicit base64 validation on decrypt for clear error messages

## Test plan

- [x] `cargo test --lib` — Rust unit tests for `AesGcmPasswordTransformer` and `AccountRepository`
- [x] Swift unit tests for `SecureEnclavePasswordTransformer` (22 tests: round-trip, key persistence, error paths)
- [x] Android instrumented tests for `KeystorePasswordTransformer` (15 tests: round-trip, key reuse, error paths, tampered ciphertext)
- [x] SwiftLint and Detekt pass
- [x] `cargo clippy` and `cargo fmt` pass


crazytonyli:
Just for my understanding, we are going to use `SecureEnclavePasswordTransformer` on Apple platforms, `AesGcmPasswordTransformer` on Linux, and `KeystorePasswordTransformer` on Android, right? I'm not familiar with Android, so I won't comment on that.

The `SecureEnclavePasswordTransformer` seems to contain two implementations:
1. one backed by a SecureEnclave private key (not directly usable by third parties), and
2. the other with a traditional private key that's stored on disk

IMO, the second kind of private key is not secure, because it's stored on disk without any protection. According to the comment on `SecureEnclavePasswordTransformer(keyFile:)`, it's intended to be used for the dev env. If that's the only use case, do we even need to go through the encryption and decryption process? I think we can just have a `PlaintextPasswordTransformer`, which has no encryption, for local development (like the example apps)?

Also, I'm not sure if the SecureEnclave encryption is necessary. I think we can delegate the encryption to the keychain? Here is a pseudo-code:

```swift
class KeychainPasswordTransformer {
  func encrypt(password: String) {
    let id = UUID()
    keychain.store(username: id, password: password, service: "wordpress-rs-accounts")
    return id
  }

  func decrypt(encrypted: String) -> String {
    keychain.get(username: encrypted, service: "wordpress-rs-accounts)
  }
}
```

----

I wonder if we can have a `SecureKeyValueStore` trait to replace the `PasswordTransformer`. The definition of `PasswordTransformer` requires the implementation to be able to encrypt and decrypt any plaintext. But I don't think we really need that level of abstraction.

Can we ask the platform to store the information for us securely, without thinking about encryption & decryption?

```rust
// pseudo-code: Just a simple key-value store.
trait SecureKeyValueStore {
  fn store(id: String /* AccountId */, data: Vec<u8>)
  fn get(id: String) -> Vec<u8>
}
```

The platform can decide how to store the data _securely_, so we likely don't need any encryption or decryption code.

jkmassel:
> Also, I'm not sure if the SecureEnclave encryption is necessary. I think we can delegate the encryption to the keychain? Here is a pseudo-code:

When running the app for macOS local development, you have to type your keychain password on every launch because the app's signature has changed. This is protection against Keychain enumeration attacks – if I were reviewing a patch someone contributed, you don't want it to read every keychain entry and upload what it finds to some remote server. 

But that makes local development super annoying. One solution for this is using an Apple Development identity for dev builds, but then you need to propagate that to CI. That's a lot of complexity.

In this PR, the Apple platforms approach works the same as the Android one – the system vends a key out of its secure hardware, and we use that to encrypt the secrets in-place. Here's an outline of how it's stored for each case:

Platform | Key Type | Key Storage | Key in Storage
-- | -- | -- | --
iOS / tvOS / watchOS / visionOS (device) | Secure Enclave P-256 | Keychain | Opaque reference only
iOS / tvOS / watchOS / visionOS (simulator) | Software P-256 | Keychain | Raw key, Keychain-protected
macOS (Apple Silicon / T2) | Secure Enclave P-256 | Keychain | Opaque reference only
macOS (no Secure Enclave) | Software P-256 | Keychain | Raw key, Keychain-protected
macOS (Apple Silicon / T2) + keyFile | Secure Enclave P-256 | File (0o600) | Opaque reference only
macOS (no SE) + keyFile | Software P-256 | File (0o600) | Raw key material
Linux | AES-256-GCM | Caller-provided shared secret | N/A

The only insecure storage mechanism here is macOS without a SEP – that's Intel machines without a T-series coprocessor. 

I don't really want the protocol to act as a raw K/V store, because another big benefit to storing a single keychain entry for the whole app is that you don't run into sync issues between what's in the keychain and what's in the app's storage and you get guarantees about avoiding race conditions or write tearing.

crazytonyli:
> When running the app for macOS local development, you have to type your keychain password on every launch because the app's signature has changed.

Yeah, I get what you want to avoid here. But I think it's okay to bypass any security measures for local development and just write the password to a file?

I don't know much about cryptography. The large chunk of cryptography code looks quite intimidating to me, which is why I'd prefer to delegate that to the system. That's pretty much where all my suggestions are centered around 😅 

> I don't really want the protocol to act as a raw K/V store, because another big benefit to storing a single keychain entry for the whole app

You can generate a secret in Rust and store it in the K/V store, and use AesGcmPasswordTransformer internally. In doing that, you don't need the cryptography code in Swift and Kotlin.

Here is a potentially similar way to remove the cryptography code in Swift and Kotlin:

```swift
class Encryption: PasswordTransformer {
  let transformer: AesGcmPasswordTransformer
  init() {
    let secret: String = getOrCreateFromKeychain()
    transformer = AesGcmPasswordTransformer(secret: secret)
  }
  
  func encrypt() {
    transformer.encrypt()
  }

  func decrypt() {
    transformer.decrypt()
  }
}
```

jkmassel:
> The large chunk of cryptography code looks quite intimidating to me, which is why I'd prefer to delegate that to the system

I've gone back and annotated the encryption process to make it easier to follow. It is substantially delegated to the system (specifically the SEP hardware), we're just using the system primitives.

> You can generate a secret in Rust and store it in the K/V store, and use AesGcmPasswordTransformer internally

This has a few downsides – it's actually substantially less secure because we're not using the system, we're using application code for the most part. On CryptoKit-supported platforms, it's also missing out on whatever hardware/software integration Apple offers (ie – using hardware-accelerated computation and the SEP). Also not a huge thing, but the Rust implementation doesn't compile on Apple Watch because it needs hardware instructions that device doesn't have. I don't except we'll need Apple Watch support any time soon, but I'd rather use platform APIs to make sure we're maintaining forward compatibility for as long as possible.

> You can generate a secret in Rust

From a security perspective, it's IMHO preferable to me to have CryptoKit and the SEP generate the secret using [its KDF internals](https://developer.apple.com/documentation/cryptokit/sharedsecret/hkdfderivedsymmetrickey(using:salt:sharedinfo:outputbytecount:)) because that's one of the more important security pinch points.

native/swift/Sources/wordpress-api/SecureEnclavePasswordTransformer.swift
```diff
@@ -0,0 +1,542 @@
...
+    /// The size of an uncompressed P-256 public key in X9.63 format.
+    private static let publicKeySize = 65
+
+    /// The size of the random HKDF salt generated per encryption.
+    private static let saltSize = 32
+
+    private static let sharedInfo = Data("wordpress-rs-password-encryption".utf8)
+
+    private static let seKeyTag: UInt8 = 0x01
+    private static let softwareKeyTag: UInt8 = 0x02
```
crazytonyli, line 134: Should this implementation be split into two types: one that uses a secure enclave, and the other that doesn't? That'd save you from storing this tag in the private key file.

native/swift/Sources/wordpress-api/SecureEnclavePasswordTransformer.swift
```diff
@@ -0,0 +1,542 @@
...
+    ///   - applicationName: An identifier for the stored key. Different names
+    ///     produce independent keys. Used as `kSecAttrService` in the Keychain.
+    ///   - keychainStorage: The storage backend used to persist the key.
+    ///     Defaults to ``SystemKeychainStorage`` which uses the system Keychain.
+    /// - Throws: ``SecureEnclavePasswordTransformerError/keychainError(_:)``
+    ///   if storage access fails.
+    public convenience init(
+        applicationName: String,
+        keychainStorage: KeychainStorage? = nil
+    ) throws {
```
crazytonyli: What do you think about making this function an internal function, which means we can keep `KeychainStorage` internal, too?

native/swift/Sources/wordpress-api/SecureEnclavePasswordTransformer.swift
```diff
@@ -0,0 +1,542 @@
...
+        try keyData.write(to: tempFile)
+        try FileManager.default.setAttributes(
+            [.posixPermissions: 0o600],
+            ofItemAtPath: tempFile.path
+        )
+
+        do {
+            // Atomic move — succeeds only if no file exists at keyFile yet.
+            try FileManager.default.moveItem(at: tempFile, to: keyFile)
+            return newKey
```
crazytonyli, line 244: Is the above a re-implementation of [`Data.write(to:atomically:)`](https://developer.apple.com/documentation/foundation/nsdata/write(to:atomically:))?

wp_mobile/src/persistence/account_repository.rs
```diff
@@ -0,0 +1,853 @@
...
+        Ok(id)
+    }
+
+    pub fn all(&self) -> Result<Vec<Account>, AccountRepositoryError> {
+        let state = self.state.lock().expect("poisoned mutex");
+        state
+            .accounts
+            .iter()
+            .map(|a| {
+                a.decrypt(self.password_transformer.as_ref())
```
crazytonyli, line 450: Any reason not to store the decrypted account to avoid calling decrypt repeatedly?

wp_mobile/src/persistence/account_repository.rs
```diff
@@ -0,0 +1,853 @@
...
+        AccountRepositoryError::PasswordError {
+            reason: e.to_string(),
+        }
+    }
+}
+
+/// Internal state guarded by a single mutex to prevent lock-ordering issues.
+struct AccountRepositoryState {
+    accounts: Vec<StoredAccount>,
+    next_id: AccountId,
```
crazytonyli, line 238: It looks like this `next_id` is only unique per `AccountRepository` instance. If you have multiple `AccountRepository` instances, their `next_id` might be the same. Since all `AccountRepository` instances write to the same account file, account IDs may be duplicated.

crazytonyli (approved):

native/swift/Sources/wordpress-api/SecureEnclavePasswordTransformer.swift
jkmassel: No, because we want a failure if the file already exists to avoid concurrency issues – we don't want to blindly overwrite, we want to handle a race properly.

Probably not something that's common, but at this layer I want to optimize for correctness under all conditions even if the code is a bit more complex inside.

wp_mobile/src/persistence/account_repository.rs
jkmassel: This is a solid point – addressed in https://github.com/Automattic/wordpress-rs/pull/1189/commits/8d021aea1beeb54bd9ea316f7b3fc3d20de83365. 

It'll enforce that there's only a single instance used per-process (on a per-file basis). If more than one class tries to access the same underlying storage in the same process, it'll panic because that's a programmer error.

It works on a per-file basis so that under test we could have `test-account-1.json` and `test-account-2.json` and everything works fine.

I've also documented the expected usage patterns.

native/swift/Sources/wordpress-api/SecureEnclavePasswordTransformer.swift
jkmassel: Yep, good call – it's just to avoid using a real keychain under test so it shouldn't be public.

wp_mobile/src/persistence/account_repository.rs
jkmassel: It's a bit more secure but the benefit is kinda theoretical – if a user has access to the apps memory space, all bets are off anyway.

That said, this should never be a performance issue because decrypting all of these should be faster than the IO required to read the JSON off the disk. 

If it becomes one, I'm happy to revisit and I'm pretty confident it wouldn't require a public API change.

native/swift/Sources/wordpress-api/SecureEnclavePasswordTransformer.swift
jkmassel: I think I prefer the complexity of an extra byte in the key reference over choosing an implementation path at runtime?

If we later discover this is the wrong decision we can roll it back by counting bytes in the key – 33 bytes for keys that are tagged, 32 for untagged ones.

PR #1190, opened 2026-02-24
Add WP.com OAuth login support

jkmassel:
## Description

Add WordPress.com OAuth login support alongside the existing self-hosted Application Passwords flow, with improvements to error handling and network availability detection.

> **Note for testers:** To test WP.com OAuth login, you need to set up OAuth credentials. Create a `wp_com_test_credentials.json` file based on the template in the same directory with your `client_id` and `client_secret` from [developer.wordpress.com/apps](https://developer.wordpress.com/apps). 

### Rust core
- Add OAuth2 configuration discovery during site login (`url_discovery`, `oauth2_configuration`)
- Add WP.com sites endpoint for site metadata
- Expose `blogId` and `blogUrl` in OAuth2 token response

### Kotlin library
- Add `NetworkAvailabilityProvider` interface to distinguish offline devices from DNS failures
- Add `ApiDiscoveryResult.userFacingErrorMessage()` for human-readable discovery errors
- Add `NetworkAvailabilityProvider` parameter to `WpApiClient`, `WpComApiClient`, `JetpackApiClient`, and `WpLoginClient` convenience constructors

### Kotlin example app
- Replace crash-on-error with toast-based error messages
- Add `ACCESS_NETWORK_STATE` permission for `ConnectivityManager`
- Wire `AccountRepository` for persistent site storage
- Fix WP.com OAuth storing `blogUrl` as API root instead of constructing proper `public-api.wordpress.com/wp/v2/sites/{blogId}` URL
- Fix WP.com accounts showing no display name

### Swift example app
- Migrate to `AccountRepository` with `SecureEnclavePasswordTransformer`
- Add WP.com OAuth login flow and credential bundle support
- Fix Xcode project relative path to be folder-name-agnostic (`../../..` instead of `../../../../wordpress-rs`)

### Tests
- Add `NetworkAvailabilityProvider { true }` to all Kotlin test files
- Add WP.com sites endpoint integration tests

## Test plan

- [ ] Self-hosted Application Passwords login works on Android
- [ ] Self-hosted Application Passwords login works on iOS
- [ ] WP.com OAuth login works on Android (you can use `wpmobilep2.wordpress.com` – you'll see it in the list, and that's about all it does right now)
- [ ] WP.com OAuth login works on iOS
- [ ] WP.com site displays site name (not blank) after login on Android
- [ ] Tapping "Users" on a WP.com site does not crash on Android (but also doesn't display anything)
- [ ] Entering a URL with no internet shows "No internet connection" toast on Android
- [ ] Entering a non-WordPress URL shows appropriate error message on Android
- [ ] Sites persist across app launches on both platforms


wp_api/src/wp_com/oauth2.rs
```diff
@@ -1,25 +1,124 @@
...
 };
 use serde::{Deserialize, Serialize};
 use url::Url;
 use wp_serde_helper::deserialize_u64_or_none_with_zero_as_none_from_string;
 use wp_serde_helper::deserialize_u64_or_string;
 
+/// WordPress.com OAuth2 permission scopes.
+///
+/// Each scope grants access to a specific area of the WordPress.com API.
+/// Scopes are backed by power-of-2 integer values suitable for bitwise operations.
```
crazytonyli: It looks like WpComOauthScope is used as a simple enum. The public API uses `Vec<WpComOauthScope>` to represent a list of scopes, rather than one `WpComOauthScope` with or values. Maybe the special values can be removed?

wp_api/src/login/login_client.rs
```diff
@@ -210,52 +211,75 @@ impl WpLoginClient {
         };
         let api_details = Self::parse_api_root(&fetch_api_details_response)?;
 
+        // Try Application Passwords first (preferred for self-hosted sites)
```
crazytonyli, line 214: I wonder if the API should return both, if they are all available in the `wp-json` response. The client can decide which one to use.

crazytonyli (approved):

wp_api/src/login/login_client.rs
jkmassel: I agree with you that this API can probably be refined – Application Passwords are what pretty much everything except WordPress.com will use, so I like them being the "primary" approach, but that doesn't mean it's the right way to structure this API!

wp_api/src/wp_com/oauth2.rs
jkmassel: Done in [292af32](https://github.com/Automattic/wordpress-rs/pull/1190/commits/292af32c333090f12773437d11abbd1da38fd4b3) – I was originally thinking this could eventually be a bit flag type using `bindings.swift.custom_types`. Probably overkill.

In that new commit, I added de-duplication (so you can't have `post post post`), which the bit flag type would've done too. That probably covers our needs for now.

PR #1191, opened 2026-02-24
Add data type screens to Android example app

jkmassel:
## Summary

Build out the Android example app's site screen with screens for all major WordPress data types, including error handling and empty states.

> **Note for testers:** To test WP.com OAuth login, you need to set up OAuth credentials. Create a `wp_com_test_credentials.json` file with your `client_id` and `client_secret` from [developer.wordpress.com/apps](https://developer.wordpress.com/apps). On Android, this file should be placed in the example app's assets; on iOS, add it to the example app's bundle. Without these credentials, the WP.com login button will not appear.

- Add screens for Comments, Media, Search, Settings, Site Health, Themes, and more
- Add loading states and infinite-scroll pagination across all data type screens
- Dynamically discover post types and taxonomies from the site's REST API
- Group site screen items into Posts, Taxonomies, Navigation, and System sections
- Remove hardcoded Pages screen in favor of dynamic post type discovery
- Add Application Passwords management screen
- Add shared `ErrorMessage` and `EmptyState` composables with `WpRequestResult.errorDescription()` for human-readable error messages
- Surface API errors and empty-but-successful results across all screens

Depends on #1190.

## Test plan
- [x] Android example app builds (`./gradlew :example:composeApp:assembleDebug`)
- [x] Data type screens load and paginate correctly
- [x] Dynamic post types and taxonomies appear on the site screen
- [x] Error messages appear when API requests fail
- [x] Empty state text appears for screens with no data

## Known Issues

WordPress.com sites (ex: mobile.blog) can be added to the app, but their content isn't browsable. This is expected – there'll be another, smaller PR later that properly sets up the special WP.com path routing.


native/kotlin/example/composeApp/src/commonMain/kotlin/rs/wordpress/example/shared/App.kt
```diff
@@ -39,55 +73,147 @@ fun App(authenticationEnabled: Boolean, authenticateSite: (String) -> Unit) {
...
+                    )
                 } else {
                     throw IllegalStateException("Authentication is disabled")
                 }
             }
             composable("site") {
                 SiteScreen(
-                    onUsersClicked = {
-                        navController.navigate("users")
+                    viewModel = currentSiteViewModel!!,
```
adalpari: Not important as it's an example app, but forcing the use of forced non-null elements like this one can be dangerous because of runtime crashes.

adalpari: Just commenting it, because I saw it in different parts of the code.

native/kotlin/example/composeApp/src/commonMain/kotlin/rs/wordpress/example/shared/ui/site/SiteViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+    val restBase: String
+)
+
+data class SiteTaxonomy(
+    val name: String,
+    val restBase: String
+)
+
+class SiteViewModel(private val apiClient: WpApiClient) {
+    private val viewModelScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
adalpari: When using a ViewModel, we should extend from actual `ViewModels` and use `viewModelScope` instead of creating our own scopes.
This is repeated along multiple the VMs of the PR

native/kotlin/example/composeApp/src/commonMain/kotlin/rs/wordpress/example/shared/App.kt
```diff
@@ -10,26 +10,60 @@ import androidx.navigation.compose.NavHost
...
 
     // State to hold the current post type slug for navigation
     var currentPostTypeSlug by remember { mutableStateOf("post") }
+    var currentWpService by remember { mutableStateOf<WpService?>(null) }
+    var currentAccount by remember { mutableStateOf<Account.SelfHostedSite?>(null) }
+    var currentWpComClient by remember { mutableStateOf<WpComApiClient?>(null) }
+    var currentApiClient by remember { mutableStateOf<WpApiClient?>(null) }
+    var currentSiteViewModel by remember { mutableStateOf<SiteViewModel?>(null) }
+    var currentPostType by remember { mutableStateOf<SitePostType?>(null) }
+    var currentTaxonomy by remember { mutableStateOf<SiteTaxonomy?>(null) }
```
adalpari, line 67: I assume we are sharing all these states along the different screens by having them on the App level. But I think this is not a good pattern and can lead to some critical problems, like the nullability issue I mentioned in my other comment, or multiple screens modifying the same state and causing state corruption.

But the biggest problem, and the reason why I think it's an antipattern, is that every ViewModel must live inside its screen and follow the screen lifecycle, which is not the case. This is also tied to the comment about creating a custom scope per VM instead of using the default one.

adalpari (approved):
I left some comments about the code, but given it's a test app, I think it's not a blocker unless you want to improve it.

All the testing plan and smoke testing works goot do me.

🚢 it!

native/kotlin/example/composeApp/src/commonMain/kotlin/rs/wordpress/example/shared/ui/site/SiteViewModel.kt
jkmassel: I updated this in https://github.com/Automattic/wordpress-rs/pull/1191/commits/507987b7d7ec96f5ce0034fbce240c45958d4ed0

native/kotlin/example/composeApp/src/commonMain/kotlin/rs/wordpress/example/shared/App.kt
jkmassel: Updated in https://github.com/Automattic/wordpress-rs/pull/1191/commits/507987b7d7ec96f5ce0034fbce240c45958d4ed0

native/kotlin/example/composeApp/src/commonMain/kotlin/rs/wordpress/example/shared/App.kt
jkmassel: For a sample app it's not really correct but I don't think it's the end of the world? I'm fine to revisit it if it becomes a problem :)

PR #1193, opened 2026-02-25
Move creating new endpoints to a skill

crazytonyli:
## Description

Move the part about creating new endpoints to a skill, which trims claude.md from 2.7 tokens to 700 tokens.


crazytonyli:
It's not a literal cut-and-paste. I asked Claude to move that part to a skill, the skill-creator skill kicked in and created the skill, and I think it made some adjustments, judging from the diff stat.

jkmassel (approved):
Seems good - I also asked Claude to make suggestions based on recent PR review comments I've made to see if there'd be anything useful in there. Here's what it came up with:

> Actionable insights from your reviews:                                                               
>                                                                                                          
>   1. Prefer strong types over raw strings (PR #1147) — You suggested a ParsedUrl type instead of raw String for URLs. The skill should emphasize favoring strongly-typed wrappers beyond just IDs, e.g. for URLs, dates, or other structured data.                                                                 
>   2. Handle WordPress/PHP quirks in deserialization (PR #1136) — PHP can produce non-sequential array indices in query params (status[0]=...&status[4]=...). The skill should note that when deserializing responses or query params, be aware of PHP serialization quirks and handle them robustly.
>   3. Consistency across similar endpoints (PR #1131) — You flagged "match the endpoints from the last PR" — when adding a new endpoint that's similar to an existing one, follow the same patterns (defaults, parameter handling, etc.) as the prior implementation.
>   4. Avoid redundant tests (PR #1131) — "Does this need to be its own test? It seems to be mostly covered by the test above – should we add more parameters to it?" The skill's testing guidance should emphasize combining test cases where possible rather than creating near-duplicate tests.
>   5. Error messages must include underlying details (PR #1130) — When localizing or wrapping errors, always include the underlying error message/detail so debugging is possible.

crazytonyli:
I'll merge this PR as is. We can probably refine it if we run into issues when using it.

PR #1195, opened 2026-02-25
Test: SecureEnclave CI hang investigation experiments

jkmassel:
Temporary PR to run 5 experiments investigating why `swift test` hangs with SecureEnclave tests but standalone scripts work fine.

**Experiments:**
- **1a**: Single SE test in `swift test` (does a single call hang?)
- **1b**: 15 concurrent SE tests (cooperative thread pool exhaustion?)
- **2**: 15 serialized SE tests (does `.serialized` prevent the hang?)
- **3**: SE test with `.timeLimit(1 min)` (can timeout interrupt it?)
- **4**: Code signing comparison (test binary vs standalone binary)

Pipeline is stripped to only these experiments. All steps have 3-5 min timeouts.


jkmassel:
Closing: investigation artifact for Secure Enclave CI hang debugging. Root cause confirmed as cooperative thread pool exhaustion — fix applied in #1189.

PR #1186, opened 2026-02-24
Generate `var` properties in Swift structs

crazytonyli:
I have made and discarded this change a few times now. I think we'll need it this time. Use `PostUpdateParams` as an example (see [this code](https://github.com/wordpress-mobile/WordPress-iOS/blob/da15e6df83db7aacca8f90c8eb72a34adf324daf/WordPress/Classes/ViewRelated/Post/PostSettings/PostSettingsViewModel.swift#L610-L628)), I want to avoid manually copying properties to a new `PostUpdateParams` instance when we need to modify a few properties.

PR #1197, opened 2026-02-25
Add a step to enable mutability on certain structs

crazytonyli:
An alternative to https://github.com/Automattic/wordpress-rs/pull/1186, where we change the Swift binding code to allow mutability on certain structs.

### DIFF
--- a/scripts/swift-bindings.sh
+++ b/scripts/swift-bindings.sh
@@ -39,9 +39,34 @@ EOF
     done
 }
 
+# UniFFI generates Swift structs with immutable (`let`) properties. Some structs
+# need mutable properties so callers can modify fields after initialization.
+# This rewrites `public let` to `public var` for the specified structs.
+#
+# Usage: enable_mutability <swift_binding_file> <struct_name>...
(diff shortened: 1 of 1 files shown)


jkmassel (changes requested):
Could you show a usage example? Where would this be called from and what would it look like?

tbh I don't love this – I'd rather address it upstream. But I could live with it for a while until the upstream change lands so you're not blocked! 😅

crazytonyli:
The shell script is called as part of `make xcframework`. The result is the two types hard-coded in the shell script now have `var` properties:

```swift
public struct PostCreateParams: Equatable, Hashable {
    public var date: String?
    public var dateGmt: WpGmtDateTime?
    public var slug: String?
    // ...
}
```

> But I could live with it for a while until the upstream change lands so you're not blocked! 😅

Yeah, I wanted to try this solution mainly to unblock the app-side changes. I can look into submitting a PR to uniffi-rs later.

jkmassel (approved):
solid workaround for now :)

PR #1203, opened 2026-02-26
Speed up Android CI builds

jkmassel:
## Summary

Parallelize Rust cross-compilation in the Kotlin/Android CI pipeline to reduce wall-clock build time by ~30%.

### Changes

1. **Split the single sequential Rust build step into parallel agents** — Replace the "Build Rust libraries" step (which built all 5 targets sequentially on one agent) with 5 independent steps: 1 host build + 4 Android architecture builds via a Buildkite matrix. Each runs on its own agent and installs only the single Rust target it needs.

2. **Skip redundant Rust builds in publish and packaging steps** — Kotlin Publish downloads the pre-built host `.so` via artifact and passes `-x cargoBuildLibraryRelease` to Gradle. Android Publish downloads the 4 pre-built JNI libraries and passes `-x cargoBuild`. Both skip rebuilding Rust entirely.

3. **Build Example App from published S3 artifacts** — Instead of building all Rust targets from source (~45 min), the Example App step now consumes the artifacts published by the upstream Kotlin/Android publish steps via `-PwpApiKotlinVersion` and `-PwpApiAndroidVersion` Gradle properties. The Gradle build files conditionally resolve these as Maven coordinates when the properties are set, falling back to project references for local development.

4. **Make Gradle cargo targets configurable** — `build.gradle.kts` reads `-PcargoTarget` to build a single architecture, defaulting to all 4 when unset.

5. **Export published Android version as Buildkite meta-data** — Android Publish writes its version to meta-data so the Example App step can consume it.

### Timing

Baseline (build #4946) vs optimized builds:

| Metric | Baseline (#4946) | Optimized (#4959) | Change |
|--------|----------|--------|--------|
| **Wall-clock time** | 49.8 min | 36.8 min | **-13.0 min (-26%)** |
| **Total agent-minutes** | 281 min | 252 min | **-29 min (-10%)** |
| **Kotlin group critical path** | 45.4 min | 31.4 min | **-14.0 min (-31%)** |

The overall build is now bottlenecked by Swift e2e tests (~34 min), not the Kotlin group.

## Test plan

- [x] All 4 Android `.so` files correctly downloaded in Android Publish (verified: "Found 4 artifacts")
- [x] Example App builds successfully from published artifacts
- [x] Kotlin Publish downloads pre-built host binary and skips cargo build
- [x] Android matrix variants only install the single needed Rust target (verified in build logs)
- [x] Full CI build passes (builds #4955, #4959)

### DIFF
--- a/.buildkite/build-android-rust-target.sh
+++ b/.buildkite/build-android-rust-target.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
+
+GRADLE_TARGET=$1
+
+echo "--- :rust: Installing Rust"
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -v -y
+
+source "$HOME/.cargo/env"
+
+echo "--- :package: Installing Rust Target"
+
+# Map Gradle target name to Rust target triple
+case "$GRADLE_TARGET" in
+    arm)    RUST_TARGET="armv7-linux-androideabi" ;;
+    arm64)  RUST_TARGET="aarch64-linux-android" ;;
+    x86)    RUST_TARGET="i686-linux-android" ;;
+    x86_64) RUST_TARGET="x86_64-linux-android" ;;
+    *)      echo "Unknown target: $GRADLE_TARGET"; exit 1 ;;
+esac
+
+# Only install the single target needed — no Apple targets or nightly toolchain
+rustup target add "$RUST_TARGET"
+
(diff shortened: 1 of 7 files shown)


nbradbury (approved):
I'm not sure how to test this, but the fact that it's passing CI is probably sufficient. :shipit: 

PS: I'm always in favor of anything that speeds up CI builds. On WPAndroid, that's one of my biggest bottlenecks.

PR #1210, opened 2026-03-01
Support filtering by "any" status

crazytonyli:
## Description

The app uses the "any" status in the "All" tab. Support "any" status so that the "All" tab displays content correctly after updates.

### DIFF
--- a/wp_mobile/src/filters/post_list_filter.rs
+++ b/wp_mobile/src/filters/post_list_filter.rs
@@ -155,7 +155,8 @@ pub struct PostListFilter {
 }
 
 impl PostListFilter {
-    /// Check if a cached post loosely matches this filter.
+    /// Check if a cached post loosely matches this filter, mimicking the
+    /// WordPress REST API's query parameter behavior.
     ///
     /// Conservative: returns `true` if the match cannot be determined locally.
     /// Only returns `false` when the post definitely does NOT match.
@@ -164,8 +165,18 @@ impl PostListFilter {
     /// and assumed to match.
     pub fn loosely_matches_post(&self, post: &wp_api::posts::AnyPostWithEditContext) -> bool {
         // Status check
-        if !self.status.is_empty() && !self.status.contains(&post.status) {
-            return false;
+        if !self.status.is_empty() {
+            if self.status.contains(&PostStatus::Custom("any".to_string())) {
+                // "any" excludes statuses with `exclude_from_search = true`.
+                // In WordPress core, `trash` and `auto-draft` have this flag set by default.
+                if post.status == PostStatus::Trash
+                    || post.status == PostStatus::Custom("auto-draft".to_string())
+                {
+                    return false;
+                }
+            } else if !self.status.contains(&post.status) {
+                return false;
+            }
         }
         // Author check
         if !self.author.is_empty()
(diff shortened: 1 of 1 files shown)


jkmassel (changes requested):
I'm wondering a few things about this:

1. Can we avoid using `PostStatus::Custom` for this? It seems like this is basically a `PostStatus::Default` as far as the API is concerned? 
2. It seems like `PostStatus::any == `PostStatus::Published && PostStatus::any == PostStatus::draft`, etc? 

This has me wondering if the type for `PostListFilter.status` should be `u64` and be backed by a bit set? That way you can represent a set of values without needing this extra custom type – you could just say "this should filter on PostListStatus::Draft|PostListStatus::Published"? This is a bit more ceremony, but it should be erased at runtime and I think more closely models what you're trying to do?

WDYT?

FWIW, the bot thinks there's some awkwardness:

> If status contains ["any", "trash"], the "any" branch wins and trash posts are excluded — contradicting the explicit "trash" filter. In practice, the app likely never constructs this combination, but the behavior is silently wrong if it does. A comment noting this assumption would help.

> The PR hardcodes trash and auto-draft as the only excluded statuses. Plugins can register custom statuses with exclude_from_search = true, which this won’t handle. 

I don't know if we support custom status values or not, so I don't know how that affects this PR. 

The bot also suggested more test coverage:

> Tests cover publish, draft, trash, and auto-draft with “any”. Missing:
>
> Pending, Private, Future statuses with “any” (should all match)
Empty status list (already covered by existing tests, just not with “any” context)

It also had one cool suggestion about avoiding allocation, but it might not be applicable if we end up changing anything:

> This method is called from update_post_membership in a loop over all live collections on every post change. The allocations can be avoided with zero-cost pattern matching:

```rust
if self.status.iter().any(|s| matches!(s, PostStatus::Custom(st) if st == "any")) {
    if post.status == PostStatus::Trash
        || matches!(&post.status, PostStatus::Custom(st) if st == "auto-draft")
    {
        return false;
    }
}
```

crazytonyli:
> Can we avoid using PostStatus::Custom for this? It seems like this is basically a PostStatus::Default as far as the API is concerned?

"any" is not a real post status. It's used as [a hard-coded magic string in post query](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-query.php#L2651). I'm not sure if we should declare it in `PostStatus`.

"Default" probably means different things in different endpoints. The default status in the wp/v2/posts endpoint is published.

> It seems like PostStatus::any == PostStatus::Published && PostStatus::any == PostStatus::draft`, etc?

Not entirely. This is what the bot says:

> 1. ?status=any — This passes any to WP_Query, which includes all statuses except those with exclude_from_search => true (which excludes trash and
  auto-draft). This is close but not identical — it may include statuses that show_in_admin_all_list is false for.

I'm using the "any" status to show posts in the iOS app's "All" tab. I want the apps' All tab to show the same data in the "All" tab in the wp-admin. Using status=any seems to be the easiest approach.

This is what the bot suggests if we really want to match the wp-admin:

<details>
<summary>🤖</summary>

The wp-admin "All" tab passes no post_status to WP_Query. When post_status is empty and is_admin is true, WP_Query includes posts with:
- Public statuses (e.g. publish)
- Protected statuses where show_in_admin_all_list is true (e.g. draft, pending, future)
- Private statuses (scoped by user capability)

It notably excludes trash.

So on the client side, you'd:

1. Fetch /wp/v2/statuses?context=edit (requires authentication — public, private, show_in_list are only in edit context)
2. Filter for statuses where public === true OR show_in_list === true OR private === true
3. Pass those slugs to your posts query: /wp/v2/posts?status[]=publish&status[]=draft&status[]=...

This dynamically picks up any custom statuses registered with those properties, matching the admin "All" tab behavior.

</details>

> This has me wondering if the type for PostListFilter.status should be u64 and be backed by a bit set?

Agreed. Probably don't need to be a bit set, as I explained above, "any" does not equal to a list of hard-coded status. I can update this PR to perform this refactor, if you think we should continue to use `status=any` to support the "All" tab, instead of the multiple-step process of providing an explicit list of status.

crazytonyli:
@jkmassel Do you think we can use this PR for now. And, if needed, we can refactor later?

> This has me wondering if the type for PostListFilter.status should be u64 and be backed by a bit set?

jkmassel (approved):
Approving to unblock – we can iterate further later.

PR #1211, opened 2026-03-02
Add a documentation about PostMetadataCollection

crazytonyli:
Add documentation about how to use PostMetadataCollection

### DIFF
--- a/Documentation/post-metadata-collection.md
+++ b/Documentation/post-metadata-collection.md
@@ -1,0 +1,583 @@
+# PostMetadataCollectionWithEditContext
+
+A mobile developer guide to displaying post lists using the wordpress-rs caching layer.
+
+## Contents
+
+1. [Overview](#1-overview)
+2. [Data Flow Diagram](#2-data-flow-diagram)
+3. [Creating a Collection](#3-creating-a-collection)
+4. [Observing Data Changes](#4-observing-data-changes)
+5. [Understanding Item States](#5-understanding-item-states)
+6. [Handling User Interactions](#6-handling-user-interactions)
+7. [Displaying Items in a List](#7-displaying-items-in-a-list)
+8. [Lifecycle Management](#8-lifecycle-management)
+9. [Types Reference](#9-types-reference)
+
+---
+
+## 1. Overview
+
+`PostMetadataCollectionWithEditContext` is the primary type for efficiently loading and displaying a paginated list of posts. It uses a **two-phase sync strategy**:
+
+1. **Phase 1 — Lightweight metadata:** Fetch only post IDs and modification timestamps from the API. This defines the list structure instantly.
+2. **Phase 2 — Selective full data:** Only fetch full post data for items that are missing from the cache or have changed since the last fetch.
+
+This means your list screen shows cached data immediately while only downloading what's actually needed.
+
+> **Key Concept:** The collection is backed by a local SQLite database. All data flows through the database, and the UI is notified of changes via **observer callbacks**. You never need to manually wire API responses to your UI — the observer pattern handles it automatically.
+
+---
+
+## 2. Data Flow Diagram
+
+The collection follows a **unidirectional data flow**: user actions trigger network fetches, results are written to the database, database changes notify observers, and observers tell your ViewModel to re-read from the cache. Data always flows in one direction through the system — your UI never receives API responses directly.
+
+```mermaid
+flowchart TD
+    subgraph APP["YOUR APP (UI Layer)"]
+        VM["ViewModel / Presenter"]
+        UI["List UI"]
+        LI["Loading / Error Indicators"]
+    end
+
+    subgraph WPRS["WORDPRESS-RS (Collection + Observers)"]
+        COL["PostMetadataCollection\nloadItems() · refresh() · loadNextPage()"]
+        API["WordPress REST API\nGET /wp/v2/posts\n_fields=id,modified_gmt (Phase 1)"]
+        DO["Data Observer"]
+        LIO["ListInfo Observer"]
+    end
+
+    subgraph DB["LOCAL DATABASE (SQLite)"]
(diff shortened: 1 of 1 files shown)


Documentation/post-metadata-collection.md
```diff
@@ -0,0 +1,583 @@
...
+### SyncResult
+
+Returned by `refresh()` and `loadNextPage()`.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `totalItems` | `UInt64` | Number of items in list after sync |
+| `fetchedCount` | `UInt64` | Items that were fetched from the API |
+| `failedCount` | `UInt64` | Items that failed to fetch |
+| `hasMorePages` | `Bool?` | Whether more pages are available |
```
jkmassel, line 566: Can we document the three value here? `true` and `false` are straightforward, but I assume `None` means we haven't fetched it yet?

I assume the same for the next two?

Documentation/post-metadata-collection.md
```diff
@@ -0,0 +1,583 @@
...
+| `search` | `String?` | Full-text search query |
+| `author` | `[UserId]` | Only show posts by specific authors |
+| `orderby` | `PostOrderBy?` | Sort field (date, title, modified, etc.) |
+| `order` | `Order?` | Sort direction (asc / desc) |
+
+> **One filter = one collection.** When the user switches filters (e.g. from "Published" to "Drafts"), close the old collection and create a new one. Each filter has its own independent cache.
+
+---
+
+## 4. Observing Data Changes
```
jkmassel, line 140: The inconsistency between Swift And Kotlin isn't ideal – I assume it's on the roadmap to address?

Documentation/post-metadata-collection.md
```diff
@@ -0,0 +1,583 @@
...
+What happens under the hood:
+
+1. `ListState` is set to `FetchingNextPage`
+2. Metadata for the next page is fetched and **appended** (not replaced) to the existing list
+3. Missing/stale items from the new page are fetched
+4. `ListState` is set to `Idle`
+
+### Changing Filters
+
+When the user switches between post lists (e.g. "Published" to "Drafts"), you need to close the old collection and create a new one.
```
jkmassel, line 346: Do I need to close the collection? Could I keep it around in case the user switches back?

Documentation/post-metadata-collection.md
```diff
@@ -0,0 +1,583 @@
...
+            when (val state = item.state) {
+                // States WITH data — show the post
+                is PostItemState.Fresh ->
+                    PostRow(post = state.data)
+                is PostItemState.Stale ->
+                    PostRow(post = state.data)
+                is PostItemState.FetchingWithData ->
+                    PostRow(post = state.data, isLoading = true)
+                is PostItemState.FailedWithData ->
+                    PostRow(post = state.data, error = state.error)
```
jkmassel, line 422: Could we combine these states using a `Result` type? (there may be no benefit, which is fine)

Documentation/post-metadata-collection.md
```diff
@@ -0,0 +1,583 @@
...
+### The Data Object
+
+When a state variant carries `data`, it is a `FullEntityAnyPostWithEditContext` containing:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `entityId` | `EntityId` | Database identity (site + type + WordPress ID) |
+| `data` | `AnyPostWithEditContext` | The full WordPress post object |
+
+Use `data.data.title`, `data.data.status`, `data.data.dateGmt`, etc. to access post fields for display.
```
jkmassel: nit: `data.data.${THING}` is a little weird – could we provide a more descriptive variant arg?

Documentation/post-metadata-collection.md
crazytonyli: The Kotlin library has a wrapper around the PostMetadataCollection. I haven't added one for Swift, because I want to see how it's used in practise, before wrapping it up. So, yes, we should have a nicer Swift API to bridge the raw Rust API and how Swift UI views present data. However, I'm not sure the Swift API will be consistent with Kotlin. I'd prioritize making it easier to work with in the context of SwiftUI over API consistency with Kotlin.

Documentation/post-metadata-collection.md
crazytonyli: I think that's a Kotlin wrapper API, and I'm not sure how it works. In Swift, you just discard the old instance as usual.

Documentation/post-metadata-collection.md
crazytonyli: My thinking is that this documentation is a high-level overview. For documentation about specific APIs, we should look into the binding code, which has more detailed info:

```swift
    /**
     * Whether there are more pages available.
     *
     * - `None` - Unknown (no metadata loaded or total_pages not provided by API)
     * - `Some(true)` - More pages available
     * - `Some(false)` - On last page
     */
    public let hasMorePages: Bool?
```

Documentation/post-metadata-collection.md
crazytonyli: Good catch. These are incorrect. Addressed in https://github.com/Automattic/wordpress-rs/pull/1211/commits/5fbc4c2d80f18c02e41785e86d209eeb314c130f

Documentation/post-metadata-collection.md
jkmassel: I'm happy to add more code to the bindings to document this stuff for sure

Documentation/post-metadata-collection.md
jkmassel: > I'd prioritize making it easier to work with in the context of SwiftUI over API consistency with Kotlin

Agreed! As long as this is still on your list, I'm good!

jkmassel (approved):
I think this is in a pretty good place and we could land it and iterate later

PR #1217, opened 2026-03-05
Add a migration script to ensure custom terms are re-fetched

crazytonyli:
## Description

The comment in the migration script should explain why it's needed.

The issue caused in the app was:
1. An older version of wprs stores posts without custom terms.
2. The new version after `additional_fields` is introduced in [the 0011 migration script](https://github.com/Automattic/wordpress-rs/blob/53a592f4642fe7a2d08c1415a6e0512bf2b1a5af/wp_mobile_cache/migrations/0011-add-additional-fields-to-posts-tables.sql) is able to store custom terms.
3. The app update to the newer version, which automatically sets the stored posts' `additional_fields` to null.
4. The posts will not be refreshed, because their state is "fresh". That means the app shows they don't have any custom terms, which may be incorrect.


jkmassel (approved):
Nice structure here reusing the existing schema setup for scripts too.

PR #1219, opened 2026-03-06
Add stats insights and stats summary endpoints

adalpari:
## Description

Add support for two WP.com REST API v1.1 stats endpoints:

1. **`GET /sites/<site_id>/stats/insights`** — Returns posting activity patterns including best hour/day to post, hourly view breakdowns, and yearly posting summaries.
2. **`GET /sites/<site_id>/stats`** — Returns aggregate site statistics (views, visitors, comments, posts, followers, etc.) and recent visit time-series data.

Both endpoints support an optional `locale` query parameter.

## Changes

- Add `StatsInsightsResponse`, `StatsInsightsYearData`, and `StatsInsightsParams` (with `locale` support)
- Add `StatsInsightsRequest` endpoint builder for `GET /sites/<site_id>/stats/insights`
- Add `StatsSummaryResponse`, `StatsSummaryStats`, and `StatsSummaryParams` (with `locale` support)
- Add `StatsSummaryRequest` endpoint builder for `GET /sites/<site_id>/stats`
- Reuse `StatsVisitsResponse` for the visits time-series data in stats summary
- Wire up both endpoints in `WpComApiRequestBuilder` and `WpComApiClient`
- Rewrite `deserialize_empty_array_or_hashmap` and `deserialize_option_empty_array_or_hashmap` to use `serde_json::Value` intermediate deserialization, fixing CI failures with non-map placeholder values
- Add JSON test fixtures and unit tests for both endpoints (11 stats insights tests, 4 stats summary tests)
- Add E2E tests for both endpoints in `wp_com_e2e`

## Test plan

- [x] Unit tests pass: `cargo test --lib -p wp_api -- stats_insights` (11 tests)
- [x] Unit tests pass: `cargo test --lib -p wp_api -- stats_summary` (4 tests)
- [x] No formatting issues: `cargo fmt --all -- --check`
- [x] No new clippy warnings
- [x] E2E test packages compile: `cargo check -p wp_com_e2e`
- [ ] E2E tests run against a live WP.com site


jkmassel (changes requested):
I don't think we should modify the general map visitor to be this broad. If you have a case where a map value can return `0`, we should define a new serde helper for it (something like `deserialize_hashmap_or_0_as_empty`).

Alternatively, that's a bug on the server and like with the date thing, we might just choose to make that impossible. We can't control WP.org code, but we can fix up WP.com to be more sensible :)

wp_api/src/wp_com/stats_summary.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Number of visitors yesterday.
+    pub visitors_yesterday: u64,
+    /// Total number of visitors.
+    pub visitors: u64,
+    /// Number of views today.
+    pub views_today: u64,
+    /// Number of views yesterday.
+    pub views_yesterday: u64,
+    /// The date with the most views (format: YYYY-MM-DD).
+    pub views_best_day: String,
```
jkmassel, line 48: What does the API return for a site with zero views?

wp_api/src/wp_com/stats_summary.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Total number of posts.
+    pub posts: u64,
+    /// Number of blog followers.
+    pub followers_blog: u64,
+    /// Number of comment followers.
+    pub followers_comments: u64,
+    /// Average comments per month.
+    pub comments_per_month: u64,
+    /// The most active recent day for comments.
+    pub comments_most_active_recent_day: String,
```
jkmassel: Should we interpret `""` as `None`?

wp_api/src/wp_com/stats_summary.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// The most active time for comments.
+    pub comments_most_active_time: String,
```
jkmassel: Should we assert "N/A" as `None`?

wp_api/tests/wpcom/stats_summary/response-01.json
```diff
@@ -0,0 +1,65 @@
+{
```
jkmassel, line 1: Can we have test fixtures for a site with zero stats?

wp_api/src/wp_com/stats_summary.rs
adalpari: Good question. I just created a new site with no views, and the endpoint is returning today as `views_best_day` 🤷‍♂️

wp_api/src/wp_com/stats_summary.rs
adalpari: I do believe so. Real response:
`"comments_most_active_recent_day": ""`

wp_api/src/wp_com/stats_summary.rs
adalpari: I think so:
`"comments_most_active_time": "N/A",`

wp_api/tests/wpcom/stats_summary/response-01.json
adalpari: Good call, on it!

adalpari:
> I don't think we should modify the general map visitor to be this broad. If you have a case where a map value can return `0`, we should define a new serde helper for it (something like `deserialize_hashmap_or_0_as_empty`).
> 
> Alternatively, that's a bug on the server and like with the date thing, we might just choose to make that impossible. We can't control WP.org code, but we can fix up WP.com to be more sensible :)

Thank you for the feedback. I've now reverted the changes in the map.rs file and created an individual deserializer for this specific case.
Also made some other changes from your other suggestions

wp_serde_helper/src/string.rs
```diff
@@ -44,6 +44,24 @@ where
     }
 }
 
+/// Deserialize a string, treating empty/whitespace-only strings and `"N/A"` as `None`.
+///
+/// Returns `None` if the string is empty, contains only whitespace, or equals `"N/A"`.
+/// Returns `Some(String)` for any other non-empty string (preserving the original value).
+pub fn deserialize_placeholder_string_as_none<'de, D>(
```
jkmassel, line 51: This should have some tests, please.

Specifically:
- "N/A" → None
- "" → None
- " " → None
- "some value" → Some("some value")

wp_api/src/wp_com/stats_summary.rs
jkmassel: What a weird default 😅 

I guess we'll go with it!

jkmassel (approved):
We should add some tests for the new serde helper, but then I think this is good to go, thanks!!

wp_serde_helper/src/string.rs
adalpari: Good call. Added some tests [here](https://github.com/Automattic/wordpress-rs/pull/1219/commits/99e004d094eebbf24689b3d11d4cb393098773e6)

PR #1220, opened 2026-03-09
Allow changing the `per_page` argument

crazytonyli:
## Description

After using the collection to show a paginated post list, the per_page argument can not be changed for the same filter, and the collection returns a `PerPageMismatch` error.

Of course, it's safe and reasonable to change the per_page. We can simply discard the stored data in the `list_metadata` table.

## Changes

Add a check at the beginning of the `refresh` call to check if the per_page value is different than the one stored in the database. Discard the database value if they are different.

### DIFF
--- a/wp_mobile/src/collection/post_metadata_collection.rs
+++ b/wp_mobile/src/collection/post_metadata_collection.rs
@@ -159,6 +159,13 @@ impl PostMetadataCollectionWithEditContext {
     pub async fn refresh(&self) -> Result<SyncResult, FetchError> {
         log::debug!("PostMetadataCollection: Refreshing collection");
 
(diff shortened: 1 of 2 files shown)


jkmassel:
When would we need to change the `per_page` for a collection? Does this refer to changing it on a live collection, or fetching with a collection then being unable to use a different page count in a new collection later?

crazytonyli:
> Does this refer to changing it on a live collection.

No, I don't think we ever need to dynamically change the value. It's almost always a hard-coded value. But we need to be able to change the hard-coded value, without breaking the meta collection. Right now, the per_page value is 20 in the iOS app. I want to change it to 100.

jkmassel (approved):

PR #1226, opened 2026-03-10
Fix loading trashed post by ids

crazytonyli:
## Description

Trashed post couldn't be loaded, because the "trash" status is not in the hard-coded list. This PR changes the `status` query value from an allowed list (publish,pending,private,draft,...) to just "any,trash". The "any,trash" should include custom post status too.

[Brief description of what this PR changes and why]

### DIFF
--- a/wp_mobile/src/service/posts.rs
+++ b/wp_mobile/src/service/posts.rs
@@ -523,7 +523,7 @@ impl PostService {
             // Use "any" to match all post statuses (including custom ones), plus
             // "trash" which WordPress excludes from "any" because it has
             // `internal: true` (and therefore `exclude_from_search: true`).
-            status: vec![PostStatus::Trash, PostStatus::Custom("any".to_string())],
+            status: vec![PostStatus::Trash, PostStatus::Any],
             ..Default::default()
         };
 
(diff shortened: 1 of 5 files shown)


wp_mobile/src/service/posts.rs
```diff
@@ -520,18 +520,10 @@ impl PostService {
...
-                PostStatus::Publish,
-                PostStatus::Draft,
-                PostStatus::Pending,
-                PostStatus::Private,
-                PostStatus::Future,
-            ],
+            // Use "any" to match all post statuses (including custom ones), plus
+            // "trash" which WordPress excludes from "any" because it has
+            // `internal: true` (and therefore `exclude_from_search: true`).
+            status: vec![PostStatus::Trash, PostStatus::Custom("any".to_string())],
```
jkmassel: Given the fact that we're using this internally and it's officially supported by the API, should we promote `Any` to its own first-class status?

jkmassel (approved):

wp_mobile/src/service/posts.rs
crazytonyli: 👍 Updated in https://github.com/Automattic/wordpress-rs/pull/1226/commits/1009e6c4e1447065b7e6136b8817a89c1d2ac3d4.

PR #1232, opened 2026-03-16
Add new APIs for deleting site data from WpApiCache

crazytonyli:
## Description

The new APIs can be used when a site is removed from the mobile apps.

### DIFF
--- a/wp_mobile_cache/src/lib.rs
+++ b/wp_mobile_cache/src/lib.rs
@@ -1,10 +1,13 @@
 use rusqlite::hooks::Action;
 use rusqlite::types::{FromSql, FromSqlResult, ToSql, ToSqlOutput};
 use rusqlite::{Connection, Result as SqliteResult, params};
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
 
 use crate::repository::entity_state::EntityStateRepository;
 use crate::repository::list_metadata::ListMetadataRepository;
+use crate::repository::sites::SiteRepository;
+use wp_api::parsed_url::ParsedUrl;
+use wp_api::wp_com::WpComSiteId;
 
 pub mod context;
 pub mod db_types;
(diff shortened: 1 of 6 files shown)


wp_mobile_cache/src/lib.rs
```diff
@@ -574,6 +598,17 @@ impl DBManager {
     }
 }
 
+fn url_candidates(url: &ParsedUrl) -> [String; 2] {
+    let s = url.url();
+    if s.ends_with('/') {
+        let stripped = s[..s.len() - 1].to_string();
```
oguzkocer: I think we can use [`String::strip_suffix`](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_suffix) here.

wp_mobile_integration_tests/tests/test_remove_site.rs
```diff
@@ -0,0 +1,198 @@
...
+use wp_api::parsed_url::ParsedUrl;
+use wp_api::request::endpoint::posts_endpoint::PostEndpointType;
+use wp_api::wp_com::WpComSiteId;
+use wp_mobile::filters::PostListFilter;
+use wp_mobile_cache::WpApiCache;
+use wp_mobile_cache::repository::sites::SiteRepository;
+use wp_mobile_integration_tests::*;
+
+/// Helper to count rows in a table filtered by db_site_id.
+fn count_rows_for_site(cache: &WpApiCache, table: &str, db_site_id: i64) -> usize {
```
oguzkocer: We might want this to be in `wp_mobile_integration_tests/src/lib.rs` instead.

oguzkocer (approved):
I shared a couple minor suggestions, but otherwise it looks good to me.

wp_mobile_cache/src/lib.rs
crazytonyli: Thanks for the suggestion. Addressed in https://github.com/Automattic/wordpress-rs/pull/1232/commits/dee1baf02ab55e4c2ef745c413d50df0c878bbcd.

PR #1240, opened 2026-03-18
Replace SwiftLint with swift-format

crazytonyli:
## Description

Xcode comes with the swift-format command. It's easy and super fast to run. Another factor is that I plan to add a Claude Code hook to format Swift code after edits. Using swiftlint for that means you may need to fetch the swiftlint package, which is annoying.

I suggest reviewing commit by commit. And, the last commit can be ignored.


jkmassel (approved):
Nice to use built-in tools.

Also seems super fast?

crazytonyli:
Yes, it's super fast. Probably because of the `--parallel` option.

PR #1250, opened 2026-03-25
Implement the Jetpack Social Publicize endpoints in the wpcom/v2 namespace

crazytonyli:
## Description

Relates to https://linear.app/a8c/issue/CMM-274.

There are lots of changes in this PR, because I had to build the client side to verify the endpoints implementation actually works in practise.

Here are the main changes:

1. Refactor parsing `PostMeta`
Previously, only footnotes were parsed. This PR now makes `PostMeta` parse everything in the `meta`. That's because the Jetpack Social custom message is in the meta field. `PostMeta` now have APIs for reading and adding arbitrary values.

2. Refactor `additional_fields` properties
A new type WpAdditionalFields is created, so that we can add similar helpers as PostMeta to add arbitrary values to it. This is used to support adding Jetpack Social properties to the `PostUpdateParams`.

3. The endpoints implementation
The new connection endpoints in wpcom/v2 are implemented in this PR. The apps are using the ones in rest/v1.1, which are deprecated.


wp_api/src/posts.rs
```diff
@@ -493,12 +496,88 @@ pub struct SparsePostExcerpt {
...
+    /// `footnotes` key in the WordPress wire format (a JSON string holding
+    /// a JSON-encoded array). The receiver `Arc` is not mutated; callers
+    /// chain: `PostMeta::new().with_footnotes(vec![...])`.
+    pub fn with_footnotes(self: Arc<Self>, footnotes: Vec<PostFootnote>) -> Arc<Self> {
+        let mut new_inner = self.inner.clone();
+        // FIXME: `Vec<PostFootnote>` is two String fields per element so
+        // serialization is infallible in practice; if `PostFootnote` ever
+        // grows a fallible Serialize impl this expect() must be revisited.
+        let encoded =
+            serde_json::to_string(&footnotes).expect("Vec<PostFootnote> is always serializable");
```
oguzkocer: I am really not a fan of the `expect` or the `FIXME` comment. Can we return a `Result` instead?At the very least, I think the `FIXME` should go. If it's not a problem, it's not a `FIXME`, if it's a problem, it should be addressed before we merge it.

To my knowledge we don't use `unwrap/expect` in the library code, so my vote would be to go with `Result`, but if we want to keep it contained, we can add a unit test to ensure it's always serializable and add the comment about this being an important check to the test instead of this call site. In that case, it won't be a `FIXME`, but a reminder if anyone wants to remove the unit test. What do you think?

oguzkocer (approved):
I raised an issue, but I don't want to block the PR, so will leave the decision to you.

wp_api/src/posts.rs
crazytonyli: I have changed the function to return a `Result`: https://github.com/Automattic/wordpress-rs/pull/1250/commits/40e9ea3d8d9529f2315f978511cf85e89730cf0a

PR #1253, opened 2026-03-26
Add stats UTM endpoint

adalpari:
## Description

Add a new WP.com REST API endpoint for UTM stats at `/sites/{site_id}/stats/utm/{utm_keys}`. This follows the existing stats endpoint patterns in the codebase.

## Changes

- Add `StatsUtmKey` enum supporting `utm_source`, `utm_medium`, and `utm_campaign`
- Add `StatsUtmKeys` newtype for comma-separated UTM keys used as a URL path segment
- Add `StatsUtmParams` with support for `max`, `date`, `days`, `start_date`, `post_id`, and `query_top_posts` query parameters
- Add `StatsUtmResponse`, `StatsUtmPost` response types with `top_utm_values` and `top_posts` fields
- Handle empty array/object responses via `deserialize_empty_array_or_hashmap`
- Register the endpoint in `client.rs`, `endpoint.rs`, and `mod.rs`

## Test plan

- [x] Unit tests for UTM key serialization (single, multiple, all keys)
- [x] Unit tests for query parameter serialization (full, minimal, query_top_posts=false)
- [x] Deserialization tests for single-key, multiple-keys, triple-keys, empty, and empty-arrays JSON fixtures


wp_api/src/wp_com/stats_utm.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub date: Option<String>,
+    /// The number of days to include in the query.
+    #[uniffi(default = None)]
+    pub days: Option<u32>,
+    /// The start date to query stats for (format: YYYY-MM-DD).
+    #[uniffi(default = None)]
+    pub start_date: Option<String>,
+    /// A specific post ID to filter results.
+    #[uniffi(default = None)]
+    pub post_id: Option<String>,
```
jkmassel: I'm surprised this is a string instead of a `u64`?

wp_api/src/wp_com/stats_utm.rs
```diff
@@ -0,0 +1,288 @@
...
+uniffi::custom_newtype!(StatsUtmKeys, String);
+/// Comma-separated UTM keys used as a URL path segment.
+///
+/// For example, `"utm_source,utm_medium"` or `"utm_campaign"`.
+/// Construct from a list of `StatsUtmKey` using `StatsUtmKeys::new`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StatsUtmKeys(pub String);
+
+impl StatsUtmKeys {
+    pub fn new(keys: &[StatsUtmKey]) -> Self {
```
jkmassel: If I call this with an empty vec, I get a 404.

IMHO a `Params` object shouldn't be able to accept a set of arguments that leads to this, so we should probably fix it.

Easy suggestions:

1. If empty, panic.
2. If empty, assume that means "use all variants of `StatsUtmKey`"

jkmassel (changes requested):

wp_api/src/wp_com/stats_utm.rs
adalpari: Good catch. Actually, I think we can remove this argument. I'm reviewing the calls on web again, and it always seems to be empty

wp_api/src/wp_com/stats_utm.rs
adalpari: Thanks for the check. Done [here](https://github.com/Automattic/wordpress-rs/pull/1253/changes/2cdcba958f7646e711e9467c7f303282380f2946)!

wp_api/tests/wpcom/stats_utm/missing-top-posts.json
```diff
@@ -0,0 +1 @@
+{"top_utm_values":[]}
```
jkmassel: nit: this file is missing a trailing newline

wp_api/src/wp_com/stats_utm.rs
```diff
@@ -0,0 +1,309 @@
...
+///
+/// For example, `"utm_source,utm_medium"` or `"utm_campaign"`.
+/// Construct from a list of `StatsUtmKey` using `StatsUtmKeys::new`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StatsUtmKeys(pub String);
+
+impl StatsUtmKeys {
+    pub fn new(keys: &[StatsUtmKey]) -> Result<Self, StatsUtmError> {
+        if keys.is_empty() {
+            return Err(StatsUtmError::EmptyUtmKeys);
```
jkmassel: Keys being empty is a programmer mistake, which IMHO should be an assertion instead of an error.

jkmassel (changes requested):

adalpari:
Sounds reasonable!
Changes [done](https://github.com/Automattic/wordpress-rs/pull/1253/commits/696358c976c8e8e241b22ef26ae5ef96068ada06)!

wp_api/src/wp_com/stats_utm.rs
```diff
@@ -0,0 +1,300 @@
...
+/// Comma-separated UTM keys used as a URL path segment.
+///
+/// For example, `"utm_source,utm_medium"` or `"utm_campaign"`.
+/// Construct from a list of `StatsUtmKey` using `StatsUtmKeys::new`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StatsUtmKeys(pub String);
+
+impl StatsUtmKeys {
+    pub fn new(keys: &[StatsUtmKey]) -> Self {
+        assert!(!keys.is_empty(), "At least one UTM key must be provided");
```
oguzkocer, line 43: I think this was added to address @jkmassel's feedback [here](https://github.com/Automattic/wordpress-rs/pull/1253/changes/2cdcba958f7646e711e9467c7f303282380f2946#r3049211256).

> Keys being empty is a programmer mistake, which IMHO should be an assertion instead of an error.

But the assertion still makes it a runtime error. So, I don't agree with the change. In my opinion, the assertions shouldn't be part of the public facing API. If this was WPiOS or WPAndroid, I can understand, although that's still questionable for a network call, because there is no need to crash the app. I suggest going back to 2cdcba958f7646e711e9467c7f303282380f2946 and return a `Result`.

If we really want to avoid it, we can use a builder-like pattern where the client adds the keys one by one.

---

There is this approach as well mentioned by Claude:

```
pub fn new(first: StatsUtmKey, rest: &[StatsUtmKey]) -> Self
```

I think it's a very awkward API and Claude agreed. Still including it for completeness.

oguzkocer (commented):
I added my 2 cents about the assertion for building `StatsUtmKeys`. I am leaving this review as a comment so it doesn't block you. I leave the final decision to Jeremy and you.

jkmassel (approved):
I chatted with @oguzkocer about a better approach for the CSV URL segment. We'll land this PR and do a fix that builds on it in `trunk`.

PR #1262, opened 2026-04-01
Group Kotlin, Compose, and Koin in Dependabot

jkmassel:
## Summary

- Groups `org.jetbrains.kotlin*`, `org.jetbrains.compose`, and `io.insert-koin:*` into a single Dependabot PR
- These have tight version coupling (e.g. Koin 4.2.0 requires Kotlin 2.3+, which requires Compose 1.11+)
- Prevents orphaned PRs like #1242 where Koin was bumped alone and broke the build

## Test plan

- [ ] Verify Dependabot creates grouped PRs on next schedule run

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,12 @@ updates:
     directory: "native/kotlin" # Location of package manifests
     schedule:
       interval: "weekly"
+    groups:
+      kotlin-ecosystem:
+        patterns:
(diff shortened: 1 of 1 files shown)


adalpari:
Test oplan says:

> Verify Dependabot creates grouped PRs on next schedule run

But I don't think that can be verified until the PR is merged, can it?

Otherwise, it looks good to me

adalpari (approved):

PR #1263, opened 2026-04-01
Add UserCapabilitiesMap to handle non-boolean capability values

kean:
## Description

Fix [CMM-2013: Issue decoding site capabilities with non-boolean values](https://linear.app/a8c/issue/CMM-2013/issue-decoding-site-capabilities-with-non-boolean-values).

## Changes

  - Add `UserCapabilitiesMap` type that stores capabilities as `HashMap<UserCapability, JsonValue>` instead of
  `HashMap<UserCapability, bool>`, fixing deserialization failures when plugins (e.g., WPBakery) add capabilities with
   non-boolean values like `"custom"` or `"any"`
  - Add `has_cap(capability)` method exported via UniFFI that returns `true` only for boolean `true` values
  - Update `WPComSite.capabilities` and `SparseUser.capabilities` to use the new type

I used the suggested naming. The usage from Swift should presumably look like this.

```swift
 // Check a known capability
  site.capabilities.hasCap(capability: .editPosts)    // true
  site.capabilities.hasCap(capability: .editUsers)     // false

  // WPBakery-style string value → false
  site.capabilities.hasCap(capability: .custom("vc_access_rules_post_types")) // false

  // Raw map access still available
  site.capabilities.map  // [UserCapability: JsonValue]
```

## Test plan

The change is covered with a unit tests that adds a `WPBakery` capability key under question.

### DIFF
--- a/wp_api/tests/wpcom/sites/v1.2-me-sites-01.json
+++ b/wp_api/tests/wpcom/sites/v1.2-me-sites-01.json
@@ -31,8 +31,7 @@
           "activate_plugins": true,
           "update_plugins": false,
           "export": true,
-          "import": true,
-          "vc_access_rules_post_types": "custom"
+          "import": true
         },
         "jetpack": false,
         "jetpack_connection": false,
--- a/wp_api/tests/wpcom/sites/v1.2-me-sites-01.json
+++ b/wp_api/tests/wpcom/sites/v1.2-me-sites-01.json
@@ -31,7 +31,8 @@
           "activate_plugins": true,
           "update_plugins": false,
           "export": true,
-          "import": true
+          "import": true,
+          "vc_access_rules_post_types": "custom"
         },
         "jetpack": false,
         "jetpack_connection": false,
--- a/wp_api/src/users.rs
+++ b/wp_api/src/users.rs
@@ -1,5 +1,5 @@
 use crate::{
-    EnumFromStrParsingError, OptionFromStr, WpApiParamOrder, WpResponseString,
+    EnumFromStrParsingError, JsonValue, OptionFromStr, WpApiParamOrder, WpResponseString,
     impl_as_query_value_from_to_string,
     url_query::{
         AppendUrlQueryPairs, FromUrlQueryPairs, QueryPairs, QueryPairsExtension, UrlQueryPairsMap,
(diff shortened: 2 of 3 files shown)


wp_api/tests/wpcom/sites/v1.2-me-sites-01.json
```diff
@@ -31,7 +31,8 @@
           "activate_plugins": true,
           "update_plugins": false,
           "export": true,
-          "import": true
+          "import": true,
+          "vc_access_rules_post_types": "custom"
```
jkmassel: This isn't a WP.com thing, it's a plugin thing. Can we move this into a new test fixture inside of `wp_api` instead?

kean: Is there a particular place or existing mock in `wp_api/tests` that you have in mind? Dotcom sites do have plugins, which is why I added it to an exiting file.

I think I actually put under the wrong request. The error from the logs doesn't indicate what request fails, as a single error is used for multiple operations:

```swift
        do {
            // site settings is only available to admin users. Ignore errors for now,
            // since we need to allow other users to sign in to the app too.
            async let siteSettings_ = try? api.siteSettings.retrieveWithViewContext().data
            async let isAdmin_ = api.users.retrieveMeWithEditContext().data.roles.contains(.administrator)
            async let jetpackSite_ = fetchJetpackSite(apiRootURL: apiRootURL, credentials: credentials)
            async let jetpackConnection_ = fetchJetpackConnectionData(apiRootURL: apiRootURL, credentials: credentials)
            async let xmlrpcOptions_ = try? loadSiteOptions(xmlrpc: xmlrpc, details: credentials)

            (siteSettings, isAdmin, jetpackSite, jetpackConnection, xmlrpcOptions) =
                try await (siteSettings_, isAdmin_, jetpackSite_, jetpackConnection_, xmlrpcOptions_)
        } catch {
            throw .loadingSiteInfoFailure(error)
        }
```

However, based on the response, it seems that what fails is actually `retrieveMeWithEditContext`  (based on presence of `first_name` in the logged JSON)

If I'm reading it correctly, it seems it maps to `/users/me` from the Dotorg REST API. I don't see any existing tests or mocks for it. 

Should I add new tests for `/users/me`?

kean: I adhered to the existing patterns in users.rs and added a targeted unit test for this fix. Since there are no higher-level JSON fixtures currently, I suggest keeping those changes for later to focus on this bug resolution.

oguzkocer (approved):

PR #1271, opened 2026-04-08
Remove conflicting serde attributes from single-field structs

oguzkocer:
## Summary

Removed `#[serde(transparent)]` and conflicting `#[serde(rename)]` attributes from single-field wrapper structs.

## Changes

- Removed `#[serde(transparent)]` from 6 structs:
  - `UserCapabilitiesMap`
  - `PostTypeSupportsMap`
  - `SparsePostTypesResponse`
  - `SparseTaxonomyTypesResponse`
  - `SparseMenuLocationsResponse`
  - `SparsePostStatusesResponse`

- Removed `#[serde(rename)]` from `UserCapabilitiesMap` and `PostTypeSupportsMap`

## Rationale

For single-field structs, `#[serde(flatten)]` on the field is functionally equivalent to `#[serde(transparent)]` on the struct.

`#[serde(transparent)]` is not supported by:
- `WpContextual` (generates concrete types without `transparent`)
- `WpDeserialize` (explicitly rejects `transparent` attribute)

`#[serde(rename)]` conflicts with `#[serde(flatten)]` which takes all keys from the parent object.

### DIFF
--- a/wp_api/src/menu_locations.rs
+++ b/wp_api/src/menu_locations.rs
@@ -6,7 +6,6 @@ use wp_contextual::WpContextual;
 wp_content_string_id!(MenuLocation);
 
 #[derive(Debug, Serialize, Deserialize, uniffi::Record, WpContextual)]
-#[serde(transparent)]
 pub struct SparseMenuLocationsResponse {
     #[serde(flatten)]
     #[WpContext(edit, embed, view)]
--- a/wp_api/src/post_statuses.rs
+++ b/wp_api/src/post_statuses.rs
@@ -3,7 +3,6 @@ use std::{collections::HashMap, fmt::Display};
 use wp_contextual::WpContextual;
 
 #[derive(Debug, Serialize, Deserialize, uniffi::Record, WpContextual)]
-#[serde(transparent)]
 pub struct SparsePostStatusesResponse {
     #[serde(flatten)]
     #[WpContext(edit, embed, view)]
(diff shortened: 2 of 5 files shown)


jkmassel (changes requested):
I think this approach _could_ fail (though it's probably unlikely).

For example:

```rust

// wp_api/src/post_types.rs — inside mod test {}
#[test]
fn test_post_type_supports_map_from_empty_array() {
    // WordPress returns `[]` instead of `{}` when a post type has no
    // supported features (e.g. `register_post_type('foo', ['supports' => false])`).
    let json = r#"[]"#;
    let supports: PostTypeSupportsMap =
        serde_json::from_str(json).expect("Should handle empty array from WordPress API");
    assert!(supports.map.is_empty());
}

// wp_api/src/users.rs — inside mod tests {}
#[test]
fn test_user_capabilities_map_from_empty_array() {
    // WordPress returns `[]` instead of `{}` when a map is empty.
    let json = r#"[]"#;
    let caps: UserCapabilitiesMap =
        serde_json::from_str(json).expect("Should handle empty array from WordPress API");
    assert!(caps.map.is_empty());
}

```

It's probably vanishingly rare, but I think we need to handle the empty case?

oguzkocer:
@jkmassel Thank you for catching that. I'm working on a different way to address this conflict, so I'll close this PR and open a new one when I am ready. I'll also make sure to add the tests you provided so they are part of our coverage.

oguzkocer:
https://github.com/Automattic/wordpress-rs/pull/1273 should address the issues.

PR #1274, opened 2026-04-08
Change `StatsUtmKeys` to `Vec<StatsUtmKey>` newtype

oguzkocer:
## Description

Follow-up to #1253. Changes `StatsUtmKeys` from a `String` newtype to a `Vec<StatsUtmKey>` newtype so callers work with typed keys instead of a pre-joined string. An empty vec now defaults to all UTM keys instead of panicking.

## Changes

- Change `StatsUtmKeys` inner type from `String` to `Vec<StatsUtmKey>`
- Remove `StatsUtmKeys::new` constructor in favor of direct construction
- Add `EnumIter` to `StatsUtmKey`; empty vec in `Display` falls back to all variants
- Add shared WP.com endpoint test helpers (`fixture_wp_com_api_url_resolver`, `validate_wp_com_rest_v1_1_endpoint`)
- Add endpoint-level unit tests for `stats_utm_endpoint`

### DIFF
--- a/wp_api/src/wp_com/endpoint.rs
+++ b/wp_api/src/wp_com/endpoint.rs
@@ -130,14 +130,34 @@ impl ApiUrlResolver for WpComApiClientInternalUrlResolver {
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use super::*;
-    use rstest::rstest;
+    use crate::{request::endpoint::ApiEndpointUrl, wp_com::WpComNamespace};
+    use rstest::*;
+    use std::sync::Arc;
+
+    const WP_COM_BASE_URL: &str = "https://public-api.wordpress.com";
+
+    #[fixture]
+    pub fn fixture_wp_com_api_url_resolver() -> Arc<dyn ApiUrlResolver> {
+        Arc::new(WpComApiClientInternalUrlResolver::default())
+    }
+
+    pub fn validate_wp_com_rest_v1_1_endpoint(endpoint_url: ApiEndpointUrl, path: &str) {
+        validate_endpoint(WpComNamespace::RestV1_1, endpoint_url, path);
+    }
+
+    fn validate_endpoint(namespace: WpComNamespace, endpoint_url: ApiEndpointUrl, path: &str) {
+        assert_eq!(
+            endpoint_url.as_str(),
+            format!("{}{}{}", WP_COM_BASE_URL, namespace.namespace_value(), path)
+        );
+    }
 
     #[rstest]
-    #[case("/wp/v2", vec!["posts".to_string()], "https://public-api.wordpress.com/wp/v2/sites/example.wordpress.com/posts")]
+    #[case(WpNamespace::WpV2, vec!["posts".to_string()], "https://public-api.wordpress.com/wp/v2/sites/example.wordpress.com/posts")]
     fn wp_com_dot_org_api_url_resolver(
-        #[case] namespace: &str,
+        #[case] namespace: WpNamespace,
         #[case] endpoint_segments: Vec<String>,
         #[case] expected_url: &str,
     ) {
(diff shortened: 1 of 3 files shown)


wp_api/src/wp_com/stats_utm.rs
```diff
@@ -30,29 +32,21 @@ pub enum StatsUtmKey {
...
 
 impl std::fmt::Display for StatsUtmKeys {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        let keys = if self.0.is_empty() {
+            StatsUtmKey::iter().collect()
+        } else {
+            self.0.clone()
+        };
+        let joined: Vec<_> = keys.iter().map(|k| k.to_string()).collect();
```
jkmassel, line 51: Should we deduplicate in case the same entry appears twice?

wp_api/src/wp_com/stats_utm.rs
```diff
@@ -30,29 +32,21 @@ pub enum StatsUtmKey {
...
-}
+pub struct StatsUtmKeys(pub Vec<StatsUtmKey>);
 
 impl std::fmt::Display for StatsUtmKeys {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        let keys = if self.0.is_empty() {
+            StatsUtmKey::iter().collect()
+        } else {
+            self.0.clone()
```
jkmassel, line 46:
```suggestion
            &self.0
```

Can we borrow instead?

wp_api/src/wp_com/endpoint.rs
```diff
@@ -130,14 +130,34 @@ impl ApiUrlResolver for WpComApiClientInternalUrlResolver {
...
+    fn validate_endpoint(namespace: WpComNamespace, endpoint_url: ApiEndpointUrl, path: &str) {
+        assert_eq!(
+            endpoint_url.as_str(),
+            format!("{}{}{}", WP_COM_BASE_URL, namespace.namespace_value(), path)
+        );
+    }
 
     #[rstest]
-    #[case("/wp/v2", vec!["posts".to_string()], "https://public-api.wordpress.com/wp/v2/sites/example.wordpress.com/posts")]
+    #[case(WpNamespace::WpV2, vec!["posts".to_string()], "https://public-api.wordpress.com/wp/v2/sites/example.wordpress.com/posts")]
```
jkmassel: I'd generally prefer matching plain strings to constants here, but in this context it might be a nit.

wp_api/src/wp_com/endpoint.rs
```diff
@@ -130,14 +130,34 @@ impl ApiUrlResolver for WpComApiClientInternalUrlResolver {
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
```
jkmassel, line 133:
```suggestion
mod tests {
```

Do we need `pub(crate)` here?

jkmassel (changes requested):
Left some nits, I think the only really important one is the deduplication

wp_api/src/wp_com/endpoint.rs
oguzkocer: Yes, because `validate_wp_com_rest_v1_1_endpoint` helper method can't be accessed from other modules without it.

wp_api/src/wp_com/endpoint.rs
oguzkocer: Addressed in c40709d371297e78063ed1586629aa18b410b632.

wp_api/src/wp_com/stats_utm.rs
oguzkocer: Addressed in c40709d371297e78063ed1586629aa18b410b632.

wp_api/src/wp_com/stats_utm.rs
oguzkocer: We can, but we need a `Vec` to deduplicate, so I skipped this one. Well, not "need", but it's the simplest way to implement it. For a `Vec` with 3 enums, simpler should be better.

jkmassel (approved):

PR #1276, opened 2026-04-11
Add domain suggestions and segments endpoints

oguzkocer:
Adds `GET /domains/suggestions` and `GET /wpcom/v2/segments` with their response types, a `SegmentId` newtype, and JSON fixture tests.

Note that `tlds` is serialized as a PHP-style array (`tlds[]=...`) since the API silently ignores CSV and honors only the last value for repeated pairs.

### DIFF
--- a/wp_api/src/wp_com/domains.rs
+++ b/wp_api/src/wp_com/domains.rs
@@ -1,0 +1,303 @@
+use crate::{
+    url_query::{AppendUrlQueryPairs, QueryPairs, QueryPairsExtension},
+    wp_com::segments::SegmentId,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, uniffi::Record)]
+pub struct DomainSuggestionsParams {
+    /// The search term used to generate domain suggestions.
+    pub query: String,
+    /// The maximum number of suggestions to return.
+    #[uniffi(default = None)]
+    pub quantity: Option<u32>,
+    /// Restrict suggestions to the given TLDs (e.g. `["com", "net"]`).
+    ///
+    /// Must be serialized as PHP-style array parameters
+    /// (`tlds[]=com&tlds[]=net`). The API's PHP backend silently ignores
+    /// CSV (`tlds=com,net`) and honors only the last value when repeated
+    /// pairs (`tlds=com&tlds=net`) are used.
+    #[uniffi(default = None)]
+    pub tlds: Option<Vec<String>>,
+    /// Restrict suggestions to a specific vendor (e.g. `"dot"`).
+    #[uniffi(default = None)]
+    pub vendor: Option<String>,
+    /// If `true`, only return WordPress.com subdomain suggestions.
+    #[uniffi(default = None)]
+    pub only_wordpressdotcom: Option<bool>,
+    /// If `true`, include WordPress.com subdomain suggestions alongside
+    /// the regular domain suggestions.
+    #[uniffi(default = None)]
+    pub include_wordpressdotcom: Option<bool>,
+    /// If `true`, include `*.home.blog` subdomain suggestions.
+    #[uniffi(default = None)]
+    pub include_dotblogsubdomain: Option<bool>,
+    /// Segment identifier used by the API to tailor suggestions.
+    ///
+    /// Fetch available segments via the `/wpcom/v2/segments` endpoint.
+    /// Note: sending a valid `segment_id` pivots results toward free
+    /// `*.wordpress.com` subdomains (and curated themed `.blog`
+    /// subdomains for the Blog segment).
+    #[uniffi(default = None)]
+    pub segment_id: Option<SegmentId>,
+}
+
+impl AppendUrlQueryPairs for DomainSuggestionsParams {
+    fn append_query_pairs(&self, query_pairs_mut: &mut QueryPairs) {
+        query_pairs_mut
+            .append_query_value_pair("query", &self.query)
+            .append_option_query_value_pair("quantity", self.quantity.as_ref())
+            .append_option_query_value_pair("vendor", self.vendor.as_ref())
+            .append_option_query_value_pair(
+                "only_wordpressdotcom",
+                self.only_wordpressdotcom.as_ref(),
+            )
+            .append_option_query_value_pair(
+                "include_wordpressdotcom",
+                self.include_wordpressdotcom.as_ref(),
+            )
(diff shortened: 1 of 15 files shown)


wp_api/src/wp_com/domains.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Formatted renewal cost (e.g. `"$18.00"`).
+    pub renew_cost: String,
+    /// Raw numeric renewal price in `currency_code`.
+    pub renew_raw_price: f64,
+    /// Raw numeric registration price in `currency_code`.
+    pub raw_price: f64,
+    /// ISO 4217 currency code for `raw_price`/`renew_raw_price` (e.g. `"USD"`).
+    pub currency_code: String,
+    /// Promotional sale price in `currency_code`, if the domain is on sale.
+    pub sale_cost: Option<f64>,
```
jkmassel, line 131: It's unfortunate that this is a float type (which famously suffers from "fun" bugs when used for Money). It's probably fine for pretty much every value that a domain would sell for within the app, but consider the values `raw_price: 11` and `sale_cost: 8.63` from `basic-query.json – if you were trying to say "Save $2.37 off" by calculating `raw_price / sale_cost` you'd end up with `11.0-8.63 != 2.37`.

Might be something for a follow-up PR, but implementing a uniffi-compatible `Decimal` type (or even just `Money`, which is dramatically simpler) would probably be worthwhile to ensure we don't get strange cross-platform rounding/conversion differences).

wp_api/src/wp_com/endpoint/domains_endpoint.rs
```diff
@@ -0,0 +1,96 @@
...
+    #[case::include_wordpresscom(
+        DomainSuggestionsParams { include_wordpresscom: Some(true), ..base_domain_suggestions_params() },
+        "/domains/suggestions?query=coolsite&quantity=5&include_wordpresscom=true"
+    )]
+    #[case::vendor(
+        DomainSuggestionsParams { vendor: Some("dot".to_string()), ..base_domain_suggestions_params() },
+        "/domains/suggestions?query=coolsite&quantity=5&vendor=dot"
+    )]
+    #[case::include_dotblogsubdomain(
+        DomainSuggestionsParams { include_dotblogsubdomain: Some(true), ..base_domain_suggestions_params() },
```
jkmassel, line 62: Should we have a test for `Some(false)`?

It's not clear to me whether `false` implies `None`

wp_api/src/wp_com/domains.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// `true` if the TLD requires HSTS (e.g. `.dev`).
+    pub hsts_required: Option<bool>,
+    /// Policy notices attached to the suggestion (e.g. HSTS warnings).
+    pub policy_notices: Option<Vec<DomainPolicyNotice>>,
```
jkmassel: I don't know if we ever settled on a convention here, but I wonder about the optionality here – should we just default to an empty `vec` here if there are no values?

wp_api/src/wp_com/domains.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Raw numeric renewal price in `currency_code`.
+    pub renew_raw_price: f64,
+    /// Raw numeric registration price in `currency_code`.
+    pub raw_price: f64,
+    /// ISO 4217 currency code for `raw_price`/`renew_raw_price` (e.g. `"USD"`).
+    pub currency_code: String,
+    /// Promotional sale price in `currency_code`, if the domain is on sale.
+    pub sale_cost: Option<f64>,
+    /// `true` if the TLD requires HSTS (e.g. `.dev`).
+    pub hsts_required: Option<bool>,
```
jkmassel, line 133: I'd imagine this will only ever be `Some(true)`, _never_ `Some(false`).

Given this, would it make sense to have this default to `false` unless present and `Some(true)`?

jkmassel (changes requested):
Some of the test fixture files seem to be identical – not sure if intentional or not.

A couple of questions about tests.

Requesting changes to avoid auto-merge, but I think this is generally ready to go

wp_api/src/wp_com/domains.rs
oguzkocer: If we know that the API specifically uses empty vecs for it, we can. However, if it doesn't, or we don't know I don't we should.

There is a semantic difference between `None` and `[]` and by hard-coding the default empty vec to our de-serialization, we are removing that difference. When in doubt, my preference is to always to use `Option<Vec<T>>` because worse case scenario it makes it slightly less ergonomic in exchange for always being correct.

wp_api/src/wp_com/domains.rs
oguzkocer: This is pretty much the same argument as the [above empty vec issue](https://github.com/Automattic/wordpress-rs/pull/1276#discussion_r3074999546). If the API contract specifies that it defaults to `false`, we can.

You might notice that I used 'can' instead of 'should' in both arguments, because WordPress.com API endpoints are known to be altered without a version change. That's why, I prefer safer type signatures and would probably keep these even if I saw that API specified default values. It's just one less thing to worry about and ergonomics of these two are pretty much equivalent to me: `if (hsts_required) {}` & `if (hsts_required = Some(true)) {}`

Having said all that, your argument for this specific case makes sense, and I am happy to make that change if you think we should.

wp_api/src/wp_com/endpoint/domains_endpoint.rs
oguzkocer: No, `false` wouldn't imply `None` - not unless we specifically implement that. I'll add the following case:

```
#[case::include_dotblogsubdomain_false(
    DomainSuggestionsParams { include_dotblogsubdomain: Some(false), ..base_domain_suggestions_params() },
    "/domains/suggestions?query=coolsite&quantity=5&include_dotblogsubdomain=false"
)]
```

wp_api/src/wp_com/domains.rs
oguzkocer: I'll open a follow up PR for this, so we have a reference to the work.

wp_api/src/wp_com/domains.rs
oguzkocer: This one changed in 4dba6545d0cf9deb6c15d9b1647ba37592757f35, because here there is no semantic difference.

jkmassel (approved):

PR #1283, opened 2026-04-15
Add `Decimal2` type for precise 2-decimal-place values

oguzkocer:
Replaces f64 for monetary fields (`raw_price`, `renew_raw_price`, `sale_cost`) in domain suggestions to avoid floating-point precision issues.

### DIFF
--- a/wp_api/src/decimal2.rs
+++ b/wp_api/src/decimal2.rs
@@ -88,7 +88,14 @@ impl de::Visitor<'_> for Decimal2Visitor {
     type Value = Decimal2;
 
     fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        formatter.write_str("a number with at most 2 decimal places")
+        formatter.write_str("a number (or numeric string) with at most 2 decimal places")
+    }
+
+    fn visit_str<E: de::Error>(self, v: &str) -> Result<Self::Value, E> {
+        let parsed: f64 = v
+            .parse()
+            .map_err(|_| E::invalid_value(Unexpected::Str(v), &self))?;
+        self.visit_f64(parsed)
     }
 
     fn visit_u64<E: de::Error>(self, v: u64) -> Result<Self::Value, E> {
@@ -141,6 +148,9 @@ mod tests {
     #[case(r#"{"value": 1.5}"#, 150)]
     #[case(r#"{"value": -3.25}"#, -325)]
     #[case(r#"{"value": 0.01}"#, 1)]
+    #[case(r#"{"value": "19.99"}"#, 1999)]
+    #[case(r#"{"value": "11"}"#, 1100)]
+    #[case(r#"{"value": "0.5"}"#, 50)]
     fn test_deserialize(#[case] json: &str, #[case] expected_hundredths: i64) {
         let wrapper: Wrapper = serde_json::from_str(json).expect("should deserialize");
         assert_eq!(wrapper.value.hundredths(), expected_hundredths);
@@ -150,7 +160,9 @@ mod tests {
     #[case(r#"{"value": 8.635}"#)]
     #[case(r#"{"value": 1.999}"#)]
     #[case(r#"{"value": 0.001}"#)]
-    fn test_deserialize_rejects_more_than_two_decimals(#[case] json: &str) {
+    #[case(r#"{"value": "8.635"}"#)]
+    #[case(r#"{"value": "abc"}"#)]
+    fn test_deserialize_rejects_invalid_values(#[case] json: &str) {
         let result: Result<Wrapper, _> = serde_json::from_str(json);
         assert!(
             result.is_err(),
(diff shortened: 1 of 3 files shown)


wp_api/src/decimal2.rs
```diff
@@ -0,0 +1,182 @@
...
+
+uniffi::custom_type!(Decimal2, i64, {
+    lower: |decimal| decimal.hundredths(),
+    try_lift: |hundredths| Ok(Decimal2::from_hundredths(hundredths)),
+});
+
+/// A decimal number with at most 2 fractional digits, stored as an integer
+/// count of hundredths to avoid floating-point precision issues.
+///
+/// Deserialization rejects values with more than 2 decimal places rather
```
jkmassel, line 15: This is probably the right answer, though I wonder if we'll eventually hit a case where an API fails to round for us. I guess in that case we just fix the API.

oguzkocer: We can add a `Decimal3` or `Decimal4` if that helps? I think `Decimal2` is still useful, so I'd suggest keeping it for cases when we know it's only 2 decimals.

I guess one option was to turn it into a `String` and not let clients do math on it, but then you might have people parsing it, so... I think a solution like this is safer.

wp_api/src/decimal2.rs
```diff
@@ -0,0 +1,182 @@
...
+        let hundredths =
+            i64::try_from(v).map_err(|_| E::invalid_value(Unexpected::Unsigned(v), &self))?;
+        Ok(Decimal2 {
+            hundredths: hundredths * 100,
+        })
+    }
+
+    fn visit_i64<E: de::Error>(self, v: i64) -> Result<Self::Value, E> {
+        Ok(Decimal2 {
+            hundredths: v * 100,
```
jkmassel: Do we want to use `checked_mul` for this and error out if the value is unreasonably large? Seems like overkill, but it'd never be incorrect?

wp_api/src/decimal2.rs
```diff
@@ -0,0 +1,182 @@
...
+
+impl<'de> Deserialize<'de> for Decimal2 {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        deserializer.deserialize_any(Decimal2Visitor)
+    }
+}
+
+struct Decimal2Visitor;
+
+impl de::Visitor<'_> for Decimal2Visitor {
```
jkmassel, line 87: Should we bother implementing `visit_str` for values like `"19.99"`?

wp_api/src/decimal2.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn test_display(#[case] hundredths: i64, #[case] expected: &str) {
+        let d = Decimal2 { hundredths };
+        assert_eq!(d.to_string(), expected);
+    }
+
+    #[rstest]
+    #[case(1100, "1100")]
+    #[case(863, "8.63")]
+    #[case(0, "0")]
+    #[case(1800, "1800")]
```
jkmassel:
```suggestion
    #[case(1100, "11.00")]
    #[case(863, "8.63")]
    #[case(0, "0")]
    #[case(1800, "18.00")]
```

nit – there's a bit of inconsistency here

wp_api/src/decimal2.rs
jkmassel: My thinking was that if it's `Decimal2` we know how how many significant digits the developer would like it rounded to, so we could coerce it and only implement `Decimal4` if we decide we have a type that needs to be that specific.

jkmassel (changes requested):
This generally looks good in its current form, but I had a few questions.

wp_api/src/decimal2.rs
oguzkocer: We started this work for money related calculations and the potential of it being wrong when we used floats. Using _any_ kind of rounding in this type feels quite wrong to me. Am I missing something?

wp_api/src/decimal2.rs
oguzkocer: Added in 3f3b0542315d0edbfcda69ea44d8d39314822a3f.

wp_api/src/decimal2.rs
oguzkocer: Addressed in 7ea2b137f473a0e978c08fc29e56aa9af2e2611b.

wp_api/src/decimal2.rs
oguzkocer: Implemented in d5210e723fb8ae7c19576bfd818d44d4133b1770.

wp_api/src/decimal2.rs
jkmassel: I'm ok to leave it for now. We can revisit if in the future some API response expresses a dollar amount as `1.635` or something.

jkmassel (approved):

PR #1287, opened 2026-04-16
Add `saleCostDisplay` to `PaidDomainSuggestion`

oguzkocer:
Adds formatting that matches the server's `combined_sale_cost_display`: currency prefix from cost field, comma thousand-separators, and two decimal places only for fractional values.

Exposed as a method on `PaidDomainSuggestion` (for Rust callers), a free-standing UniFFI function (for cross-language export), and a Kotlin extension property on `PaidDomainSuggestion`.

Test cases derived from real `/products/?type=domains` responses to ensure the formatting matches the server exactly.

---

@jkmassel WPAndroid currently shows this value from the products endpoint, which we can implement. (may even need to for other screens) However, we also have the necessary components from the suggestions endpoint to construct it ourselves, saving an extra endpoint call. I figured I'd open the PR so you see it in action instead of pinging you and explaining it. So, the short story is that I am not sure if we really want to go about it this way.


oguzkocer:
We decided to keep the current logic and make a request to `/products` endpoint for this information. https://github.com/Automattic/wordpress-rs/pull/1289

PR #1289, opened 2026-04-17
Add `/products` endpoint

oguzkocer:
Implement `GET /products` with optional `type` filter parameter. The `Product` type covers all product variants (domain registrations, transfers, Jetpack plans, etc.) with optional fields for domain-specific data (`tld`, `sale_coupon`, HSTS) and non-domain data (`introductory_offer`, price tiers).


wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[uniffi(default = None)]
+    pub is_dot_gay_notice_required: Option<bool>,
+    /// Formatted monthly cost (e.g. `"$1.50"`).
+    #[serde(default)]
+    #[uniffi(default = None)]
+    pub cost_per_month_display: Option<String>,
+    /// Numeric sale price when a coupon applies.
+    #[serde(default)]
+    #[uniffi(default = None)]
+    pub sale_cost: Option<f64>,
```
jkmassel: #1283 has landed, so we could probably use `Decimal2` here?

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,236 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+pub struct IntroductoryOffer {
+    /// Unit of the offer interval (e.g. `"month"`).
+    pub interval_unit: String,
+    pub interval_count: u32,
+    #[serde(default)]
+    #[uniffi(default = None)]
+    pub usage_limit: Option<u32>,
+    /// Cost per interval during the offer period.
+    pub cost_per_interval: f64,
```
jkmassel: Same point about `Decimal2`

wp_api/src/wp_com/products.rs
oguzkocer: It hadn't landed yet when I worked on this. Let me make that change.

oguzkocer: Addressed in 6b6b53d9cc188b32dca9616a21f0273da31f675a.

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,239 @@
...
+};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+/// Parameters for `GET /products`.
+#[derive(Debug, Default, Clone, PartialEq, Eq, uniffi::Record)]
+pub struct ProductsParams {
+    /// Filter by product type (e.g. `"domains"`).
+    #[uniffi(default = None)]
+    pub product_type: Option<String>,
```
jkmassel: This endpoint has a bunch of localized fields, so we need to provide the locale we want strings in.

The argument is `locale`, and it can be `WPComLanguage` – that should automatically handle serialization into query strings for you.

jkmassel: WDYT about introducing a `ProductType` enum for this?

It would need the `Other` case to handle any new values, but it'd avoid stringly-typing for an operation like "get all the domain registration products" – the API returns all results instead of no results if something is misspelled.

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub product_term_localized: String,
+    pub price_tier_slug: String,
+    #[serde(default)]
+    #[uniffi(default = [])]
+    pub price_tier_list: Vec<PriceTier>,
+    /// Top-level domain for registration products (e.g. `"com"`, `"net"`).
+    /// Absent for non-registration products like domain mapping.
+    #[serde(default)]
+    #[uniffi(default = None)]
+    pub tld: Option<String>,
```
jkmassel: `tld`, `is_privacy_protection_product_purchase_allowed`, `is_hsts_required`, `is_dot_gay_notice_required` are all only defined on domains.

WDYT about defining `DomainProductInfo` to hold these?

jkmassel (changes requested):
Some suggestions for you to work around the strangeness of this API shape.

jkmassel (changes requested):
Left some suggestions

wp_api/src/wp_com/products.rs
oguzkocer: Addressed in 900783af. 

Note that upon reviewing the API implementation, I've also updated `tld_rank` to `Option<f64>` per Claude's suggestion.

wp_api/src/wp_com/products.rs
oguzkocer: Addressed in bb3d213f.

oguzkocer: Addressed in f3b6c71e.

wp_api/src/wp_com/products.rs
oguzkocer: Addressed in 8d88fb4e.

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,369 @@
...
+/// Parameters for `GET /products`.
+#[derive(Debug, Default, Clone, PartialEq, Eq, uniffi::Record)]
+pub struct ProductsParams {
+    /// Filter by product type.
+    #[uniffi(default = None)]
+    pub product_type: Option<ProductTypeFilter>,
+    /// Locale for localized product names and descriptions.
+    #[uniffi(default = None)]
+    pub locale: Option<WPComLanguage>,
+}
```
jkmassel, line 67: There's also a `currency` param that controls which currency is returned.

We should implement it, probably as an enum with `USD`, `EUR`, `AUD`, etc. We can have a catch-all `Other` for it. It can be a separate PR, but we should update the places where currency is emitted to use that too.

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,369 @@
...
+    pub cost_display: String,
+    /// Formatted combined cost without decimal places (e.g. `"$18"`).
+    pub combined_cost_display: String,
+    /// Numeric cost in the account's currency.
+    pub cost: Decimal2,
+    /// Cost in the smallest currency unit (e.g. cents).
+    pub cost_smallest_unit: u64,
+    pub currency_code: CurrencyCode,
+    /// Billing period (e.g. `"year"`).
+    pub product_term: String,
```
jkmassel: This seems like a constrained set of values – should we make it an enum as well?

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,369 @@
...
+    pub transform_quantity_divide_by: Option<u64>,
+    #[serde(default)]
+    #[uniffi(default = None)]
+    pub transform_quantity_round: Option<String>,
+}
+
+/// Details of an active sale coupon applied to a product.
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+pub struct SaleCoupon {
+    pub start_date: String,
```
jkmassel: Based on the field name, I think I'd expect this to be a `Date` type?

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,369 @@
...
+    /// Localized billing period label.
+    pub product_term_localized: String,
+    pub price_tier_slug: String,
+    #[serde(default)]
+    #[uniffi(default = [])]
+    pub price_tier_list: Vec<PriceTier>,
+    /// Domain-specific fields. Fields are populated only for domain
+    /// registration products; all will be `None` for other product types.
+    #[serde(flatten)]
+    pub domain_info: DomainProductInfo,
```
jkmassel: This should probably be optional, because not all products are domains. When it is, all of the fields can be non-optional because they're only populated for domain-shaped objects.

jkmassel (changes requested):
Left some suggestions

wp_api/src/wp_com/products.rs
```diff
@@ -0,0 +1,369 @@
...
+    #[serde(default)]
+    #[uniffi(default = None)]
+    pub tld_rank: Option<f64>,
+}
+
+/// Introductory pricing offer for a product.
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+pub struct IntroductoryOffer {
+    /// Unit of the offer interval (e.g. `"month"`).
+    pub interval_unit: String,
```
jkmassel: This is actually an enum on the server too:

| Value | Meaning |                                                                                                                                                               
  |---|---|                                                                                                                                                                         
  | `"day"` | Daily interval |
  | `"week"` | Weekly interval |                                                                                                                                                    
  | `"month"` | Monthly interval |
  | `"year"` | Annual interval |                                                                                                                                                    
  | `"indefinite"` | No end / indefinite |  


Could have `Other` just in case but it seems nicely shaped.

wp_api/src/wp_com/products.rs
oguzkocer: Addressed in 7b88560fe08519dec69490fd4e377582209c3e84.

wp_api/src/wp_com/products.rs
oguzkocer: Addressed in 18f6b592ffdc43ba84fcb0c8265cbc6de34864d0.

wp_api/src/wp_com/products.rs
oguzkocer: Addressed in 229811b743e893c9870694471f98c906e2e6db53.

wp_api/src/wp_com/products.rs
oguzkocer: Addressed in a5ea0e13d01fc7dfcf704f05c6429e12bb27fa06.

wp_api/src/wp_com/products.rs
oguzkocer: I'll address this in a separate PR to avoid some merge conflicts.

crazytonyli (approved):

jkmassel (approved):
echoing @crazytonyli's approval

PR #1304, opened 2026-04-28
`/template-parts` endpoint

oguzkocer:


wp_api/src/template_parts.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub post_id: Option<PostId>,
+    #[WpContext(edit, embed, view)]
+    pub has_theme_file: Option<bool>,
+    #[WpContext(edit, embed, view)]
+    pub author: Option<UserId>,
+    #[WpContext(edit, view)]
+    #[WpContextualOption]
+    #[serde(
+        default,
+        deserialize_with = "wp_serde_helper::deserialize_false_or_string"
```
crazytonyli, line 70: `WpResponseString` parses false or string, I think? Should we use that?

crazytonyli (approved):

wp_api/src/template_parts.rs
oguzkocer: Let me look into and update this in a subsequent PR because we have a few existing usages of this and it'd be good to replace all of them in one go.

PR #1308, opened 2026-04-28
`/templates/<parent>/revisions` endpoint

oguzkocer:


wp_api/src/template_revisions.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub post_id: Option<PostId>,
+    #[WpContext(edit, embed, view)]
+    pub has_theme_file: Option<bool>,
+    #[WpContext(edit, embed, view)]
+    pub author: Option<UserId>,
+    #[WpContext(edit, view)]
+    #[WpContextualOption]
+    #[serde(
+        default,
+        deserialize_with = "wp_serde_helper::deserialize_false_or_string"
```
crazytonyli, line 116: Use `WpResponseString`?

crazytonyli (approved):

wp_api/src/template_revisions.rs
oguzkocer: Let me look into and update this in a subsequent PR because we have a few existing usages of this and it'd be good to replace all of them in one go.

oguzkocer: I looked into it and shared the situation as reported by Claude below the `---` line. I _think_ this is a safe and reasonable change, but wanted to double check with you and @jkmassel first. What do you think, should I make the change?

---

**Can `WpResponseString` replace `deserialize_false_or_string`?** Functionally almost — both handle `false` → `None` and `true` → error. The difference is `deserialize_false_or_string` also treats the *string* `"false"` as `None`, while `WpResponseString` treats it as `Some("false")`. In practice this shouldn't matter since WordPress returns the boolean `false`, not the string.

**The catch:** Changing field types from `Option<String>` to `WpResponseString` is a public API change. Fields like `modified` would go from `Option<String>` to `WpResponseString` (or `Option<WpResponseString>` with `#[WpContextualOption]`). Callers would need `.0` to get the inner `Option<String>`.

**Affected fields (8 total):**
- `templates.rs:128` — `SparseTemplate.modified`
- `template_parts.rs:72` — `SparseTemplatePart.modified`
- `template_revisions.rs:118` — `SparseTemplateRevision.modified`
- `template_autosaves.rs:54` — `SparseTemplateAutosave.modified`
- `login.rs:73` — `WpApiDetails.site_icon_url`
- `jetpack/connection.rs:96` — `JetpackUser.avatar`
- `wp_com/subscribers.rs:78` — `SubscriberCountry.code`
- `wp_com/subscribers.rs:80` — `SubscriberCountry.name`

PR #1239, opened 2026-03-17
Use SiteInfo to create WpService

crazytonyli:
## Description

There are two separate constructors to create WpService, based on the type of the site. This PR unifies them by using the existing `SiteInfo`.

I have also updated the Swift wrapper `WordPressAPI` to hold on to a `SiteInfo` instance, to simplify the creating WpService API.

PR #1317, opened 2026-05-04
Tolerate URL normalization differences in cache lookups

crazytonyli:
## Description

Two integration tests in `wp_mobile_integration_tests/tests/test_remove_site.rs` panicked with `site should exist`. The root cause was an interaction between two PRs that landed in sequence on trunk:

- **PR #1232** added `test_remove_site.rs`, which looks up the test site with the raw credentials string `http://localhost`, and a `url_candidates` workaround that tolerates trailing-slash differences but only on the *delete* path.
- **PR #1239** then changed `WpService` to accept URLs as `Arc<ParsedUrl>` and store them via `ParsedUrl::url()`, which normalizes `http://localhost` to `http://localhost/`.

Neither PR's CI saw the combination. Post-merge on trunk, the cache stored the normalized form (`http://localhost/`) while the tests still queried the raw form (`http://localhost`), and `select_self_hosted_site_by_url`'s bare `WHERE url = ?` lookup returned `None`. The trailing-slash tolerance from `url_candidates` had never been wired into the read path.

## Changes

Register a `urls_eq(a, b)` SQLite UDF that compares URLs via `ParsedUrl` parsing, replace `WHERE url = ?` with `WHERE urls_eq(url, ?)` in `select_self_hosted_site_by_url`, and remove `url_candidates`. Every URL-keyed read now uses a single tolerant mechanism, so historical caches stored in any normalization shape continue to match.

## Test plan

CI green.

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Support both Integer and String for `WPApiDetails.gmt_offset`](https://github.com/Automattic/wordpress-rs/pull/209)
 - Pin `wp-cli/restful` to v0.4.1 to fix Docker build (v0.4.2+ requires unreleased `wp-cli ^2.13`)
+- **Internal:** `WpApiCache` lookups for a self-hosted site by URL now tolerate trailing-slash and other URL-normalization differences.
 
 ## [0.1]
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,6 +71,7 @@ Test credentials are configured in:
 ## Development Tips
 
 - Platform bindings are generated automatically - don't edit generated files directly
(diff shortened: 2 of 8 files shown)


oguzkocer:
I am reviewing this as-is, but I wonder if it'd be better to handle the existing data through a migration and normalize the insert/look-up instead of trying to work around it.

crazytonyli:
My main concern is that we'll have new code to insert random strings as "site url". To ensure the stored url _exactly_ matches `ParsedUrl`, we'd need to do something like this for each insertion/update:

```
// normalized the `url_str` input to another url string.
let url = ParsedUrl.parse(url_str) // TODO: error handling needed, even though 99.99% time we'll insert legal url strings.
let normalized = url.to_str()

sqlite.exec("INSERT ... VALUE (?)", noramlized)
```

I don't like this approach because:
1. We may need to surface the url parsing error to the function API. That introduce unnecessary noises, when the url input is almost certain a legit url.
2. It'd be easier to remember (or read/write) the new `urls_eq` function than the url parsing dance.

oguzkocer (approved):

PR #1321, opened 2026-05-05
Clean up `pr-build/*` branches for closed PRs on trunk pushes

jkmassel:
## Description

Each PR build creates a `pr-build/<PR#>` branch via `fastlane publish_pr_xcframework` so the iOS apps can pull the snapshot xcframework. Nothing prunes these afterwards, so they accumulate — there are 118 of them on origin today, most for long-merged PRs.

This adds a Buildkite step that runs on every trunk push, lists the `pr-build/*` refs, asks GitHub for each PR's state, and deletes the refs whose PR is `closed` (covers both merged and rejected — GitHub collapses them).

## Changes

- **`.buildkite/cleanup-pr-build-branches.sh`** (new): guards on `BUILDKITE_BRANCH == "trunk"`, sources `use-bot-for-git`, enumerates via `git ls-remote --heads origin 'refs/heads/pr-build/*'`, queries `GET /repos/automattic/wordpress-rs/pulls/<n>` with `GITHUB_TOKEN`, and deletes closed-PR refs with batched `git push origin :refs/heads/...` (chunks of 50). Skips on non-200 responses so we never delete a ref we couldn't verify.
- **`.buildkite/pipeline.yml`**: new top-level step gated to `build.branch == "trunk" && build.env("NEW_VERSION") == null` (skips release builds, matching the publish-trunk-xcframework gate). Default agent queue, like the other git+API helper steps (`validate-changelog.sh`).

## Test plan

- [ ] First trunk merge after this lands sweeps the ~118 existing orphans
- [ ] Subsequent trunk merges delete the just-merged PR's `pr-build/<n>` branch
- [ ] Step is skipped on PR builds and on release builds (`NEW_VERSION` set)
- [ ] Open-PR `pr-build/*` branches are left alone
- [ ] If GitHub returns non-200 for a PR (deleted, rate-limit, transient), the corresponding branch is preserved and the step continues

## Changelog

- [x] No `CHANGELOG.md` entry — CI-only change with no user-visible impact.

### DIFF
--- a/.buildkite/cleanup-pr-build-branches.sh
+++ b/.buildkite/cleanup-pr-build-branches.sh
@@ -12,75 +12,82 @@ if [[ "${BUILDKITE_BRANCH:-}" != "trunk" ]]; then
 fi
 
 if [[ -z "${GITHUB_TOKEN:-}" ]]; then
-  echo "GITHUB_TOKEN not set, cannot query PR state" >&2
+  echo "GITHUB_TOKEN not set, cannot reach GitHub API" >&2
   exit 1
 fi
 
-GITHUB_REPO="automattic/wordpress-rs"
+API="https://api.github.com/repos/automattic/wordpress-rs"
 
-echo '--- :robot_face: Use bot for Git operations'
-source use-bot-for-git
-
-echo "--- :mag: Listing pr-build/* branches on origin"
-mapfile -t branches < <(
-  git ls-remote --heads origin 'refs/heads/pr-build/*' \
-    | awk '{print $2}' \
-    | sed 's|^refs/heads/||'
-)
-
-echo "Found ${#branches[@]} pr-build branches"
-
-if [[ ${#branches[@]} -eq 0 ]]; then
-  exit 0
-fi
-
-echo "--- :github: Checking PR state for each branch"
-to_delete=()
-for branch in "${branches[@]}"; do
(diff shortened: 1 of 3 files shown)


.buildkite/cleanup-pr-build-branches.sh
```diff
@@ -0,0 +1,86 @@
...
+GITHUB_REPO="automattic/wordpress-rs"
+
+echo '--- :robot_face: Use bot for Git operations'
+source use-bot-for-git
+
+echo "--- :mag: Listing pr-build/* branches on origin"
+mapfile -t branches < <(
+  git ls-remote --heads origin 'refs/heads/pr-build/*' \
+    | awk '{print $2}' \
+    | sed 's|^refs/heads/||'
```
crazytonyli: I'm not sure if those branches are all fetched during the CI checkout. Using GH API (like the PR state check) probably would be more robust.

crazytonyli (approved):

.buildkite/cleanup-pr-build-branches.sh
jkmassel: Good catch — switched to all-API in f5569e38. `ls-remote` is technically a remote query so it wouldn't be affected by partial checkouts, but you're right that using the same auth mechanism throughout is cleaner. Dropped `use-bot-for-git` since we don't push anymore.

PR #1335, opened 2026-05-12
Remove 'nav_menu_item' investigation artifacts

oguzkocer:
I think these artifacts got accidentally merged at some point. I am going to admin-merge this after approval because adding a `CHANGELOG` for this should be unnecessary and the change won't impact the build.


jkmassel (approved):
Looks fine, but I don't think it'd hurt to add a CHANGELOG.md entry just outlining that we deleted some unneeded documentation 🤷

oguzkocer:
> Looks fine, but I don't think it'd hurt to add a CHANGELOG.md entry just outlining that we deleted some unneeded documentation 🤷

Hmm.. That's unexpected. I guess I'll add one then.

Issue #1339, opened 2026-05-13
UniFFI checksum mismatch on armeabi-v7a Android builds (115 Sentry users)


nbradbury:
## Summary

`uniffi.wp_api.IntegrityCheckingUniffiLib.<clinit>` throws `RuntimeException: UniFFI API checksum mismatch` on a subset of Android devices in production, almost exclusively 32-bit `armeabi-v7a` split APKs. The embedded Kotlin-side checksums don't match the checksums queried from the packaged native `.so` at load time.

Sentry impact (last 90 days, version `alpha-2026-04-20`):
- [JETPACK-ANDROID-1G48](https://a8c.sentry.io/issues/JETPACK-ANDROID-1G48): 49 users / 538 events
- [WORDPRESS-ANDROID-3EZ1](https://a8c.sentry.io/issues/WORDPRESS-ANDROID-3EZ1): 66 users / 315 events

Combined: **115 users**. Affected devices include Galaxy A02 (`SM-A022F`), Redmi 6, and similar 32-bit ARM hardware. Once `<clinit>` fails the JVM caches the failure, so every subsequent call to anything in `uniffi.wp_api` throws `NoClassDefFoundError` — that's why per-user event counts are high.

**Note:** [This JPAndroid draft PR](https://github.com/wordpress-mobile/WordPress-Android/pull/22860) attempted to resolve the problem by catching the `LinkageError`, but the real source of the problem is in this repo. 

## Reproduction stack trace

```
ExceptionInInitializerError
  at uniffi.wp_mobile.WpService$Companion.wordpressCom-E0BElUM(wp_mobile.kt:8016)
  at uniffi.wp_mobile.UniffiLib.<clinit>(wp_mobile.kt:906)
  at uniffi.wp_api.Wp_apiKt.uniffiEnsureInitialized(wp_api.kt:6571)
Caused by: RuntimeException: UniFFI API checksum mismatch: try cleaning and rebuilding your project
  at uniffi.wp_api.Wp_apiKt.uniffiCheckApiChecksums(wp_api.kt:4388)
  at uniffi.wp_api.IntegrityCheckingUniffiLib.<clinit>(wp_api.kt:867)
```

### 1. UniFFI bindgen reads from the host `.so`, not the Android `.so`

`native/kotlin/build.gradle.kts` (around lines 34-35):

```kotlin
val nativeLibraryPath =
[...]

PR #1348, opened 2026-05-20
Replace small integer types (`u8`, `u16`) with `u32` as defensive measure

oguzkocer:
[An 11-year-old bug in Android's ART AOT compiler for ARM32](https://github.com/jkmassel/uniffi-armv7-aot-checksum-bug) causes incorrect sign-extension of `Short` (16-bit) JNI return values, which can lead to UniFFI checksum mismatches at runtime (#1339).

While the upstream fix is being pursued separately, this change defensively widens all small integer types (`u8`, `u16`) to `u32` across struct fields, function parameters, return types, and enum reprs — particularly those on or near the UniFFI surface. This eliminates the possibility of hitting the AOT sign-extension bug if any of these types are ever exposed through UniFFI bindings.

Byte data (`Vec<u8>`, `&[u8]`, `[u8; N]`) used for request bodies, cryptographic keys, and certificates is left unchanged. Internal `usize` usage tied to `rusqlite` APIs and array indexing is also unchanged, as these types don't cross the FFI boundary.

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Replace `u8`, `u16` types with `u32` across struct fields, function parameters, return types, and enum reprs as a defensive measure against an [Android ART AOT compiler bug](https://github.com/jkmassel/uniffi-armv7-aot-checksum-bug) that mishandles small integer JNI return values on ARM32 ([#1339](https://github.com/Automattic/wordpress-rs/issues/1339))
+- **BREAKING:** Replace `u8`, `u16` types with `u32` across struct fields, function parameters, return types, and enum reprs as a defensive measure against an [Android ART AOT compiler bug](https://github.com/jkmassel/uniffi-armv7-aot-checksum-bug) that mishandles small integer JNI return values on ARM32 ([#1339](https://github.com/Automattic/wordpress-rs/issues/1339))
 
 ## [0.3.0] - 2026-05-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Replace `u8`, `u16` types with `u32` across struct fields, function parameters, return types, and enum reprs as a defensive measure against an [Android ART AOT compiler bug](https://github.com/jkmassel/uniffi-armv7-aot-checksum-bug) that mishandles small integer JNI return values on ARM32 ([#1339](https://github.com/Automattic/wordpress-rs/issues/1339))
```
jkmassel: nit – this is a breaking change and should be marked as such:

```suggestion
- **BREAKING:** Replace `u8`, `u16` types with `u32` across struct fields, function parameters, return types, and enum reprs as a defensive measure against an [Android ART AOT compiler bug](https://github.com/jkmassel/uniffi-armv7-aot-checksum-bug) that mishandles small integer JNI return values on ARM32 ([#1339](https://github.com/Automattic/wordpress-rs/issues/1339))
```

wp_mobile/src/sync/sync_result.rs
```diff
@@ -50,13 +50,13 @@ impl SyncResult {
...
     pub fn no_op(
-        total_items: usize,
+        total_items: u32,
         has_more_pages: Option<bool>,
         current_page: Option<u32>,
         total_pages: Option<u32>,
     ) -> Self {
         Self {
-            total_items: total_items as u64,
+            total_items: total_items.into(),
```
jkmassel: In `new()`, you're using `as u64` – should we use the same approach in both place?

wp_api/src/middleware.rs
jkmassel: We can avoid the cast below if we modify this argument to use u32

jkmassel (approved):
Looks good, just left a few nits

CHANGELOG.md
oguzkocer: Addressed in c40ed339035251c1742ef483cd13e9324f8affd1

wp_mobile/src/sync/sync_result.rs
oguzkocer: Addressed in bbb4c9519e6ad78a966020220728a18ca2c1b7ad

wp_api/src/middleware.rs
oguzkocer: Addressed in 23a8c96e7eb888ae242bcff8d3e121161bc76eb0

PR #1363, opened 2026-05-29
Publish releases automatically on version-bump merges to trunk

crazytonyli:
A new Detect release step inspects each trunk build's CHANGELOG.md change and, when the commit added a new version header that is not yet tagged, uploads a step that builds and publishes the release. This removes the manual Buildkite trigger from the normal flow, while NEW_VERSION is kept as a manual override for recovery or forcing a release.

Please note: manually starting a new build with `NEW_VERSION` is still available. But it should be treated as a fallback now.

### DIFF
--- a/.buildkite/commands/detect-release.sh
+++ b/.buildkite/commands/detect-release.sh
@@ -56,6 +56,16 @@ else
   fi
 fi
 
+# Both trigger paths feed $version into a tag and a release command, so reject
+# anything that is not a semver (e.g. a typo'd NEW_VERSION) before it reaches
+# them. The auto path's extraction already enforces this shape, but a manual
+# NEW_VERSION is taken verbatim, so validate here to cover both. Pattern is
+# X.Y.Z with an optional pre-release/build suffix.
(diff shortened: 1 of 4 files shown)


.buildkite/commands/detect-release.sh
```diff
@@ -0,0 +1,84 @@
...
+# normal "nothing to release" case), so the `|| true` keeps it off pipefail.
+extract_added_version() {
+  grep -E '^\+## \[' \
+    | grep -oE '[0-9]+\.[0-9]+\.[0-9]+[0-9A-Za-z.+-]*' \
+    | head -n 1 \
+    || true
+}
+
+if [ -n "${NEW_VERSION:-}" ]; then
+  version="$NEW_VERSION"
```
oguzkocer, line 30: We should probably run a semver check just in case `NEW_VERSION` is not in our expected format.

oguzkocer (approved):

.buildkite/commands/detect-release.sh
crazytonyli: Added here: https://github.com/Automattic/wordpress-rs/pull/1363/commits/bee1cd8471e464be8304a11d7426f4dc8ae28597

PR #1369, opened 2026-06-01
Fix self-hosted login on plain-permalink sites (#1366)

### LINKED ISSUE #1366: Self-hosted login fails on plain-permalink sites: the ?rest_route= API root is path-extended into non-routing URLs
**Summary**

WordPressLoginClient cannot complete the self-hosted login / site-details fetch when the target site uses plain permalinks (permalink_structure empty). Discovery succeeds and an Application Password is issued, but every subsequent REST request resolves to the API index (or 404s) instead of the requested endpoint, so site details never load. In the WordPress iOS app this surfaces as "Cannot load the WordPress site details" (wordpress-mobile/WordPress-iOS, SelfHostedSiteAuthenticator.swift).

**Root cause**

On a plain-permalink site, WordPress core advertises the REST API root — in the Link; rel="[https://api.w.org/](<https://api.w.org/>)" header and in every \_links href — as the query-parameter form, not a path:

Link: [https://example.com/index.php?rest_route=/](<https://example.com/index.php?rest_route=/>); rel="[https://api.w.org/](<https://api.w.org/>)"

This is core-default behavior (get_rest_url(), the non-pretty branch deliberately builds …/index.php?rest_route=) and the documented universal form.

WordPressLoginClient takes that href as the API root (wp_api/src/login/url_discovery.rs, API_ROOT_LINK_HEADER = "[https://api.w.org/](<https://api.w.org/>)") and then builds endpoint URLs by path-extending it:

* wp_api/src/request/endpoint.rs → WpOrgSiteApiUrlResolver::resolve calls
* wp_api/src/parsed_url.rs → by_extending_and_splitting_by_forward_slash, which is url.extend(segments) — itappends to the path and leaves the ?rest_route=/ query untouched.

So a request for /wp/v2/users/me becomes:

[https://example.com/index.php/wp/v2/users/me?rest_route=/](<https://example.com/index.php/wp/v2/users/me?rest_route=/>)

WordPress ignores the path segments and honors rest_route=/, returning the API index for every endpoint (or, on servers without a front-controller catch-all, a hard 404). The client parses that root document as the site-details response and fails. The hard-coded /wp-json fallback never fires because discovery itself "succeeded." A repo-wide grep for rest_route finds no handling in the self-hosted discovery/login path (only the [WP.com](<http://WP.com>) endpoint).

**Reproduction**

Any self-hosted site with Settings → Permalinks = Plain:

```bash
# 1. Advertised API root is the query form:
curl -sI https://SITE/ | grep -i '^link:'
#   link: <https://SITE/index.php?rest_route=/>; rel="https://api.w.org/"

# 2. The exact URL the client builds collapses to the API index:
curl -s "https://SITE/index.php/wp/v2/users/me?rest_route=/" | head -c 120
#   {"name":"…","description":"…","url":"…", …}   ← API ROOT, not the user resource
```

(wp_rs_cli also can't express this root: --api-root is required to end with /wp-json.)

**Expected**

The client should support the ?rest_route= API-root form. Either:

1. When the discovered root carries a rest_route query param, build endpoints by appending to that query value (…/index.php?rest_route=/wp/v2/users/me) rather than the path; or
2. On discovering the query-form root, probe /wp-json/ and prefer it only if it routes — otherwise use the rest_route query form consistently.

**Environment**

* Surfaced via WordPress iOS 26.8 on iOS 26.5 — companion: wordpress-mobile/WordPress-iOS#25604
* Server Apache/PHP 8.3.30, WordPress 7.0, plain permalinks

**Workaround (server-side — and a clean controlled confirmation)**

Making WordPress advertise the path form fixes it without touching the client, which isolates the client-side path-extension as the defect:

* .htaccess: RewriteRule ^wp-json/(.\*)$ index.php?rest_route=/$1 \[QSA,L\]
* mu-plugin: add_filter('rest_url', fn($u) => str_replace('/index.php?rest_route=/', '/wp-json/', $u));
* (or enable pretty permalinks)

**General Note**<br>I am not a PHP developer or familiar with the low level WordPress code. The above was determined by debugging my system with Claude Code after updating to WordPress 7.0 for my WordPress instance.

jkmassel:
## Description

Fixes [#1366](https://github.com/Automattic/wordpress-rs/issues/1366). Self-hosted login fails on sites with plain permalinks: every REST request resolves to the API index instead of the requested endpoint, so site details never load and the WordPress iOS app surfaces "Cannot load the WordPress site details."

## Root cause

On plain-permalink sites, WordPress advertises the REST API root in the `Link: rel="https://api.w.org/"` header as the query-parameter form:

```
Link: <https://example.com/index.php?rest_route=/>; rel="https://api.w.org/"
```

`WpLoginClient` accepts that URL as the API root, and `WpOrgSiteApiUrlResolver::resolve` then builds endpoint URLs by **path-extending** it:

```
https://example.com/index.php?rest_route=/  +  /wp/v2/users/me
        ↓ (path extension)
https://example.com/index.php/wp/v2/users/me?rest_route=/
```

WordPress ignores the path segments and honors `rest_route=/`, returning the API root JSON for every endpoint. The deserializer tries to parse it as the requested resource and fails.

## Changes

1. **The fix** (`wp_api/src/parsed_url.rs`, `wp_api/src/request/endpoint.rs`): new `ParsedUrl::by_extending_rest_api_path` helper that detects a `rest_route` query parameter on the API root and appends into its value rather than the path. `WpOrgSiteApiUrlResolver::resolve` calls it. No constructors or callers change — the URL itself encodes which form to use.

2. **`wp_rs_cli`** (`wp_rs_cli/...`): the `--api-root` flag never actually enforced a `/wp-json` suffix in code; only the help text and error message claimed it did. Updates the help text and README to acknowledge both forms, and adds a worked example for plain-permalink sites.

3. **Dedicated integration test** (`docker-compose.plain-permalinks.yml`, `scripts/setup-plain-permalinks-test-site.sh`, `wp_api_integration_tests/tests/test_login_plain_permalinks.rs`, Makefile targets): a separate WordPress Docker container with plain permalinks, on port 8081, running an end-to-end login + `users/me` test. Test self-skips when the dedicated server isn't up, so the default suite is unaffected.

4. **CI** (`.buildkite/pipeline.yml`): new step in the e2e group that brings up the dedicated server and runs the integration test.

## Test plan

- [x] `cargo test --lib` — 1736 tests pass
- [x] Unit tests for `by_extending_rest_api_path` cover pretty permalinks, `?rest_route=/`, `?rest_route=` (no trailing slash), and `rest_route` mixed with other query params
- [x] Resolver-level test confirms both URL shapes produce the right endpoint URL
- [x] Integration test passes against the dedicated plain-permalinks WordPress instance
- [x] Integration test **fails** on `trunk` without the fix (response parse error: `missing field 'id' at line 1 column 186231` — the deserializer hitting the API root JSON instead of `users/me`)
- [x] `wp_rs_cli fetch-post --api-root 'http://localhost:8081/index.php?rest_route=/' --post-id 1 --username … --password … --pretty` returns the expected post JSON
- [x] `cargo fmt --all -- --check`, clippy clean on touched files (one pre-existing `iter-kv-map` lint in `url_discovery.rs:121` is unrelated and present on `trunk`)
- [ ] CI Buildkite e2e job for plain permalinks passes

## Related issues

- Companion: [wordpress-mobile/WordPress-iOS#25604](https://github.com/wordpress-mobile/WordPress-iOS/issues/25604)
- Reported by @dekmeister

## Changelog


scripts/setup-plain-permalinks-test-site.sh
```diff
@@ -0,0 +1,63 @@
...
+wp core install \
+	--url=http://localhost \
+	--title=plain-permalinks-test-site \
+	--admin_user="$ADMIN_USERNAME" \
+	--admin_email="$ADMIN_USERNAME" \
+	--admin_password="$ADMIN_ACCOUNT_PASSWORD" \
+	--skip-email
+
+# Default permalink_structure is empty, but be explicit so the intent is clear.
+wp rewrite structure ''
```
crazytonyli: Is this command the only difference between this new local site and the existing local site? If that is the case, is it possible to simplify the test setup by making the new test a "mut" test where it calls the same wp command to modify the site?

jkmassel: It's possible, but I'd worry about downstream breakage.

WDYT @oguzkocer – I know you built this to start with?

oguzkocer: I think it'd work. The whole setup is built in a way that restores the site to a known state. So, if I am not mistaken, all you need is to add a new endpoint to `wp_api_integration_tests_backend/src/main.rs` to run the command `wp rewrite structure ''` and then you can follow the existing mut tests setup with `#[serial]` annotation and call `RestoreServer::db().await;` at the end.

wp_api/src/parsed_url.rs
```diff
@@ -37,6 +37,60 @@ impl ParsedUrl {
...
+                s.as_ref()
+                    .split('/')
+                    .filter_map(|x| match x.trim() {
+                        "" => None,
+                        y => Some(y.to_string()),
+                    })
+                    .collect::<Vec<String>>()
+            })
+            .collect::<Vec<_>>()
+            .join("/");
```
oguzkocer, line 63: I believe the no-`rest_route` case can delegate directly to `by_extending_and_splitting_by_forward_slash`. If we handle that path first and early-return, we can drop the pre-join here (lines 51-63) and the second split → filter_map pass it triggers at line 92 — only the `rest_route` branch actually needs the joined string.

oguzkocer (commented):
The fix looks good. I haven't reviewed the setup changes because as mentioned by @crazytonyli we can probably make this a regular mut test instead. I've also left one suggestion.

crazytonyli (approved):

PR #1376, opened 2026-06-04
Support attachments when replying to unified conversations

adalpari:
## Description

Enables attachment uploads when replying to a unified support conversation. Previously, the unified conversations reply endpoint only accepted a plain-text `message`, while the older support-tickets reply endpoint already supported attachments. This brings the unified endpoint to parity using the exact same multipart machinery.

## Changes

- Added an `attachments: Vec<String>` field to `ReplyToUnifiedConversationParams` (defaults to empty, skipped during JSON serialization).
- Implemented `RequiresMultipartForm` for `ReplyToUnifiedConversationParams`, mapping each attachment path to an `attachment_N` file part — mirroring `AddMessageToSupportConversationParams`.
- Marked the `ReplyToUnifiedConversation` endpoint as `multipart = true` so the request builder sends `multipart/form-data` instead of JSON.

Endpoint (unchanged URL): `POST /wpcom/v2/mobile-support/unified-conversations/<conversation_id>`

## Test plan

- `cargo build -p wp_api` and `cargo clippy -p wp_api --all-features` pass with no warnings.
- `cargo test -p wp_api --lib unified_conversations` passes (existing endpoint URL + deserialization tests).
- ⚠️ Server-side caveat: this only makes the client *send* multipart attachments. Confirm the WordPress.com `unified-conversations/<id>` reply route actually accepts `attachment_N` file parts; otherwise uploaded files will be ignored by the backend.

## Changelog

- [ ] I've added an entry to `CHANGELOG.md` under `## [Unreleased]`, using the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) categories (`Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security`). Prefix breaking changes with `**BREAKING:**`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Live membership updates on `MediaMetadataCollectionWithEditContext`, so cached media lists reflect items moving in or out of the filter without a full refresh
 - WordPress.com `/domains/{name}/is-available` endpoint for checking domain availability, pricing, and transfer status
 - WordPress.com `/all-domains` endpoint for listing all domains across a user's sites
-- WordPress.com `/mobile-support/unified-conversations` endpoints for listing, fetching, and replying to unified support conversations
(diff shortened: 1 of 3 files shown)


adalpari:
This seems to be working good. So, Im merging it into the feature branch and we can review it there

PR #1378, opened 2026-06-05
Fix WpComApiClient crash when refreshing authentication

crazytonyli:
Alternative to https://github.com/Automattic/wordpress-rs/pull/1127

This PR solves the same problem, with a different approach: instead of checking if the `/wp/v2` namespace can be "resolved", this PR checks if the authentication method is using an application password. Since the `fetch_authentication_state`'s implementation is verifying the application password is still valid, it makes more sense to early exit the function unless an application password is used.

### DIFF
--- a/wp_api/src/request.rs
+++ b/wp_api/src/request.rs
@@ -1,7 +1,7 @@
 use self::endpoint::WpEndpointUrl;
 use crate::{
     api_error::{ParsedRequestError, RequestExecutionError, WpApiError, WpErrorCode},
-    auth::WpAuthenticationProvider,
+    auth::{WpAuthentication, WpAuthenticationProvider},
     url_query::{FromUrlQueryPairs, UrlQueryPairsMap},
 };
 use base64::Engine;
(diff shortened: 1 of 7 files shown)


wp_api/src/request.rs
```diff
@@ -944,7 +944,21 @@ pub async fn fetch_authentication_state(
...
+    // authenticates with one (HTTP Basic). For any other scheme (e.g. a bearer
+    // token on WordPress.com) there is no application password to introspect, so
+    // there is no authentication state to report (`Ok(None)`). This also avoids
+    // resolving the `/wp/v2` namespace on resolvers that don't serve it (e.g.
+    // `WpComApiClientInternalUrlResolver`), which would panic.
+    if !matches!(
+        authentication_provider.current_auth(),
+        WpAuthentication::AuthorizationHeader { .. }
+    ) {
+        return Ok(None);
```
jkmassel, line 959: WDYT about having AuthenticationState::Invalid as a signal that the app needs to rethink what it's trying to do?

That's kinda what `None` is doing here, but this is more explicit, I think?

oguzkocer: I'd be fine with using a different `enum` variant, but I suggest re-considering `AuthenticationState::Invalid` name simply because it's easy to mistake it as authentication state is invalid. Wait. I just I wrote that in sentence form 😅 Basically it can mean the same thing as `AuthenticationState::Unauthorized` without full context.

Having said that, I think the part of the reason we are struggling to find the right solution is maybe because we are trying to work within a structure that's slightly shaky. Specifically, the `fetch_authentication_state` name is probably the wrong name, something I'm only realizing in hindsight, and something like `fetch_application_password_authentication_state` would be a lot more explicit and returning `None` would then make more sense.

Furthermore, we probably shouldn't be calling `fetch_application_password_authentication_state` in the first place if the authentication type is not relevant. It might be fine to keep the proposed logic from this PR just for practicality, but we could also consider if we can move the logic elsewhere, by maybe adding a `should_check_application_password_authentication_state` function and utilizing that at the call site. We'd of course still need to check that same `should_check_application_password_authentication_state` function within `fetch_application_password_authentication_state`, but this'd make everything more explicit.

oguzkocer (approved):
This looks good to me as is. I've added a few suggestions for consideration, but none of them are blockers in my opinion.

wp_api/src/request.rs
crazytonyli: I treat the `Ok(None)` return value as "I don't know".

The current state of `fetch_authentication_state` is incomplete. We only validate one authentication method: application passwords. The complete solution would look like this:
```rust
match current_auth() {
  AuthenticationHeader => validate_application_password()
  Bearer => validate_token()
  ...
}
```

We can safely assume that `wp/v2/application-password/introspect` is the way to implement `validate_application_password`, regardless of where the site is hosted, which is what the current code does.

But I don't think we should hard-code using WP.com endpoints in `validate_token`, if we want to implement it _correctly_, because not all bearer tokens belong to WP.com sites. To validate the token, we need someone who knows where the site is hosted. In the current architecture, that is the `ApiUrlResolver`, which is why I suggested moving the `fetch_authenticate_state` to `ApiUrlResolver` (or `HostingProvider`): https://github.com/Automattic/wordpress-rs/pull/1127#issuecomment-3820329400.

I will merge the PR as it is for now. We can continue the discussion to find the most appropriate solution.

PR #1379, opened 2026-06-05
Support encrypted log IDs when replying to unified conversations

adalpari:
## Description

Mirrors `CreateSupportTicketParams.encrypted_log_ids` on the unified-conversations reply endpoint. Apps that have already uploaded logs to the encrypted logging service can pass the resulting IDs when replying to a unified conversation, so the server can correlate the logs with the conversation — the same flow already supported when creating a support ticket.

The IDs are opaque pass-through strings serialized into the reply body; no upload logic lives in this library.

## Changes

- Add an `encrypted_log_ids: Vec<String>` field to `ReplyToUnifiedConversationParams`, matching the `CreateSupportTicketParams` pattern (serialized as a regular field, `#[uniffi(default = [])]` so existing Swift/Kotlin callers stay source-compatible)
- Update the unified-conversations CHANGELOG entry to mention encrypted log ID support when replying

## Test plan

- `cargo test --lib -p wp_api` — 1733 passed
- `cargo fmt --all -- --check` — clean
- Existing unified-conversations deserialization and endpoint URL tests still pass

## Changelog


adalpari:
Merging it into the feature branch

PR #1392, opened 2026-06-10
Add wp.com `POST /sites/<site_id>/domains/primary` endpoint

oguzkocer:
## Description

Adds the `POST /rest/v1.1/sites/<site_id>/domains/primary/` endpoint for setting a site's primary domain. The request body carries a single `domain` field, and the response is `{ "success": bool }`. Used after a successful domain registration to designate the newly registered domain as the site's primary.

## Changes

- New `SetPrimaryDomainParams` (request body, single `domain: DomainName` field) and `SetPrimaryDomainResponse` (`success: bool`) types in `domains.rs`
- New `SetPrimary` `POST` variant on the `DomainsRequest` enum, served from the `RestV1.1` namespace
- Idempotent `domains::set_primary` e2e test: reads the site's current primary domain via `site_domains` and sets it again, so the endpoint is exercised against the live test site without changing its effective primary domain

## Changelog

### DIFF
--- a/wp_api/src/wp_com/domains.rs
+++ b/wp_api/src/wp_com/domains.rs
@@ -1000,6 +1000,20 @@ pub struct SiteDomain {
     pub whois_update_unmodifiable_fields: Option<Vec<String>>,
 }
 
+/// Request body for `POST /rest/v1.1/sites/{siteId}/domains/primary/`.
+#[derive(Debug, Clone, Serialize, uniffi::Record)]
+pub struct SetPrimaryDomainParams {
+    /// The domain name to set as the site's primary domain.
+    pub domain: DomainName,
+}
+
+/// Response from `POST /rest/v1.1/sites/{siteId}/domains/primary/`.
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+pub struct SetPrimaryDomainResponse {
+    /// Whether the primary domain was set successfully.
+    pub success: bool,
+}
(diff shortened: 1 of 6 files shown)


wp_api/src/wp_com/domains.rs
```diff
@@ -980,6 +980,20 @@ pub struct SiteDomain {
...
+pub struct SetPrimaryDomainParams {
+    /// The domain name to set as the site's primary domain.
+    pub domain: DomainName,
+}
+
+/// Response from `POST /rest/v1.1/sites/{siteId}/domains/primary/`.
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+pub struct SetPrimaryDomainResponse {
+    /// Whether the primary domain was set successfully.
+    pub success: bool,
```
crazytonyli, line 1014: What if success is false? I remember some WP.com API would return error message and code in the response. Not sure about this one.

crazytonyli (approved):

wp_api/src/wp_com/domains.rs
oguzkocer: At the time of this comment, the server never returns `false` for this field as it's hard-coded. I'll be reporting a bug relevant to this, and the fix _might_ change that - although I don't expect that to be the case. Now that the PR is approved, I'll land it as is and if needed, follow-up on it with a subsequent PR.

PR #1398, opened 2026-06-13
Add injectable request error logging to Kotlin clients

oguzkocer:
## Description

Diagnosing a failed request previously meant each caller hand-rolling a description of the `WpRequestResult` error variant, and it was easy to drop useful fields (status code, error code, response body).

This adds a small, opt-in mechanism for routing failed requests to the caller's own logger or crash reporter, instead of any built-in sink.

## Changes

- Add `WpRequestResult.toLogErrorString()`: a concise, log-only description of a failed request (or `null` on success) covering every error variant, including the request method and URL. Documented as log/crash-reporting only — never for user display, as it may contain raw response bodies and URLs.
- Add a `RequestErrorLogger` functional interface — a single `logError(message: String)` sink that callers implement to forward messages to their platform logger or crash-reporting breadcrumbs.
- Add an optional `errorLogger: RequestErrorLogger? = null` constructor parameter to `WpComApiClient`, `WpApiClient`, and `JetpackApiClient`. When set, failed requests are reported to it via `toLogErrorString()`; when left unset, no logging occurs.
- Suppress `LongParameterList` on the interceptor convenience constructors: the new parameter pushes them to 7, but they are all optional config with defaults, so `errorLogger` is kept on every constructor for a consistent API rather than dropped asymmetrically.

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - WordPress.com REST API implementation checklist (`WPCOM_REST_API_CHECKLIST.md`)
 - WordPress.com `GET /me/transactions/supported-countries` endpoint for listing countries supported in payment transactions
 - WordPress.com `GET /me/domain-contact-information` endpoint for fetching WHOIS/domain contact details
+- Kotlin: `WpRequestResult.toLogErrorString()` returns a concise, log-only description of a failed request (or `null` on success) for diagnostics
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


crazytonyli (approved):
I wonder if the Rust library should log HTTP errors, like using the log crate, and the iOS and Android apps can opt-in to record the logs in the platform implementations.

oguzkocer:
> I wonder if the Rust library should log HTTP errors, like using the log crate, and the iOS and Android apps can opt-in to record the logs in the platform implementations.

I have been wondering the same. I think it'd be worth spending some time on this at some point and solve it once and for all. Having said that, this PR is a temporary practical step that gets us most of what we need in Kotlin side, so I don't think we have to prioritize it.

PR #1413, opened 2026-06-29
Default media upload progress total to the file size

crazytonyli:
When the progress passed to uploadMedia has a total of zero, default it to the byte size of the file being uploaded.

### DIFF
--- a/native/swift/Sources/wordpress-api/WordPressAPI.swift
+++ b/native/swift/Sources/wordpress-api/WordPressAPI.swift
@@ -280,9 +280,11 @@ extension Progress {
     func setTotalUnitCountToFileSizeIfUnset(filePath: String) throws {
         guard totalUnitCount == 0 else { return }
         let attributes = try FileManager.default.attributesOfItem(atPath: filePath)
-        if let size = attributes[.size] as? NSNumber {
-            totalUnitCount = size.int64Value
-        }
+        let size = (attributes[.size] as? NSNumber)?.int64Value ?? 0
+        // A zero-byte file (or an unreadable size) would leave the total at 0 and trip
+        // `fulfill`'s precondition. Use 1 as a placeholder so the request proceeds and the
+        // caller gets the server's "file is empty" error instead of a crash.
+        totalUnitCount = max(size, 1)
     }
 }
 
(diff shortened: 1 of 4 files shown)


native/swift/Sources/wordpress-api/WordPressAPI.swift
```diff
@@ -269,6 +270,22 @@ public final class WordPressAPI: Sendable {
...
+    /// Defaults an unset total (0) to the byte size of the file at `filePath`.
+    ///
+    /// Media upload callers otherwise have to stat the file themselves just to satisfy
+    /// `fulfill`'s requirement that the total be greater than zero. A missing or unreadable
+    /// file throws rather than leaving the total unset, so the failure surfaces here instead
+    /// of as a confusing zero-total upload.
+    func setTotalUnitCountToFileSizeIfUnset(filePath: String) throws {
+        guard totalUnitCount == 0 else { return }
+        let attributes = try FileManager.default.attributesOfItem(atPath: filePath)
+        if let size = attributes[.size] as? NSNumber {
```
jkmassel: What happens if the size is zero? (for instance – `touch foo` will produce a zero-byte file – you can validate this with `touch foo && du -sh foo`)

I'd suggest emitting an `EmptyFileError` here – there's really nothing for the server to do, and it's a unique enough error that it probably points to some invariant being violated (maybe some iCloud resource hasn't materialized properly) that we can't do much about besides notifying the user?

jkmassel (changes requested):
One small but important (IMHO) suggestion, then I think this is ok.

I don't think this is actually tested in macOS (because integration tests only run on Linux). Ideally it would, but I don't think that's in scope for this PR.

native/swift/Sources/wordpress-api/WordPressAPI.swift
crazytonyli: I have hard-coded a minimal 1 total value. I don't think it's worth adding an error here, though. The server side will return an error `rest_upload_unknown_error` when uploading empty files.

crazytonyli:
> I don't think this is actually tested in macOS (because integration tests only run on Linux).

We can't run macOS integration tests on CI, because it requires docker. You can run them locally though.

jkmassel (approved):

PR #1416, opened 2026-06-30
Generate and publish AI reference docs for `rs.wordpress.api:kotlin`

oguzkocer:
### Description

This adds a compact, per-endpoint Markdown API reference (the "AI docs") that is generated from the UniFFI-generated Kotlin bindings and published alongside `rs.wordpress.api:kotlin` as an `ai-docs` Maven classifier zip.

The goal is to let agents and tooling fetch and read one small Markdown file per endpoint instead of crawling the ~191K-line generated `wp_api.kt` (or the Rust source). A consumer resolves the `ai-docs` artifact for the `rs.wordpress.api` version it already depends on and unpacks one file per endpoint domain.

All of the logic lives in `native/kotlin/buildSrc` and is self-contained: the whole feature is wired into a module with a single `id("ai-docs")` line.

## Changes

- **Doc generator** (`native/kotlin/buildSrc/src/main/kotlin/aidocs/`): a `GenerateAiDocsTask` that parses the generated `wp_api.kt` into a structured model and renders Markdown — executor interfaces become method-signature lists, `data class`es become field tables, and `sealed`/`enum` classes become variant lists.
- Structured for testability: `BindingsParser` (`parse()` → `ParsedBindings`) and `DocsGenerator` (`generate()` → `List<GeneratedDoc>`) are pure, with all filesystem I/O isolated in `GenerateAiDocsTask`.
- **Publishing**: a precompiled script plugin `ai-docs.gradle.kts` (applied via `id("ai-docs")` in `api/kotlin/build.gradle.kts`) generates the docs, zips them reproducibly with Gradle's built-in `Zip` (stable order, fixed timestamps), and attaches the zip to the `rs.wordpress.api:kotlin` Maven publication under the `ai-docs` classifier.
- **Unit tests**: `BindingsParserTest` covers the parser's non-trivial cases (method-signature parsing, `data class` fields and defaults, `sealed`/`enum` variants, and generic-aware parameter splitting), using `kotlin("test")` on the JUnit Platform.
- **CI**: a dedicated `:kotlin: buildSrc Unit Tests` Buildkite step runs `./gradlew -p buildSrc test`, since `buildSrc`'s test task is not part of the main build's task graph.

## Changelog


native/kotlin/buildSrc/src/main/kotlin/aidocs/BindingsParser.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    // Split a parameter list on its top-level commas, keeping commas inside generic `<...>` brackets
+    // attached to their parameter (e.g. `filter: Map<String, Int>` stays a single parameter).
+    private fun splitParams(params: String): List<String> {
+        val parts = mutableListOf(StringBuilder())
+        var depth = 0
+        for (char in params) {
+            when (char) {
+                '<' -> depth++
+                '>' -> depth--
```
mokagio: According to my AI, this could end up with negative depth, resulting in commas stopping to split.

```suggestion
                '>' -> depth = maxOf(0, depth - 1)
```

mokagio (approved):
I asked Opus 4.8 to review this with an eye on test coverage. Relevant notes:

- detekt doesn't cover buildSrc — the kotlin-detekt step runs ./gradlew detektMain detektTest on the main build, and buildSrc has its own settings.gradle.kts, so ~600 new lines of Kotlin ship unlinted.
- `DocsGenerator` has no test coverage at all — the type closure, markdown rendering, index building, and `EXCLUDED_METHODS` filtering are untested despite the parser/generator split being framed as "structured for testability".
- Every parser fixture is hand-written to mirror UniFFI's formatting, so a codegen change would silently produce wrong docs with the suite still green — could we assert against a checked-in slice of the real generated `wp_api.kt` instead?
- If the parser matches nothing the task still succeeds and publishes a well-formed empty archive — failing when `parsed.executors` is empty would turn that into a build error.

oguzkocer:
Thanks for the review @mokagio.

> DocsGenerator has no test coverage at all — the type closure, markdown rendering, index building, and EXCLUDED_METHODS filtering are untested despite the parser/generator split being framed as "structured for testability".

This is intended as the generated documents are WIP. Even when we are happy with the result, it's extremely low value to test the generated text.

> Every parser fixture is hand-written to mirror UniFFI's formatting, so a codegen change would silently produce wrong docs with the suite still green — could we assert against a checked-in slice of the real generated wp_api.kt instead?

There would be no point in an abstract verification. If the uniffi's formatting has changed, we can't produce a usable documentation from it. So, what would be the point of verifying that the parser works on an input we no longer care about?

---

We can add detekt the next time we work on this, but to be clear, none of that code is being _shipped._ `buildSrc` by design is not something that's shipped to others.

Same thing with parser not matching anything. It'll never happen in practice, so whenever we work on it, we can add the extra check.

---

> I asked Opus 4.8 to review this with an eye on test coverage

In my opinion, that was the wrong thing to focus on for this PR, but I appreciate you taking a look. 🙇‍♂️

PR #1430, opened 2026-07-08
[Tooling] Add remote login test-suite health digest

oguzkocer:
## Description

Adds a scheduled Buildkite pipeline plus a Fastlane lane (`report_remote_login_integration_test_suite_health`) that post a weekly Slack digest of the quarantined `test_login_remote` step: pass rate and timeout-vs-other error counts, read from the Buildkite API with failing runs linked for investigation.

## Changelog

### DIFF
--- a/.buildkite/report-remote-login-integration-test-suite-health.yml
+++ b/.buildkite/report-remote-login-integration-test-suite-health.yml
@@ -1,11 +1,8 @@
-# Scheduled health digest for the quarantined remote login-discovery integration tests.
-#
-# The schedule that runs this pipeline is managed externally (Terraform), not from
-# this repo.
-#
-# The report itself is the `report_remote_login_integration_test_suite_health`
-# Fastlane lane (see `fastlane/Fastfile`). It is pure Ruby (Buildkite API + Slack
-# webhook), so it runs on the Linux `android` queue — no Xcode/mac VM needed.
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+# Runs the health digest for the quarantined remote login-discovery integration tests.
+# Meant to run on a schedule configured on the Buildkite end.
 steps:
   - label: ":rust: :mag: Remote Login Integration Test Suite Health"
     command: |
--- a/.buildkite/report-remote-login-integration-test-suite-health.yml
+++ b/.buildkite/report-remote-login-integration-test-suite-health.yml
@@ -1,0 +1,16 @@
+# Scheduled health digest for the quarantined remote login-discovery integration tests.
+#
+# The schedule that runs this pipeline is managed externally (Terraform), not from
+# this repo.
+#
+# The report itself is the `report_remote_login_integration_test_suite_health`
+# Fastlane lane (see `fastlane/Fastfile`). It is pure Ruby (Buildkite API + Slack
+# webhook), so it runs on the Linux `android` queue — no Xcode/mac VM needed.
+steps:
+  - label: ":rust: :mag: Remote Login Integration Test Suite Health"
+    command: |
+      install_gems
+      bundle exec fastlane report_remote_login_integration_test_suite_health
+    plugins: [$CI_TOOLKIT]
+    agents:
+      queue: android
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -516,21 +516,20 @@ lane :report_remote_login_integration_test_suite_health do
   require 'net/http'
   require 'json'
 
-  api_token = ENV['BUILDKITE_TOKEN'] || ENV['BUILDKITE_API_TOKEN'] ||
-              UI.user_error!('Set BUILDKITE_TOKEN (with the read_builds scope) to run this lane.')
+  api_token = get_required_env!('BUILDKITE_TOKEN')
   slack_webhook = get_required_env!('SLACK_WEBHOOK')
   days = Integer(ENV.fetch('REPORT_DAYS', '7'))
 
   org = 'automattic'
-  pipeline = 'wordpress-rs'
+  pipeline = PROJECT_NAME
   slack_channel = '#wordpress-rs'
   step_label = 'Remote Login Discovery' # substring of the soft-fail step's label
   investigate_cap = 3
   timeout_signature = /HttpTimeoutError|timed out/
   created_from = (Time.now.utc - (days * 86_400)).strftime('%Y-%m-%dT%H:%M:%SZ')
 
-  # The step is soft-fail, so its failures never reach GitHub. Read each run's real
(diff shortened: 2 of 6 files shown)


crazytonyli (approved):
Have you looked into if buildkite can generate such report for us? https://buildkite.com/platform/test-engine/

.buildkite/report-remote-login-integration-test-suite-health.yml
```diff
@@ -0,0 +1,16 @@
+# Scheduled health digest for the quarantined remote login-discovery integration tests.
+#
+# The schedule that runs this pipeline is managed externally (Terraform), not from
+# this repo.
+#
+# The report itself is the `report_remote_login_integration_test_suite_health`
+# Fastlane lane (see `fastlane/Fastfile`). It is pure Ruby (Buildkite API + Slack
+# webhook), so it runs on the Linux `android` queue — no Xcode/mac VM needed.
```
mokagio: We usually add the scheme to the pipeline.

```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
---

# Scheduled health digest for the quarantined remote login-discovery integration tests.
#
# The schedule that runs this pipeline is managed externally (Terraform), not from
# this repo.
#
# The report itself is the `report_remote_login_integration_test_suite_health`
# Fastlane lane (see `fastlane/Fastfile`). It is pure Ruby (Buildkite API + Slack
# webhook), so it runs on the Linux `android` queue — no Xcode/mac VM needed.
```

Additionally, but this might just be me, I find these AI-generated comments to be too verbose. I don't know if it's useful to keep them for other LLMs to read when they encounter the code next, but I think most of the context they provide is redundant and can be derived by looking at the code itself.


```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
---

# Runs the health digest for the quarantined remote login-discovery integration tests.
# Meant to run on a schedule configured on the Buildkite end.
```

fastlane/Fastfile
```diff
@@ -510,6 +510,91 @@ lane :set_up_signing_release do |readonly: true|
...
 end
 
+desc 'Post a Slack health digest for the quarantined remote login-discovery integration tests'
+lane :report_remote_login_integration_test_suite_health do
+  require 'buildkit'
+  require 'net/http'
+  require 'json'
+
+  api_token = ENV['BUILDKITE_TOKEN'] || ENV['BUILDKITE_API_TOKEN'] ||
+              UI.user_error!('Set BUILDKITE_TOKEN (with the read_builds scope) to run this lane.')
```
mokagio: What do you think of supporting only one name for the env var? It's strict, sure, but makes the code simple as it would allow using 

```suggestion
  api_token = get_required_env!('BUILDKITE_TOKEN')
```

That's assuming the failure message with the read_builds scope memo is not critical.

fastlane/Fastfile
```diff
@@ -510,6 +510,91 @@ lane :set_up_signing_release do |readonly: true|
...
+  slack_webhook = get_required_env!('SLACK_WEBHOOK')
+  days = Integer(ENV.fetch('REPORT_DAYS', '7'))
+
+  org = 'automattic'
+  pipeline = 'wordpress-rs'
```
mokagio: Nitpick. We have `GITHUB_REPO` and `PROJECT_NAME` defined up top. They could be tidied up and reused her, just to DRY a bit.

https://github.com/Automattic/wordpress-rs/blob/d1b2965d610c5984fff48273d71d6d9fec201261/fastlane/Fastfile#L14-L22

But just a nit/idea, no stress.

fastlane/Fastfile
```diff
@@ -510,6 +510,91 @@ lane :set_up_signing_release do |readonly: true|
...
+  slack_channel = '#wordpress-rs'
+  step_label = 'Remote Login Discovery' # substring of the soft-fail step's label
+  investigate_cap = 3
+  timeout_signature = /HttpTimeoutError|timed out/
+  created_from = (Time.now.utc - (days * 86_400)).strftime('%Y-%m-%dT%H:%M:%SZ')
+
+  # The step is soft-fail, so its failures never reach GitHub. Read each run's real
+  # exit_status from the Buildkite API instead. Builds come back newest-first.
```
mokagio: The soft-fail assumption may not hold in the future, but the "read from API" approach remains solid regardless, IMHO.

```suggestion
  # Read each run's real exit_status from the Buildkite API for accuracy.
  # Builds come back newest-first.
```

fastlane/Fastfile
```diff
@@ -510,6 +510,91 @@ lane :set_up_signing_release do |readonly: true|
...
+    else
+      failed = content.scan(/(login_spec_[a-z0-9_]+) \.\.\. FAILED/).flatten.uniq
+      other << { build: run[:build], url: run[:url], test: failed.empty? ? 'unknown' : failed.join(', ') }
+    end
+  end
+
+  total = runs.size
+  passed = runs.count { |run| run[:exit_status].zero? }
+
+  # --- Compose the Slack digest ---
```
mokagio: Nit. If we keep the comment, then let's format it like the ones above.

```suggestion
  # Compose the Slack digest.
```

Otherwise, the code already speaks for itself

```suggestion
```

mokagio (approved):
Only left a few nitpicks.

oguzkocer:
@mokagio All the review feedback have been addressed in 065b3358127f6593679c3bf6c0a8cf5d557b5b06.

@crazytonyli I hadn't looked at it Buildkite Test Engine, but I did after you brought it up. Cargo doesn't produce the necessary output for it, and `cargo-nextest` that does wouldn't work for our needs - specifically our mutable test setup with `#[serial]`. We could build something for it, but it's outside the scope of this PR, which is trying to address a much smaller, albeit important, issue.

.buildkite/report-remote-login-integration-test-suite-health.yml
oguzkocer: Addressed in 065b3358

fastlane/Fastfile
oguzkocer: Addressed in 065b3358

fastlane/Fastfile
oguzkocer: Addressed in 065b3358 — reused `PROJECT_NAME` for the pipeline slug; kept `org` as a literal since the Buildkite org is a separate concept from `GITHUB_REPO` (they only coincidentally share the `automattic` slug).

fastlane/Fastfile
oguzkocer: Addressed in 065b3358

fastlane/Fastfile
oguzkocer: Addressed in 065b3358

PR #1432, opened 2026-07-13
Replace `rust-android-gradle` with `cargo-ndk`

oguzkocer:
## Description

Builds the Android JNI libraries with the `cargo-ndk` cargo subcommand instead of the `org.mozilla.rust-android-gradle` Gradle plugin. The plugin has no Android Gradle Plugin 9 support, which blocks upgrading AGP.

## Changes

- Remove the `rust-android-gradle` plugin from the version catalog and all modules (it was unused in `:api:kotlin`).
- Reimplement `:api:android:cargoBuild` as a Gradle `Exec` task that runs `cargo ndk`, keeping the task name, `-PcargoTarget` ABI names, and `build/rustJniLibs/android/<abi>/` output path.
- Prune all but `libwp_mobile.so` from the output, since `cargo-ndk` copies every cdylib the build produces.
- Install `cargo-ndk` in the Android target CI script.

## Changelog

### DIFF
--- a/.buildkite/build-android-rust-target.sh
+++ b/.buildkite/build-android-rust-target.sh
@@ -23,6 +23,9 @@ esac
 # Only install the single target needed — no Apple targets or nightly toolchain
 rustup target add "$RUST_TARGET"
 
+echo "--- :package: Installing cargo-ndk"
+cargo install cargo-ndk --locked
+
 # This is a temporary step until we implement a more graceful way to handle missing credentials
 printf "site_url\nadmin_username\nadmin_password\nadmin_password_uuid\nsubscriber_username\nsubscriber_password\nsubscriber_password_uuid\n" > test_credentials
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Publish the Kotlin bindings' per-endpoint Markdown API reference as an `ai-docs` Maven classifier zip on `rs.wordpress.api:kotlin`, generated from the UniFFI bindings for agent/tooling consumption
 
+### Changed
+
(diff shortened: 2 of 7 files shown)


oguzkocer:
I've tested this in WPAndroid by using `wordpress-rs = '1432-4eb660caff7c4bafd38b2679ce3f6a77975e3cb5'` as the dependency which worked as expected. I have then compared the AARs before and after this change, and the results look identical. So, this seems like a safe change as far as I can tell, but would be great if the reviewer(s) also confirm it.

.buildkite/build-android-rust-target.sh
```diff
@@ -23,6 +23,9 @@ esac
 # Only install the single target needed — no Apple targets or nightly toolchain
 rustup target add "$RUST_TARGET"
 
+echo "--- :package: Installing cargo-ndk"
+cargo install cargo-ndk --locked
```
crazytonyli, line 27: Should we add the same command to the Makefile and/or Dockerfile?

crazytonyli (approved):

.buildkite/build-android-rust-target.sh
oguzkocer: That's a good question. I'll check tomorrow and address it if it looks like we should.

Btw, I'm going to convert this to a draft PR for now because AGP 9 update is introducing some new issues and I want to make sure we want to take this step. I'm 99% sure, but I might as well finish the AGP 9 update before we merge it.

oguzkocer: Added to `Makefile` in 78e8f8a6cc864b18c692a1f2b0023174a9c61162. We don't need it for the `Dockerfile` because NDK is only needed for Android targets which is not used for the Kotlin integration tests as they rely on the native builds instead. (the desktop version)

PR #1441, opened 2026-07-14
Bump and pin CI Rust toolchain to 1.97.0

### LINKED ISSUE #1436: Bump CI Rust toolchain so clippy catches everything newer versions flag
CI runs `clippy` inside a Docker image pinned to `rust:1.90.0`, so we don't flag lints that newer toolchains catch. We should bump the CI toolchain to current stable, fix whatever it surfaces, and pin the version so local and CI can't drift.

## Summary

- CI pins `public.ecr.aws/docker/library/rust:1.90.0` (`Makefile:7`); there is no `rust-toolchain.toml`, so local dev floats to whatever stable is installed while CI stays frozen.
- `clippy` adds lints and widens existing lint coverage across releases. Running an old toolchain in CI means real findings pass unflagged.
- Bump the CI image to current stable, fix the newly-surfaced findings, and pin the toolchain so local and CI can't drift.

## Evidence

`clippy::iter_kv_map` fires on newer toolchains but not on CI's `1.90.0`. Same code, same command — the only variable is the compiler:

| Toolchain | `cargo clippy --all -- -D warnings` on `wp_api/src/login/url_discovery.rs:121` |
| --- | --- |
| `1.90.0` (CI's pin) | ✅ `Finished`, no error |
| `1.95.0` (newer stable) | ❌ `error: iterating on a map's values` (`iter_kv_map`) |

The offending line is pre-existing (it dates to April 2025) and unrelated to any recent change — it simply isn't flagged by the toolchain CI runs.

## Proposed changes

- [ ] Bump `rust_docker_container` in `Makefile` from `1.90.0` to current stable.
- [ ] Run `cargo clippy --all -- -D warnings` and `cargo clippy --tests --all -- -D warnings` on the new toolchain and fix everything it surfaces (at minimum the `iter_kv_map` in `url_discovery.rs:121` → `.attempts.values().flat_map(|a| ...)`).
- [ ] Add a `rust-toolchain.toml` pinning the same version so local `cargo` and CI stay in lockstep and this drift stops recurring.
- [ ] Decide how we keep the toolchain current going forward — Dependabot doesn't bump the base image referenced in the `Makefile`, so this needs a deliberate cadence (a scheduled bump, or a CI check).

## Notes

Bumping the toolchain may surface more than the single lint above — the fix should enumerate the full set from a clean `clippy` run on the new version, not just the known one.

jkmassel:
## Description

CI runs `clippy` inside a Docker image pinned to `rust:1.90.0`, so real findings that newer toolchains flag pass unnoticed. There was no `rust-toolchain.toml`, so local dev floated to whatever stable was installed while CI stayed frozen. This bumps CI to current stable (`1.97.0`), fixes everything the newer `clippy` surfaces, pins the toolchain so local and CI can't drift, and adds a nightly check so the pin doesn't silently fall behind again.

Fixes #1436.

## Changes

- **Bumped the pinned toolchain `1.90.0` → `1.97.0`.** Added a single `rust_stable_toolchain` variable in `Makefile` that drives the Docker image tag, and bumped `wp_rs_web/Dockerfile` to match.
- **Added `rust-toolchain.toml`** pinning `1.97.0` (`clippy` + `rustfmt`, `minimal` profile) so local `cargo` and CI use the same compiler.
- **Pointed `setup-rust*` at the pinned toolchain** instead of the floating `stable`, so Apple target std installs onto the toolchain the pin actually selects — otherwise builds would fail with a missing `core`.
- **Fixed the 9 `clippy` findings** the new toolchain surfaces:
  - `iter_kv_map` — `wp_api/src/login/url_discovery.rs` (`.iter().flat_map(|(_, a)| …)` → `.values().flat_map(|a| …)`)
  - `manual_filter` — `wp_derive_request_builder/src/variant_attr.rs` (→ `Option::filter`)
  - `useless_borrows_in_formatting` ×7 — a redundant `&` in `xcframework/src/main.rs` and 6 in `comments_endpoint.rs` / `media_endpoint.rs` test cases
- **Added a nightly toolchain-currency check** (`.buildkite/commands/check-rust-toolchain-current.sh`, also `make check-rust-toolchain`) wired into `.buildkite/nightly.yml` as a `soft_fail` step. When the pin falls behind the latest stable it posts a Slack nudge; it no-ops on any network/parse failure so it never false-nags. Dependabot can't bump these strings, so this is the deliberate cadence.

## Test plan

- [x] `cargo clippy --all -- -D warnings` — clean on `1.97.0`
- [x] `cargo clippy --tests --all -- -D warnings` — clean on `1.97.0`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` — no additional findings
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo test -p wp_derive_request_builder` and `cargo test -p wp_api --lib` (1777 passed) — lint fixes are behavior-preserving
- [x] Currency-check script exercised for current / behind / offline; Slack payload validated; `Makefile` and `nightly.yml` parse-checked
- [ ] CI green on the new image

## Note for reviewers

After pulling this, the first `cargo` invocation auto-installs `1.97.0`, and anyone doing local Apple builds should re-run `make setup-rust` so the target std lands on the pinned toolchain.

## Changelog

### DIFF
--- a/.buildkite/commands/check-rust-toolchain-current.sh
+++ b/.buildkite/commands/check-rust-toolchain-current.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+
+# Warns when the Rust toolchain pinned in `rust-toolchain.toml` has fallen behind
+# the current stable release. Dependabot can't bump the toolchain (it lives in a
+# `rust-toolchain.toml` / `Makefile` string, not a package manifest), so this runs
+# on a schedule to give us a deliberate nudge to bump it.
+# See https://github.com/Automattic/wordpress-rs/issues/1436.
+#
+# Usage: check-rust-toolchain-current.sh [--notify-slack]
+#   --notify-slack  Post to Slack (needs SLACK_WEBHOOK): a reminder when the pin is
+#                   behind, or a low-urgency heads-up when the check can't run at all
+#                   (so a persistently broken check doesn't just go silently green).
+#                   Off by default so local `make` runs stay quiet.
+#
+# Exit codes:
+#   0 - the pin is current, OR we couldn't determine the latest version. We skip
+#       rather than fail on a transient network/parse issue; with --notify-slack a
+#       skip also posts a heads-up so a persistent failure stays visible.
+#   1 - the pin is behind the current stable release; time to bump it.
+
+set -uo pipefail
+
+NOTIFY_SLACK=false
+for arg in "$@"; do
+    case "$arg" in
+        --notify-slack) NOTIFY_SLACK=true ;;
+        *) echo "Unknown argument: $arg" >&2; exit 2 ;;
+    esac
+done
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+TOOLCHAIN_FILE="$REPO_ROOT/rust-toolchain.toml"
+
+STABLE_MANIFEST_URL="https://static.rust-lang.org/dist/channel-rust-stable.toml"
+
+pinned_version() {
+    # Extract e.g. `1.97.0` from `channel = "1.97.0"`.
+    grep -E '^[[:space:]]*channel[[:space:]]*=' "$TOOLCHAIN_FILE" \
+        | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' \
+        | head -1
+}
+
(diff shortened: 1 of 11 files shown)


oguzkocer (approved):
Looks good :shipit: 

It might have been slightly better to use fastlane instead of bash script for the toolchain check - or rather a combination of it, so the Slack notification and such wouldn't need to be repeated, but I don't feel strongly about it and didn't investigate it thoroughly.

PR #1445, opened 2026-07-16
Fetch posts and media by ID in batches of `5`

oguzkocer:
## Description

During metadata sync, posts and media are fetched by ID in batches. On sites that can't render a large batch within the request timeout, the whole batch failed. This lowers the fetch batch size from `100` to `5` for both posts and media so those sites can still sync — at the cost of more, smaller requests.

A fixed small batch is nearly as efficient as starting at `100` and backing off on timeouts, and it's far simpler. The adaptive alternative was explored in [#1444](https://github.com/Automattic/wordpress-rs/pull/1444) (degrade `100 → 20 → 5` on timeout, or the reverse — climb `5 → 20 → 100` and back off when a size times out); we can revisit it if a more dynamic approach proves worthwhile.

## Changes

- Lower `BATCH_FETCH_SIZE` from `100` to `5` in `PostService` and `MediaService` fetch-by-ID.

## Changelog

- [x] Added an entry under `## [Unreleased]` → `### Changed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Kotlin: `WpHttpClient.DefaultHttpClient` now applies more forgiving default OkHttp timeouts (connect 15s, read 60s, write 60s) instead of relying on OkHttp's 10s per-operation defaults, preventing premature timeouts when fetching large or slow-to-render responses. The values are configurable via a new `HttpClientTimeouts` type, accepted by both `DefaultHttpClient` and the `WpRequestExecutor` convenience constructor so callers can override them without building an `OkHttpClient` by hand.
+- Posts and media are now fetched by ID in batches of `5` instead of `100`, so sites that can't render a large batch within the request timeout can still sync (at the cost of more, smaller requests).
 - Swift: `WordPressAPI.uploadMedia` and `WpService.uploadMedia` now accept a `Progress` whose total is zero, defaulting the total to the upload file's byte size.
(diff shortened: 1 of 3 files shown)


crazytonyli:
> On sites that can't render a large batch within the request timeout, the whole batch failed.

This does not seem very likely to be caused by the batch size? Prepare the JSON payload should not take seconds. If a request does timeout, it's likely due to network connection, rather than the core REST API itself?

crazytonyli:
Reducing the batch size this much would compromise UX. There would be way more HTTP requests (which means user sees more loading state) to fetch the same amount of data.

oguzkocer:
> This does not seem very likely to be caused by the batch size? Prepare the JSON payload should not take seconds. If a request does timeout, it's likely due to network connection, rather than the core REST API itself?

@crazytonyli Unfortunately that's not the case at all. I've been investigating an issue where the `content.rendered` field for a site with [Katex plugin](https://wordpress.org/plugins/katex/) is taking a significant time to retrieve leading the timeout errors in WPAndroid.

> Reducing the batch size this much would compromise UX. There would be way more HTTP requests (which means user sees more loading state) to fetch the same amount of data.

This is not entirely correct either. The batch size is for fetching individual posts/media, so contrary to your point, the individual items will show up faster with this implementation. Instead of waiting one large request, the users would see items trickling in faster. You're correct that the total time will be more, although it'll be somewhat negligible in most cases.

Note that we are only making these type of requests for new & stale items, so the number of requests will only affect the initial load. And some servers, such as wp.com will have caching in place to serve it even faster. As such our goal is that the requests will actually succeed, rather than fetching the data in the most efficient way.

crazytonyli:
> The batch size is for fetching individual posts/media, so contrary to your point, the individual items will show up faster with this implementation.

Right. I misunderstood that for post list "page size". Thanks for explaining.

crazytonyli (approved):

PR #1446, opened 2026-07-17
Bump version to 0.6.0

crazytonyli:


oguzkocer (approved):
I think we should land https://github.com/Automattic/wordpress-rs/pull/1447 before cutting the release, so I've disabled the auto merge.

PR #1467, opened 2026-07-29
 Add WpService cache cleanup

crazytonyli:
## Description

Add a `WpService.remove_cached_data` to delete the stored cache. The app should use this function to delete cache upon logout or removing site.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Apps can now clear all cached data for an individual self-hosted or WordPress.com site without affecting other sites, supporting clean logout and site removal
 - WordPress.com `POST /me/transactions` endpoint for redeeming a shopping cart with the account's WordPress.com credits, completing a domain purchase
 - Publish the Kotlin bindings' per-endpoint Markdown API reference as an `ai-docs` Maven classifier zip on `rs.wordpress.api:kotlin`, generated from the UniFFI bindings for agent/tooling consumption
 - WordPress.com `POST /sites/<site_id>/domains/primary` endpoint for setting a site's primary domain
--- a/wp_mobile/src/service/sites.rs
+++ b/wp_mobile/src/service/sites.rs
@@ -112,6 +112,25 @@ impl SiteService {
             Ok(entity_id.db_site)
         })
     }
+
+    pub(crate) fn remove_cached_data(&self) -> Result<bool, WpServiceError> {
+        self.cache
+            .execute(|conn| -> Result<bool, SqliteDbError> {
+                let foreign_keys_were_enabled =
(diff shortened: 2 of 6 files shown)


wp_mobile/src/service/sites.rs
```diff
@@ -112,6 +112,25 @@ impl SiteService {
...
+
+                let deletion_result = SiteRepository.delete_site(conn, &self.db_site);
+
+                if !foreign_keys_were_enabled {
+                    conn.pragma_update(None, "foreign_keys", "OFF")?;
+                }
+                deletion_result
+            })
+            .map_err(Into::into)
+    }
```
oguzkocer, line 133: In my opinion, this is the wrong way to go about this. Regardless of where it's called from - or why - `SiteRepository.delete_site` needs to be doing the right thing by cleaning up everything related to it. If there is a gap there, that's something we should fix, but otherwise we shouldn't do a `foreign_keys` dance, which btw is a no-op at the moment because foreign keys are already on for this project and we can't actually disable them without us leaving orphans all over the place.

We should definitely guard against disabling foreign keys though, so let me look into that as a separate PR against `trunk`, so it doesn't end up happening.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Apps can now clear all cached data for an individual self-hosted or WordPress.com site without affecting other sites, supporting clean logout and site removal
```
oguzkocer, line 12: Might be worth referencing the new API here.

oguzkocer (changes requested):

wp_mobile/src/service/sites.rs
crazytonyli: Based on [this documentation](https://www.sqlite.org/foreignkeys.html#fk_enable) and test results, we need to explicitly call this pragma statement.

> Assuming the library is compiled with foreign key constraints enabled, it must still be enabled by the application at runtime, using the [PRAGMA foreign_keys](https://www.sqlite.org/pragma.html#pragma_foreign_keys) command.
> ...
> Foreign key constraints are disabled by default (for backwards compatibility), so must be enabled separately for each [database connection](https://www.sqlite.org/c3ref/sqlite3.html).

The reason for introducing this foreign_keys dance is to avoid overwriting the default behaviour. But I'm okay with simply calling the statement to enable foreign_keys on the function that creates the connection.

crazytonyli: I found the root cause now. The bundled sqlite3 has foreign_keys defaults to on. However, the WordPress iOS app links the system sqlite3, instead of the bundled sqlite3, and the system sqlite3 foreign_keys is off by default.

crazytonyli: I created an alternative PR here: https://github.com/Automattic/wordpress-rs/pull/1471

PR #1471, opened 2026-07-29
Enable foreign keys on cache connections

crazytonyli:
> [!Note]
> This is an alternative of https://github.com/Automattic/wordpress-rs/pull/1467

The existing `remove_cached_data` functions rely on foreign key constraints to delete all data associated with a given site. The foreign key constraint is enabled by default on the bundled sqlite3. However, the WordPress iOS app has a linker flag to link the system sqlite3, and ended up stripping off the bundled static sqlite3 in wordpress-rs out of the app binary. That means, in the iOS apps, the Rust library uses the iOS sqlite3 rather than the bundle sqlite3 at runtime. And, iOS sqlite3's foreign key constraint is not enabled by default.

This PR adds a `connection.pragma_update(None, "foreign_keys", "ON")` statement to all connections. I'm open to add a compilation condition to only compile this statement for Apple platforms. The reason I did not add it is because "do no compile this statement on Apple OS" is not the right condition. Because whether the end product uses the bundled sqlite3 is used or the system depends on the linker flag that the end product uses, not the OS.


crazytonyli:
Closing in favor of https://github.com/Automattic/wordpress-rs/pull/1486

PR #1487, opened 2026-08-05
Route Kotlin integration test dependencies through Reposilite

wzieba:
## Description

Route Gradle dependency resolution through internal proxy. We don't want to use Maven Central directly, as it often flags our shared EIP on CI as a heavy consumer, and starts to return `429` errors.

## Test plan

CI build is green. You can run the script locally.

## Changelog

### DIFF
--- a/scripts/reposilite-mirror.gradle.kts
+++ b/scripts/reposilite-mirror.gradle.kts
@@ -1,0 +1,84 @@
+// Copy of buildkite-ci's Gradle mirror init script. Keep in sync with it:
+// https://github.com/Automattic/buildkite-ci/blob/trunk/src/agents/_shared_/roles/android.install-gradle/files/mirror.gradle.kts
+//
+// Reposilite dependency mirror — redirects Gradle's dependency repositories
+// through our internal Reposilite pull-through cache. Cache misses then egress
+// once through the NAT gateway instead of every ephemeral agent re-resolving
+// from the upstreams, which is what triggers Maven Central's per-source-IP
+// `429 Too Many Requests` throttling.
+//
+// Two escape hatches, both evaluated up front — Gradle does NOT fall through to
+// another repository on transport errors (timeouts/5xx/connection-refused; it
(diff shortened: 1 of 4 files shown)


scripts/reposilite-mirror.gradle.kts
wzieba, line 1: I've added the script here because the job that uses it (`run-kotlin-integration-tests`) runs on `default` queue, and we only shipped this script to Android AMI.

This is a hotfix. The long-term solution will be to put this script to `default` queue as well and just reference it. But I want to address the problem quickly and without the "build AMI", time-consuming process.

scripts/reposilite-mirror.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+            isAllowInsecureProtocol = target.startsWith("http://")
+        }
+    }
+}
+
+when {
+    !mirrorEnabled ->
+        logger.lifecycle("ℹ️  Reposilite mirror disabled via REPOSILITE_MIRROR_ENABLED — using public repositories.")
+    !mirrorReachable() ->
+        logger.warn("⚠️  Reposilite unreachable at $reposilite — using public repositories.")
```
wzieba, line 63: This condition will be triggered locally. It adds maybe 2s - if it's not preferable, we can add a "is CI" check somewhere to exit earlier.

iangmaia (approved):
There are a couple of things in the script (Robolectric not needed for wordpress-rs I think? Potential better handling for http codes?) but given this is supposed to be temporary, perhaps we should already create a follow-up issue to handle it?

oguzkocer:
Thank you for the PR! I'm going to enable auto-merge on this so that CI for our other PRs will start working as soon as possible.

Note that we had some unrelated CI failures. This was due to one of our tests installing the `Newspack Newsletters` plugin's `3.38.1` version. The issue seems to be fixed in `3.38.2` which was released a few minutes ago, so the current retry should pass 🤞

PR #1490, opened 2026-08-06
docs: guide local xcframework builds away from the full 11-target build

dcalhoun:
## Description

Capturing practical local development guidance as someone less familiar with the `wordpress-rs` project after I consumed (and subsequently cleaned up) _nearly_ the entirety of my 1TB MacBook Pro disk space while working on #1488. **If any of this is unhelpful or incorrect, please suggest changes or removal.** 

`CLAUDE.md` currently recommends `make xcframework` to "verify UniFFI changes work correctly". That target builds **11 Apple targets across 4 platforms**, consumes ~100 GB in `target/`, and takes tens of minutes — when verifying a binding change locally only needs the macOS slice.

There was also no help text on any of the `xcframework*` targets, so `make help` gave no indication that `xcframework-only-macos` exists or that it is dramatically cheaper.

This adds a short doc covering how to pick a build, and wires the existing entry points to it.

## Changes

- **`Documentation/local-development.md`** (new) — a cost table for the xcframework targets, the caveat that `xcframework-only-<platform>` *replaces* the framework rather than adding a slice to it, and the fast path for verifying a UniFFI change across Rust, Swift, and Kotlin.
- **`Makefile`** — `Help:` lines for the `xcframework-only-*` and `xcframework-assemble` targets.
- **`README.md`** — links the new doc, and lists JDK 21 in the Kotlin dependency table (the Kotlin bindings pin `languageVersion 21`, which was previously undocumented).
- **`CLAUDE.md`** — points agents at `xcframework-only-macos` plus `swift build` instead of `make xcframework`.

Two things the doc calls out that cost real debugging time:

- `xcframework-only-<platform>` replacing the framework means running the `ios` variant and then `swift build` fails with `cannot find '...' in scope` for every symbol — which reads like a bindings problem but is a missing macOS slice.
- A partially failed `make xcframework` can still exit `0` and assemble a framework from whatever slices already exist in `target/`, producing stale binaries behind freshly generated Swift. The doc includes an `nm` check for confirming new symbols actually made it in.

`xcframework-all` deliberately gets a plain comment rather than a `Help:` line. The help extractor takes the last `# Help:` in an expanded recipe, so annotating it leaks that description onto every dependent target — `xcframework`, `xcframework-package`, and all five `test-swift-*`.

## Test plan

- [x] `make help` lists the new descriptions, and no unrelated target's help output changed
- [x] `make -n xcframework-only-macos` still expands correctly
- [x] No dangling internal links — every `Documentation/local-development.md` anchor referenced from `README.md` and `CLAUDE.md` resolves

Docs and Makefile comments only; no build behaviour changes.

## Changelog

Added under `### Changed` with the `**Internal:**` prefix, since this has no effect on the published libraries.


Documentation/local-development.md
```diff
@@ -0,0 +1,73 @@
...
+`make xcframework` builds **11 targets across 4 platforms** (macOS, iOS, tvOS,
+watchOS). It is what CI and releases need, but it is rarely what a local change
+needs — a full build consumes on the order of **100 GB** in `target/` and takes
+tens of minutes.
+
+Pick the smallest build that verifies your change:
+
+| Command                       | Targets | Approx. `target/` cost | Use when                                           |
+| ----------------------------- | ------- | ---------------------- | -------------------------------------------------- |
+| `cargo check -p wp_api`       | 0       | ~0                     | Verifying Rust compiles                            |
```
jkmassel: We probably want to check all packages – it takes a while the first time but should be pretty fast for subsequent checks.


```suggestion
| `cargo check`       | 0       | ~0                     | Verifying Rust compiles                            |
```

Documentation/local-development.md
```diff
@@ -0,0 +1,73 @@
...
+
+The tvOS and watchOS slices build with `-Z build-std` on the pinned nightly
+toolchain, which `make setup-rust` installs. If you skipped that step, those
+targets fail with `can't find crate for 'core'`.
+
+## Verifying a UniFFI change
+
+When adding or changing an exported function, the fast path is:
+
+1. **`cargo test -p wp_api --lib`** — the Rust side compiles and behaves.
```
jkmassel: We need to test `wp_api` *and* `wp_mobile`, at least. 

I'd suggest we run all unit tests, but no integration tests:


```suggestion
1. **`cargo test --lib`** — the Rust side compiles and behaves.
```

jkmassel (approved):
Generally good – couple of nits

Issue #1491, opened 2026-08-07
Swift executor drops URLSession timeouts to `GenericError` instead of `HttpTimeoutError`


jkmassel:
The Swift `WpRequestExecutor` never emits `RequestExecutionErrorReason.HttpTimeoutError` — a URLSession timeout (`URLError.timedOut`) has no branch in the catch dispatch and falls through to `.genericError`, so Apple-platform callers cannot distinguish a timeout from any other unclassified failure. reqwest and Kotlin both classify their equivalent, so `HttpTimeoutError` is effectively unconstructible on iOS/macOS.

## Summary

- `URLError.timedOut` (`NSURLErrorTimedOut`, -1001) is not handled in `WpRequestExecutor.perform(_:)` and lands in the `.genericError` fallback.
- reqwest maps `error.is_timeout()` → `HttpTimeoutError` (`wp_api/src/reqwest_request_executor.rs:192`); Kotlin maps `SocketTimeoutException` → `HttpTimeoutError` (`native/kotlin/.../WpRequestExecutor.kt:236`). Swift is the only executor that drops it. ❌
- Net: `RequestExecutionErrorReason.HttpTimeoutError` is unreachable on Apple platforms — a caller wanting "retry on timeout" UX has to fall back to matching `GenericError`, which also covers unrelated failures.

## Root Cause

`native/swift/Sources/wordpress-api/SafeRequestExecutor.swift` — `perform(_:)` classifies `errorIsHttpsError`, `errorIsNonExistentSiteError`, `errorIsDeviceIsOffline`, and `URLError.cancelled`, then falls through to `.genericError`. There is no `.timedOut` case, so timeouts are reported as generic. The executor's full vocabulary today is `{ cancellationError, genericError, invalidSslError, nonExistentSiteError, deviceIsOfflineError }`.

## Fix

Add a `.timedOut` branch alongside the existing `.cancelled` handling:

```swift
if let urlError = error as? URLError, urlError.code == .timedOut {
    return .failure(
        .RequestExecutionFailed(
            statusCode: nil,
            redirects: nil,
            reason: .httpTimeoutError,
            requestUrl: request.url(),
            requestMethod: request.method()
        )
    )
}
```

`.timedOut` is disjoint from the existing `URLError`-code sets, so ordering is safe, and this only reclassifies cases currently caught by the `.genericError` fallback — `isSiteUnreachable` / `isDeviceOffline` are unaffected.

## Caveat

[...]

PR #1523, opened 2026-08-07
Narrow `NonExistentSiteError` to DNS; add `ConnectionError` + `isConnectivityFailure`

### LINKED ISSUE #1495: `isSiteUnreachable` classifies a refused connection inconsistently across platforms
A connection that is *refused* — the host resolves via DNS, but nothing accepts the connection (server down, wrong port, service not listening) — is classified differently by each executor, so the `isSiteUnreachable` predicate exposed in [#1488](<https://github.com/Automattic/wordpress-rs/issues/1488>) returns a different answer per platform for the same failure.

## Summary

| Executor | Refused connection → reason | `isSiteUnreachable` |
| -- | -- | -- |
| Swift (`SafeRequestExecutor`) | `NonExistentSiteError` (via `URLError.cannotConnectToHost`) | `true` |
| Kotlin (`WpRequestExecutor`) | `HttpError` (via `ConnectException`) | `false` |
| `reqwest` (`ReqwestRequestExecutor`) | `HttpError` (via the io-error branch) | `false` |

Only a **DNS-resolution failure** is classified as `NonExistentSiteError` by all three, so it is the only portable `isSiteUnreachable` signal. A refused connection — a common, reachable failure (a site whose server is down while its domain still resolves) — diverges: `true` on iOS, `false` on Android and the Rust-native path.

Confirmed empirically: reqwest against a closed local port classifies as `HttpError` — the `as_io_error` branch intercepts `ECONNREFUSED` before the `is_connect` branch is reached.

## Why it matters

The point of exposing `isSiteUnreachable` ([#1488](<https://github.com/Automattic/wordpress-rs/issues/1488>)) is a uniform cross-platform signal. As-is, a consumer keying UX off it — "we couldn't reach your site" — fires on iOS but not on Android for an identical outage. Documenting the divergence (done in [#1493](<https://github.com/Automattic/wordpress-rs/issues/1493>)) mitigates but does not remove the footgun.

## The decision to make first

This is **not** a mechanical fix — it hinges on what `isSiteUnreachable` / `NonExistentSiteError` is supposed to *mean*, and the two readings lead to opposite fixes:

* **"The site's domain doesn't resolve"** (narrow) → a refused connection is *not* unreachable; it's a transport failure (`HttpError`), and **Swift** is the outlier to change.
* **"We couldn't establish a connection to the site, for any reason"** (broad) → a refused connection *is* unreachable, and **Kotlin/reqwest** are the outliers to change.

The semantic decision comes first — ideally with product input on what the apps actually want to surface to the user.

## Options (each a real tradeoff — not a recommendation)

**A — Narrow to DNS-only; align Swift to Kotlin/reqwest.** Move Swift `.cannotConnectToHost` out of the non-existent-site set → `HttpError`.

* Makes `isSiteUnreachable` a precise, portable "the domain is wrong/expired/mistyped" signal, and leaves `HttpError`/`HttpTimeoutError` to model transport failures as they already do.
* iOS `isSiteUnreachable` stops firing on refused connections — a behavior change (mitigated: the predicate is new/unmerged, so nothing depends on it yet).

**B — Broaden to any-connection-failure; align Kotlin/reqwest to Swift.** Map `ConnectException`/`NoRouteToHostException` (Kotlin) and refused (reqwest) → `NonExistentSiteError`.

* `isSiteUnreachable` becomes "couldn't connect at all."
* Conflates "wrong domain" with "server refused," blurring the line against `HttpError`/`HttpTimeoutError`; touches two executors.

**C — Leave the classification as-is; rely on the** [#1493](<https://github.com/Automattic/wordpress-rs/issues/1493>) **docs.** No behavior change.

* Each platform stays internally sensible; the predicate stays non-portable for refused connections and consumers must know to rely on the DNS-failure case alone.

**D — Add a distinct signal.** Keep `isSiteUnreachable` strictly DNS and add a broader *portable* predicate (e.g. `isConnectivityFailure`, unioning DNS + refused + no-route), or a dedicated reason.

* Gives both a precise and a broad signal; more API surface to design and thread through all three platforms.

## Careful-evaluation notes

* **Decide before** [#1488](<https://github.com/Automattic/wordpress-rs/issues/1488>) **ships.** The predicates are new and have no consumers yet, so the semantic is cheapest to set now; every release makes a later change more disruptive.
* **reqwest gotcha (learned the hard way in** [#1493](<https://github.com/Automattic/wordpress-rs/issues/1493>)**):** reqwest's `is_connect()` branch is *not* dead — it is the fallback for a DNS `NXDOMAIN` that didn't surface as a `ResolveError`, and it maps to `NonExistentSiteError`. A refused connection reaches `HttpError` via the earlier `as_io_error` branch instead. Any reqwest-side change (Option B) must not reclassify that DNS fallback, or it breaks `test_login_remote::login_spec_16_invalid_url`.
* **Kotlin offline coupling:** Kotlin's offline detection sits on the same DNS-failure path (`UnknownHostException` + `NetworkAvailabilityProvider`); refused connections (`ConnectException`) never consult the provider — so Option B wouldn't interact with offline detection, but that's worth confirming when the time comes.

## Provenance

Surfaced while reviewing [#1488](<https://github.com/Automattic/wordpress-rs/issues/1488>) (connectivity error classification) and documented in [#1493](<https://github.com/Automattic/wordpress-rs/issues/1493>). Independent of Automattic/wordpress-rs#1491 (Swift dropping timeouts) and Automattic/wordpress-rs#1492 (Kotlin misclassifying cancellation), which are separate classification gaps.

### LINKED ISSUE #1502: `HttpError { reason }` is unconstructible from the Swift executor
>

**Severity: medium.** The direct sibling of Automattic/wordpress-rs#1491's shape: the Swift executor's full output vocabulary is `{ cancellationError, genericError, invalidSslError, nonExistentSiteError, deviceIsOfflineError }` — no construction site for `.httpError` exists under `native/swift/Sources`. Protocol-level codes (`.badServerResponse`, `.cannotParseResponse`, `.cannotDecodeRawData`, `.cannotDecodeContentData`, `.zeroByteResource`, `.requestBodyStreamExhausted`, `.dataLengthExceedsMaximum`) all flatten into `GenericError`, while both peer executors use `HttpError` for exactly this class (reqwest: io/h2-level failures; Kotlin: `ConnectException` / `NoRouteToHostException`). An `errorIsHttpProtocolError` predicate mapping those codes to `.httpError(reason: error.localizedDescription)` closes the gap with no vocabulary change and keeps `GenericError` meaning "truly unclassified". Related Rust-side response-format issues: Automattic/wordpress-rs#606, [#182](<https://github.com/Automattic/wordpress-rs/issues/182>) (link, not duplicate).

jkmassel:
## Description

Stacked on #1493 — the top of the #1488 → #1493 connectivity stack. **Draft** for now.

Fixes #1495: a refused connection (the host resolves, but nothing is listening) was classified as `NonExistentSiteError` on Swift and `HttpError` on Kotlin/reqwest, so the `isSiteUnreachable` predicate from #1488 returned a different answer per platform for the same outage.

The semantic decision (per #1495): `NonExistentSiteError` / `isSiteUnreachable` means **the site's host did not resolve** — a DNS failure, nothing else. A refused/unreachable connection is a distinct failure with its own `ConnectionError` reason, and a new broad `isConnectivityFailure` predicate unions the two for a single "we couldn't reach the site's server" signal.

### 1. Narrow `NonExistentSiteError` to DNS-only (`f7d305b0`)
- Swift: `.cannotConnectToHost` no longer maps to `NonExistentSiteError`, so `isSiteUnreachable` no longer fires on a refused connection. `NonExistentSiteError` now means a DNS-resolution failure on every executor.

### 2. Add `ConnectionError` + `isConnectivityFailure` (`f8f3db63`)
- New `RequestExecutionErrorReason::ConnectionError { reason }` — the host resolved, but no connection to the server could be established.
- Reclassify connection-establishment failures to it across all three executors:
  - **reqwest** — `io::ErrorKind::{ConnectionRefused, HostUnreachable, NetworkUnreachable}`
  - **Kotlin** — `ConnectException`, `NoRouteToHostException`
  - **Swift** — `URLError.cannotConnectToHost`
- New `isConnectivityFailure` = `NonExistentSiteError ∪ ConnectionError`, exposed on Swift + Kotlin alongside `isSiteUnreachable`.
- Localized `connection_error` message (`en-US`; other locales fill in via the translation process).

**Connect timeout is deliberately excluded** — neither reqwest's `is_timeout()` nor OkHttp's `SocketTimeoutException` separates a connect timeout from a read timeout, so those stay `HttpTimeoutError`. A dark IP that times out is therefore not caught; that's #1491's territory.

## Test plan

- [x] Rust `cargo test --lib` — new `refused_connection_is_connection_error` (reqwest against a closed loopback port → `ConnectionError`) and `test_is_connectivity_failure_matches_dns_and_connection`; #1488's predicate tests still green (**1830** pass).
- [x] `cargo fmt`, `cargo clippy --tests --all-targets --all-features -D warnings`, `cargo doc` clean.
- [ ] Swift `testRefusedConnectionIsConnectionError` (needs Xcode + uniffi binding regen — CI).
- [ ] Kotlin executor reclassification (CI).
- [ ] Regenerate bindings (`make xcframework` / Kotlin uniffi) so the new variant + `isConnectivityFailure` export reach the native wrappers.

## Related issues

- Fixes #1495
- Stacked on #1493 (docs) → #1488 (predicates)
- Does **not** resolve #1502 — `HttpError` is still unconstructible from the Swift executor (refused connections now go to `ConnectionError`, not `HttpError`).
- #1491 (Swift connect-timeout → `GenericError`) is the remaining piece for the connect-timeout leg of `isConnectivityFailure`.

## Notes for review

- **New enum variant `ConnectionError`:** consumers with *exhaustive* matches on `RequestExecutionErrorReason` (Swift `switch`, Kotlin `when`-expression) will need to handle it. Everything here is unreleased (stacked on the unreleased #1488), so nothing ships broken — flag if you want a `**BREAKING:**` marker.
- Swift/Kotlin can't be compiled in my environment (needs uniffi binding regen); the native code is written against #1488's patterns and relies on CI to confirm.

## Changelog

- [x] Added `Added` + `Fixed` entries under `## [Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - WordPress.com `POST /sites/<site_id>/domains/primary` endpoint for setting a site's primary domain
 - WordPress.com `GET /sites/<site_id>/plans` endpoint for listing the plans a site can buy, priced for that site, with the plan it's currently on flagged.
 - `RequestExecutionErrorReason` gained `isSiteUnreachable` and `isDeviceOffline` for distinguishing a site that could not be reached (the host did not resolve) from a device with no network connection. Previously consumers had to match the `NonExistentSiteError` / `DeviceIsOfflineError` variants themselves. Available on both platforms as properties on the reason, which is reachable from `WpRequestResult.RequestExecutionFailed` and `WpApiException.RequestExecutionFailed` on Kotlin. Swift additionally exposes both as convenience properties on `WpApiError` and `RequestExecutionError`.
+- `RequestExecutionErrorReason` gained a `ConnectionError` variant — the host resolved, but no connection to the server could be established (the connection was refused, there was no route, or the host was unreachable) — and an `isConnectivityFailure` predicate (the union of `isSiteUnreachable` and `ConnectionError`) for a single portable "we couldn't reach the site's server" signal, distinct from `isDeviceOffline`. Refused and unreachable connections now classify as `ConnectionError` on every executor (previously the generic `HttpError` on Kotlin and reqwest). A connect timeout is not included; it remains `HttpTimeoutError`.
 
 ### Changed
 
@@ -32,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Swift now classifies a refused connection — the host resolves, but nothing is listening (server down, wrong port) — as `RequestExecutionErrorReason.HttpError`, matching the Kotlin and reqwest executors. iOS previously reported it as `NonExistentSiteError`, which now uniformly means a DNS-resolution failure across all three executors.
+- Swift no longer classifies a refused connection — the host resolves, but nothing is listening (server down, wrong port) — as `NonExistentSiteError`, so `isSiteUnreachable` no longer fires for it. `NonExistentSiteError` now uniformly means a DNS-resolution failure across all three executors (refused connections are a `ConnectionError`; see Added).
 
 ## [0.6.0] - 2026-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Publish the Kotlin bindings' per-endpoint Markdown API reference as an `ai-docs` Maven classifier zip on `rs.wordpress.api:kotlin`, generated from the UniFFI bindings for agent/tooling consumption
 - WordPress.com `POST /sites/<site_id>/domains/primary` endpoint for setting a site's primary domain
 - WordPress.com `GET /sites/<site_id>/plans` endpoint for listing the plans a site can buy, priced for that site, with the plan it's currently on flagged.
-- `RequestExecutionErrorReason` gained `isSiteUnreachable` and `isDeviceOffline` for distinguishing a site that could not be reached (most reliably, a DNS failure) from a device with no network connection. Previously consumers had to match the `NonExistentSiteError` / `DeviceIsOfflineError` variants themselves. Available on both platforms as properties on the reason, which is reachable from `WpRequestResult.RequestExecutionFailed` and `WpApiException.RequestExecutionFailed` on Kotlin. Swift additionally exposes both as convenience properties on `WpApiError` and `RequestExecutionError`.
+- `RequestExecutionErrorReason` gained `isSiteUnreachable` and `isDeviceOffline` for distinguishing a site that could not be reached (the host did not resolve) from a device with no network connection. Previously consumers had to match the `NonExistentSiteError` / `DeviceIsOfflineError` variants themselves. Available on both platforms as properties on the reason, which is reachable from `WpRequestResult.RequestExecutionFailed` and `WpApiException.RequestExecutionFailed` on Kotlin. Swift additionally exposes both as convenience properties on `WpApiError` and `RequestExecutionError`.
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


jkmassel:
Superseded by #1524. GitHub won't let me reopen this one — its base branch (`fix/converge-executor-error-classification`) was deleted when #1493 merged, and a closed PR with a deleted base can't be reopened or retargeted. #1524 is the same two commits rebased onto `trunk`.

PR #1524, opened 2026-08-07
Add `ConnectionError`; make `isSiteUnreachable` cover DNS + connection failures

### LINKED ISSUE #1495: `isSiteUnreachable` classifies a refused connection inconsistently across platforms

### LINKED ISSUE #1502: `HttpError { reason }` is unconstructible from the Swift executor

jkmassel:
## Description

Based on `trunk` — its stack dependencies **#1488** (predicates) and **#1493** (docs) have both merged. Supersedes closed #1523 (GitHub won't reopen a PR whose base branch was deleted on merge).

Fixes #1495: a refused connection (the host resolves, but nothing is listening) was classified as `NonExistentSiteError` on Swift and `HttpError` on Kotlin/reqwest, so the `isSiteUnreachable` predicate from #1488 returned a different answer per platform for the same outage.

The resolution: give refused/unreachable connections their own `ConnectionError` reason, and make **`isSiteUnreachable` mean "couldn't reach the site at all"** — a DNS failure *or* a failed connection. It's now the same answer on every executor. Callers that need to tell "wrong domain" from "server down" match the `NonExistentSiteError` / `ConnectionError` variants directly.

### 1. Narrow `NonExistentSiteError` to DNS-only
- Swift: `.cannotConnectToHost` no longer maps to `NonExistentSiteError`. `NonExistentSiteError` now means a DNS-resolution failure on every executor.

### 2. Add `ConnectionError`; broaden `isSiteUnreachable` to cover it
- New `RequestExecutionErrorReason::ConnectionError { reason }` — the host resolved, but no connection to the server could be established.
- Reclassify connection-establishment failures to it across all three executors:
  - **reqwest** — `io::ErrorKind::{ConnectionRefused, HostUnreachable, NetworkUnreachable}`
  - **Kotlin** — `ConnectException`, `NoRouteToHostException`
  - **Swift** — `URLError.cannotConnectToHost`
- `is_site_unreachable` now matches `NonExistentSiteError | ConnectionError`, so it's one portable "couldn't reach the site" signal (no separate predicate). Match the two variants directly to distinguish DNS from a failed connection.
- Localized `connection_error` message (`en-US`; other locales fill in via the translation process).

**Connect timeout is deliberately excluded** — neither reqwest's `is_timeout()` nor OkHttp's `SocketTimeoutException` separates a connect timeout from a read timeout, so those stay `HttpTimeoutError`. A dark IP that times out is therefore not caught; that's #1491's territory.

## Test plan

Refused-connection coverage added and passing on all three executors:

- [x] **Rust** — `refused_connection_is_connection_error` (reqwest against a closed loopback port → `ConnectionError`); `is_site_unreachable` now covers both DNS and connection failures; #1488's predicate tests updated. `cargo test --lib` **1829 pass**. `cargo fmt` / `clippy --tests --all-targets --all-features -D warnings` / `cargo doc` clean.
- [x] **Swift** — built the macOS bindings (`make xcframework-only-macos` + assemble) and ran `swift test --filter testRefusedConnectionIsConnectionError`: **passes** (closed loopback port → `.connectionError`). The full Swift package compiles clean against the regenerated bindings (generated case is `.connectionError`, `isConnectivityFailure` gone).
- [x] **Kotlin** — added `WpRequestExecutorTest` refused-connection test (`:api:kotlin:integrationTest` in Docker: **passes**, regenerating the bindings with `ConnectionError`). The new variant also required a `ConnectionError` branch in the example app's exhaustive `when` (`ErrorMessage.kt`, which had no `else`); verified with `:example:composeApp:compileKotlinDesktop` (**BUILD SUCCESSFUL**) — the same commonMain CI compiles via `:example:androidApp:assembleDebug`.

## Related issues

- Fixes #1495
- Follows the now-merged #1488 (predicates) and #1493 (docs); supersedes closed #1523
- Does **not** resolve #1502 — `HttpError` is still unconstructible from the Swift executor (refused connections now go to `ConnectionError`, not `HttpError`).
- #1491 (Swift connect-timeout → `GenericError`) is the remaining piece for the connect-timeout leg of `isSiteUnreachable`.

## Notes for review

- **`isSiteUnreachable` semantics changed** from #1488's DNS-only to the broad "couldn't reach the site." It's unreleased (merged today, no shipped consumers), so this is the cheapest moment to set it.
- **New enum variant `ConnectionError`** — marked `**BREAKING:**` in the changelog: adding a variant breaks exhaustive matches on `RequestExecutionErrorReason` (Swift `switch`, Kotlin `when`-expression, Rust `match`) for consumers on the last release (0.6.0).

## Changelog

- [x] Updated the `Added` entries and `Fixed` entry under `## [Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - WordPress.com `POST /sites/<site_id>/domains/primary` endpoint for setting a site's primary domain
 - WordPress.com `GET /sites/<site_id>/stats/post/<post_id>` endpoint for a post's view history, like count, comment count, and metadata
 - WordPress.com `GET /sites/<site_id>/plans` endpoint for listing the plans a site can buy, priced for that site, with the plan it's currently on flagged.
-- `RequestExecutionErrorReason` gained `isSiteUnreachable` and `isDeviceOffline` for distinguishing a site that could not be reached (the host did not resolve) from a device with no network connection. Previously consumers had to match the `NonExistentSiteError` / `DeviceIsOfflineError` variants themselves. Available on both platforms as properties on the reason, which is reachable from `WpRequestResult.RequestExecutionFailed` and `WpApiException.RequestExecutionFailed` on Kotlin. Swift additionally exposes both as convenience properties on `WpApiError` and `RequestExecutionError`.
+- `RequestExecutionErrorReason` gained `isSiteUnreachable` and `isDeviceOffline` for distinguishing a site that could not be reached (the host did not resolve, or a connection to it could not be established) from a device with no network connection. Previously consumers had to match the `NonExistentSiteError` / `DeviceIsOfflineError` variants themselves. Available on both platforms as properties on the reason, which is reachable from `WpRequestResult.RequestExecutionFailed` and `WpApiException.RequestExecutionFailed` on Kotlin. Swift additionally exposes both as convenience properties on `WpApiError` and `RequestExecutionError`.
+- **BREAKING:** `RequestExecutionErrorReason` gained a `ConnectionError` variant — the host resolved, but no connection to the server could be established (the connection was refused, there was no route, or the host was unreachable). Exhaustive matches over `RequestExecutionErrorReason` (Swift `switch`, Kotlin `when`, Rust `match`) must now handle the new case. `isSiteUnreachable` covers it alongside DNS failures, so it's a single portable "we couldn't reach the site" signal that returns the same answer on every executor; match `NonExistentSiteError` / `ConnectionError` directly to tell a bad domain apart from a server that's down. Refused and unreachable connections previously classified as the generic `HttpError` on Kotlin and reqwest. A connect timeout is not included; it remains `HttpTimeoutError`.
 
 ### Changed
 
@@ -42,14 +43,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Swift: `WpRequestExecutor.sleep(millis:)` converted milliseconds to nanoseconds with the wrong factor (`* 1_000` instead of `* 1_000_000`), so it slept 1000× too short — a `Retry-After: 30` waited 30 ms instead of 30 s. `RetryAfterMiddleware` then re-sent immediately, the server kept returning 429, and after `max_retries` the caller observed `MisconfiguredRateLimitError` where honoring the backoff would usually have succeeded. The executor now waits the full interval, and no longer risks a `fatalError` if the sleep's task is cancelled.
 - Swift: Classify invalid-SSL failures from the failed handshake's `SecTrust` (`URLError.failureURLPeerTrust`), via `SecTrustCopyCertificateChain`, instead of reading the undocumented `NSErrorPeerCertificateChainKey` `userInfo` string that has no public constant. Behavior is unchanged on every platform: iOS/macOS/tvOS still surface the presented certificate as `certificateNotValidForName`, and watchOS — which exposes no peer trust — still degrades to `genericSslError`. ([#1510](https://github.com/Automattic/wordpress-rs/issues/1510))
+- `isSiteUnreachable` now returns the same answer for a refused connection — the host resolves, but nothing is listening (server down, wrong port) — on every executor. Previously it was `NonExistentSiteError` on Swift (so `isSiteUnreachable` was `true`) but the generic `HttpError` on Kotlin and reqwest (so it was `false`); a refused connection is now a `ConnectionError` everywhere, which `isSiteUnreachable` covers. `NonExistentSiteError` is reserved for a DNS-resolution failure.
 
 ### Security
 
 - **Internal:** Bumped the transitive `rand` dependency to `0.9.3` and `0.8.6` to clear [RUSTSEC-2026-0097](https://rustsec.org/advisories/RUSTSEC-2026-0097.html) (`GHSA-cq8v-f236-94qc`), a low-severity unsoundness in `rand` 0.9.2 / 0.8.5. Lockfile-only; the affected code path (a custom `log` logger calling `rand::rng()` during reseed) is not exercised here.
-
-### Fixed
-
-- Swift now classifies a refused connection — the host resolves, but nothing is listening (server down, wrong port) — as `RequestExecutionErrorReason.HttpError`, matching the Kotlin and reqwest executors. iOS previously reported it as `NonExistentSiteError`, which now uniformly means a DNS-resolution failure across all three executors.
 
 ## [0.6.0] - 2026-07-16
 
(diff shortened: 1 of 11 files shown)


crazytonyli (approved):
The Swift side looks good to me.

However, I have a feeling that we the error cases are too granular. _If_ all we do with these errors are presenting error messages, the final fallback `.RequestExecutionFailed(reason: .genericError(errorMessage: error.localizedDescription))` might help us the most, potentially the user too. The message comes from `URLError`, and they'd see the same error message in Safari and the app when connecting to their site.

jkmassel:
> The message comes from URLError, and they'd see the same error message in Safari and the app when connecting to their site.

The tricky thing is providing user support – we want to be able to differentiate what's going on because an unregistered domain (ie DNS failure) looks suspiciously similar to a "there's no server running at that IP address" error. Being able to tell the difference makes it much easier to point the user in the right direction.

crazytonyli:
Yeah, I mentioned that using URLError error message might help us the most because I thought that'd help us most in terms providing support.

We usually "reverse" the error message that user sends us to the concrete errors in the code. If user sees URLError error message, then we'd easily map them to the exact URLError error code (I presume the error codes and error messages are one-to-one).

Right now, we map a selected few URLError to custom error messages, which means when we get the custom error messages, we'd get a few possible error codes, rather than pin point to one specific error.

For example, when user says they got an "site does not exists" error, the best we can do is one of three possible erorrs. But if the error message says "DNS Lookup failed", we'd know for sure there is something wrong with their DNS configuration.

```
func errorIsNonExistentSiteError()
        [
            .badURL,
            .cannotFindHost,
            .dnsLookupFailed
        ]
        .contains((error as? URLError)?.code)
```

PR #1526, opened 2026-08-07
Model the per-post stats response against the wp.com source

oguzkocer:
## Description

Follow-up to #1489, targeting `per-post-stats` branch.

The per-post stats endpoint has no public documentation, so the response types were worked out from live responses. Comparing them against the endpoint's implementation in `wpcom` — `class.wpcom-json-api-stats-post-views-v1-1-endpoint.php` and `stats_get_post()` in `wp-content/mu-plugins/stats.php` — turned up a field the API sends that wasn't modelled, three types that are wider than what the API can produce, and several doc comments describing behaviour the source contradicts.

Two of those could only be settled against the live API, so four responses were captured from a test site: one post with views, two never-viewed posts, and the site's home page.

### Types

- **`StatsPostTarget`** replaces `PostId` as the endpoint's path argument, with `Post { id }` and `HomePage` variants. The API addresses the home page as post `0`, which isn't a valid `PostId` anywhere else in the crate. `stats_utm` takes a typed path argument the same way.
- **`average`, `averages.overall` and `averages.months` are now `u64`.** The endpoint casts all three to `int` before sending them, so a fractional value can't reach the wire — confirmed across all four captured responses. `change` is left as `f64`; it's computed from the uncast average and is the one genuine float in the response.
- **`post_author` is now `WpComUserId`** — on a WordPress.com site the row carries the account's global id, not the site-scoped one `UserId` names. This needed a `deserialize_u64_or_string_as_t` in `wp_serde_helper`, the counterpart to the existing `i64` variant.
- **`post_date_gmt` is now `WpGmtDateTime`**. `wp_utc_date_format` already parses the MySQL datetime format the API uses, so no new deserializer was needed.
- **`From<PostId> for StatsPostTarget`** resolves the home page id, so callers working from a list that includes it — `/stats/top-posts` reports one — don't each repeat the check.

### Fields the API sends

- **`permalink`** — `stats_get_post()` attaches a computed permalink to the post row, and it reaches the wire as the row's last key. Callers previously had no way to get a post's URL from this response. Typed `Option<String>` and read through `deserialize_false_as_none`: it's derived per request rather than read from a column, so an unexpected shape shouldn't cost the whole response.
- **`post_modified_gmt`** — the unambiguous counterpart to `post_modified`, which is in the site's local timezone.
- **`post_excerpt`**.

### Documentation

- **`PostId(0)` doesn't always mean the home page.** The endpoint branches on the site's `show_on_front` setting: with a "latest posts" front page the figures are the home page's own views, but with a static front page they're the whole site's view history. The response is identical in both cases, so the distinction is documented on `StatsPostResponse`.
- **`highest_week_average` is neither weekly nor an average.** It's the highest single-day view count across the last few weeks. `highest_day_average` is the highest monthly average of daily views.
- **`years` and `averages` start at 1970 for a target with no views.** With no view to anchor on, the endpoint reports every year from 1970 to the present, each with an empty `months` map — 57 entries for a post published last year. Verified on all three never-viewed posts captured.
- **`daily_views` is padded, not sparse.** A never-viewed post still gets one zero-count entry per day since publication.

### Serialization

`StatsPostResponse` deserializes from the API's `fields`/`data` column table but derived `Serialize` over the flattened shape, so serializing a response produced JSON it could not read back. It now serializes into the column table as well, matching how `StatsPostChange` handles its own wire form.

### `wp_serde_helper`

`deserialize_false_as_none` mapped every boolean to `None`, including `true`. Its siblings — `deserialize_false_or_string`, `deserialize_false_or_string_or_null`, `deserialize_u64_or_none` — all reject `true` explicitly, so it now does too. Its doc also listed a missing field as yielding `None`, which `deserialize_with` cannot do without `#[serde(default)]`; that line is removed.

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- WordPress.com `GET /sites/<site_id>/stats/post/<post_id>` endpoint for per-post stats. Returns the post's view history, like and comment counts, and post metadata — everything the "Latest Post Summary" card needs. The API's `fields`/`data` column table is flattened into `daily_views` while deserializing; callers wanting a trailing window slice its tail, which can run to thousands of entries. Passing `PostId(0)` returns stats for the site's home page, for which `post`, `discussion`, and `like_count` are `None`.
 - WordPress.com `POST /me/transactions` endpoint for redeeming a shopping cart with the account's WordPress.com credits, completing a domain purchase
 - WordPress.com `GET /sites/<site_id>/purchases` endpoint for listing a site's purchases (plans, domains, and other subscriptions)
 - Publish the Kotlin bindings' per-endpoint Markdown API reference as an `ai-docs` Maven classifier zip on `rs.wordpress.api:kotlin`, generated from the UniFFI bindings for agent/tooling consumption
 - WordPress.com `POST /sites/<site_id>/domains/primary` endpoint for setting a site's primary domain
+- WordPress.com `GET /sites/<site_id>/stats/post/<post_id>` endpoint for a post's view history, like count, comment count, and metadata
 
 ### Changed
 
--- a/wp_api/src/wp_com/endpoint/stats_post_endpoint.rs
+++ b/wp_api/src/wp_com/endpoint/stats_post_endpoint.rs
@@ -1,13 +1,15 @@
 use crate::{
-    posts::PostId,
     request::endpoint::{AsNamespace, DerivedRequest},
-    wp_com::{WpComNamespace, WpComSiteId, stats_post::StatsPostResponse},
+    wp_com::{
+        WpComNamespace, WpComSiteId,
+        stats_post::{StatsPostResponse, StatsPostTarget},
+    },
 };
 use wp_derive_request_builder::WpDerivedRequest;
 
 #[derive(WpDerivedRequest)]
 enum StatsPostRequest {
-    #[get(url = "/sites/<wp_com_site_id>/stats/post/<post_id>", output = StatsPostResponse)]
+    #[get(url = "/sites/<wp_com_site_id>/stats/post/<stats_post_target>", output = StatsPostResponse)]
     GetStatsPost,
 }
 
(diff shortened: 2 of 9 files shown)


nbradbury:
@oguzkocer Thanks for tackling this! I'll approve this to unblock it, but Claude's review did find some issues, and the first one involving `permalink` likely needs to be addressed.

[review-pr-1526-per-post-stats-2026-08-08.pdf](https://github.com/user-attachments/files/30855378/review-pr-1526-per-post-stats-2026-08-08.pdf)

Also, I noted the "PostId(0) doesn't always mean the home page" remark and wasn't sure what it meant - does the client need to change how it gets stats for the home page?

nbradbury (approved):
:shipit:

oguzkocer:
@nbradbury Thank for the review. I've addressed the ones that made sense and also updated the PR description to reflect some of the changes.

> Also, I noted the "PostId(0) doesn't always mean the home page" remark and wasn't sure what it meant - does the client need to change how it gets stats for the home page?

The documentation in the PR description seems somewhat clear and somewhat confusing at the same time: (shared as excerpt below)

> The endpoint branches on the site's `show_on_front` setting: with a "latest posts" front page the figures are the home page's own views, but with a static front page they're the whole site's view history.

 I don't know if anything needs to be changed in WPAndroid as a result of this. I suggest having a conversation with Claude on this with the context you have related to WPAndroid implementation.

nbradbury:
For completeness, here's what Claude had to say about home page stats:

How we retrieve home page stats doesn’t need to change. But the caveat is real and worth recording.

What the comment is pointing at. The PostId(0) note is about interpretation, not retrieval. From the doc on StatsPostResponse in rs trunk:

StatsPostTarget::HomePage requests post 0 … What the view figures cover then depends on how the site’s front page is configured, and the two cases are indistinguishable in the response:

a “latest posts” front page — the views recorded against the home page
a static front page — the whole site’s view history
There’s no alternative endpoint, parameter or id for the second case — the request is /stats/post/0 either way. So there is nothing to change about how we fetch it. Our data source already does the only correct thing: it passes through whatever id the Posts & Pages row carries, and maps 0 to StatsPostTarget.HomePage.

It’s also not a regression. Old stats routes that same row’s id 0 to the same endpoint via PostDetailStore, so both screens have always shown whatever WP.com returns for post 0. Nothing this branch did changed that.

On the From<PostId> for StatsPostTarget they added “so callers working from a list that includes it don’t each repeat the check” — that’s a plain Rust trait impl with no #[uniffi::export], and trait impls don’t cross the FFI boundary. I confirmed the generated Kotlin has no equivalent helper. So my if (postId == HOME_PAGE_POST_ID) HomePage else Post(postId) is the Kotlin counterpart, not a duplicate of something available.

Issue #1507, opened 2026-08-07
`presentedHostnames` carries only the leaf certificate's CN, omitting SANs


jkmassel:
> Split out from Automattic/wordpress-rs#1497 — *Swift executor URLSession error audit*. Section C — SSL payload and platform gaps.
>
> File references are to `native/swift/Sources/wordpress-api/SafeRequestExecutor.swift` unless noted; `:NNN` line numbers were verified against `fix/converge-executor-error-classification`.

**Severity: medium.** `handleHttpsError` builds `presentedHostnames` as `[siteCertificate.commonName()]` even though the parsed `SslCertificateInfo` exports `alternativeNames()` over UniFFI. Modern certificates carry their identities in SANs — CN is often empty or unrelated — so the payload can present an empty or misleading name list. reqwest forwards the full presented-names list (`presented.to_vec()`). One-line fix; `LoginTests.swift` pins the CN-only shape and needs updating. Relates to Automattic/wordpress-rs#657 (that issue is test coverage; this is the payload), and becomes load-bearing if Automattic/wordpress-rs#1498's host-vs-names comparison lands.

jkmassel:
Resolved by #1531 (merged 2026-08-12).

The Swift executor no longer builds `presentedHostnames` from `[siteCertificate.commonName()]`. `handleHttpsError` now uses `siteCertificate.presentedHostnames()` (`SafeRequestExecutor.swift:219`), and the new `presented_hostnames()` on `SslCertificateInfo` (`wp_api/src/ssl.rs:105`) returns the Common Name (when present) followed by the de-duplicated SANs — so the payload carries the full presented-names set on every executor, and stays correct for SAN-only certificates with no CN. The `LoginTests.swift` CN-only assertion was de-pinned in the same PR.

Closing as resolved. The classification concern in #1498 and the trust-acceptance concern in #1512 are separate and remain open.

Issue #1498, opened 2026-08-07
Every TLS failure with a peer certificate chain becomes `CertificateNotValidForName` — expired, self-signed, and unknown-root included


jkmassel:
> Split out from Automattic/wordpress-rs#1497 — *Swift executor URLSession error audit*. Section A — Converted to the wrong Rust error.
>
> File references are to `native/swift/Sources/wordpress-api/SafeRequestExecutor.swift` unless noted; `:NNN` line numbers were verified against `fix/converge-executor-error-classification`.

**Severity: high.** `errorIsHttpsError` (`:233-246`) matches five codes — `.secureConnectionFailed`, `.serverCertificateUntrusted`, `.serverCertificateHasBadDate`, `.serverCertificateNotYetValid`, `.serverCertificateHasUnknownRoot` — but `handleHttpsError` (`:172-208`) never inspects `urlError.code`: chain present → `.certificateNotValidForName`, chain missing → `.genericSslError`. An expired or self-signed certificate issued for the **correct** hostname — the common misconfigured self-hosted case — is reported as a hostname mismatch, often with `hostname == presentedHostnames[0]`, a self-contradictory payload. ❌

reqwest reserves `CertificateNotValidForName` for rustls's actual `NotValidForNameContext` and maps every other TLS error away from it (`wp_api/src/reqwest_request_executor.rs:230-247`). ✅

This also compounds with Automattic/wordpress-rs#1512: an app seeing `certificateNotValidForName` may offer the `allowSSL(altNames:forCommonName:)` remediation (`:75-77`), steering users into trusting an expired/untrusted certificate they believed was a mere name mismatch.

**Fix:** map `.serverCertificateHasBadDate` / `.serverCertificateNotYetValid` / `.serverCertificateHasUnknownRoot` to `.genericSslError` (they are by definition not name problems). For the remaining two codes, only claim `certificateNotValidForName` when the request host is genuinely absent from the certificate's CN + SANs — `SslCertificateInfo` already exposes both (`wp_api/src/ssl.rs:73-79`).

**Caveat:** genuine name mismatches surface as `.serverCertificateUntrusted` on Darwin, so a per-code switch alone would break the existing mismatch test (`LoginTests.swift`, `testInvalidHTTPsFails`) — the host-vs-names comparison is required to keep true mismatches while fixing the false positives.

PR #1531, opened 2026-08-07
fix: parse Common-Name-less certificates and report their real hostnames

### LINKED ISSUE #1508: A CN-less (SAN-only) leaf certificate is unparseable, so the intermediate CA's CN can be reported as the presented hostname
> Split out from Automattic/wordpress-rs#1497 — *Swift executor URLSession error audit*. Section C — SSL payload and platform gaps.
>

**Severity: medium.** `parse_certificate` requires a subject CN (the `?` at `wp_api/src/ssl.rs:23`) even though SANs parse independently. `getPeerCertificateChain` compactMaps the `nil` away, and `handleHttpsError` then treats element 0 of the surviving array as the site certificate — with a CN-less leaf and parseable intermediates, `presentedHostnames` becomes the intermediate's CN (e.g. `["R10"]`). The same parse failure means `allowSSL` exceptions silently stop matching CN-less leaves. CA/Browser Forum direction is to omit CN, so frequency will grow. Fix: tolerate a missing subject CN in the Rust parser, and stop assuming index 0 survived the compactMap.

jkmassel:
## Description

Fixes a bug where a leaf certificate that omits its subject Common Name — a SAN-only certificate — failed to parse at all, so a TLS name-mismatch error reported the wrong hostname to the user. On Swift the executor could report an unrelated intermediate CA's Common Name (e.g. `R10`) as the certificate's presented hostname; on Kotlin it reported nothing. Modern certificates increasingly carry their identities only in Subject Alternative Names — CA/Browser Forum direction is to drop the Common Name — so this only gets more frequent over time.

Fixes #1508.

## Root cause

`parse_certificate` (`wp_api/src/ssl.rs`) required a subject Common Name via `?`, even though the SANs parse independently. A certificate with no CN therefore returned `None` from the whole parse. Downstream:

- **Swift** — `getPeerCertificateChain` `compactMap`ped the unparseable leaf away, then `handleHttpsError` treated element 0 of the *surviving* array as the site certificate. With a CN-less leaf and parseable intermediates, element 0 is the intermediate CA, so `presentedHostnames` became the CA's name rather than the site's.
- **Kotlin** — the same parse failure left `certificates.first()?.commonName()` null, so `presentedHostnames` was an empty list.
- A self-signed SAN-only certificate — the misconfigured-site case the SSL-exception machinery exists for — also has no CN on its *issuer*, which `parse_certificate` required too (a second `?`).

## Changes

- **`wp_api/src/ssl.rs`**: Make `common_name` on `SslCertificateInfo` and its `issuer` `Option<String>` (`String?` on Swift/Kotlin) so a missing CN no longer fails the parse. Add `presentedHostnames()`, returning the Common Name (when present) followed by the SANs, de-duplicated, so every executor reports the same set.
- **`SafeRequestExecutor.swift`**: Parse the leaf directly from element 0 of the *raw* chain instead of the `compactMap` survivor, so an unparseable leaf degrades to `genericSslError` rather than promoting the intermediate into the site's slot. Build `presentedHostnames` from `presentedHostnames()`. Handle the now-optional CN in `alternateNames(forCertificate:)`.
- **`WpRequestExecutor.kt`**: Report `certificates.firstOrNull()?.presentedHostnames()` instead of only the leaf's Common Name.
- **Tests**: Add a Rust unit test for a synthetic self-signed SAN-only cert (no CN on subject or issuer) and one for the real `no-common-name.badssl.com` leaf. De-pin the existing Swift name-mismatch assertion from the CN-only shape, and add live specs against badssl — `no-common-name.badssl.com` (SAN-only) on Swift and `wrong.host.badssl.com` (name mismatch) on Swift and Kotlin.
- **`CHANGELOG.md`**: Add a `Fixed` entry and a `**BREAKING:**` note for the `Option<String>` change.

This also subsumes the SAN-omission half of #1507: once the CN is optional, `[commonName()]` had to change, and the correct replacement is CN + SANs.

**Not in this PR:** `allowSSL(altNames:forCommonName:)` is keyed on the Common Name, so a truly CN-less certificate still can't be allow-listed — there is no key to register it under. The parse no longer fails silently, and the limitation is documented in `alternateNames(forCertificate:)`; widening that trust-acceptance path belongs with #1512, which questions the path's missing chain validation.

## Test plan

- [x] `cargo test -p wp_api --lib ssl::` — 6 unit tests pass. `parses_san_only_certificate_without_common_name` (a synthetic self-signed cert whose subject *and* issuer carry only `O=Test Org`) and `parses_real_world_no_common_name_certificate` (the real `no-common-name.badssl.com` leaf: no subject CN, one SAN) both panic against the old `?`-based parser, which returned `None`.
- [x] `cargo fmt --all -- --check` and `cargo clippy -p wp_api --tests --all-features -- -D warnings` are clean.
- [x] Swift — `swift test` runs two live specs, both green: `no-common-name.badssl.com` (SAN-only) asserts `presentedHostnames == ["no-common-name.badssl.com"]` (the bug reported the COMODO issuer CA here), and `wrong.host.badssl.com` (name mismatch) asserts the payload contains the `badssl.com` SAN, not only the `*.badssl.com` CN.
- [x] Kotlin — `:api:kotlin:integrationTest --tests "*ApiUrlDiscoveryTest.testNameMismatchReportsAllPresentedNames"` against `wrong.host.badssl.com` passed (1 test, 0 failures). The CN-less endpoint isn't exercised on Kotlin: its certificate is expired, which OkHttp raises as `SSLHandshakeException` (not the `SSLPeerUnverifiedException` that routes to certificate inspection), and the inspection re-connect would fail on the expiry — the Rust unit test covers that parse.
- [x] Swift and Kotlin bindings regenerate and compile against the new `commonName() -> String?` / `presentedHostnames()` (`make xcframework-only-macos` + `swift build`; `:api:kotlin:compileKotlin` / `detekt`).

## Changelog

- [x] I've added an entry to `CHANGELOG.md` under `## [Unreleased]`, using the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) categories. Prefixed the breaking change with `**BREAKING:**`.


native/swift/Tests/wordpress-api/LoginTests.swift
```diff
@@ -341,6 +346,91 @@ class LoginTests {
...
+                #if os(watchOS) // watchOS doesn't make the underlying certificate available to us
+                guard case .genericSslError = underlyingReason else {
+                    Issue.record("The underlying error must be `genericSslError`")
+                    return false
+                }
+                #else
+                guard case .certificateNotValidForName(_, let presentedHostnames) = underlyingReason else {
+                    Issue.record("The underlying error must be `certificateNotValidForName`")
+                    return false
+                }
```
oguzkocer, line 384: I think this assertion will start failing when #1498 is addressed in case you want to leave a breadcrumb here.

native/swift/Tests/wordpress-api/LoginTests.swift
```diff
@@ -341,6 +346,91 @@ class LoginTests {
...
+
+                // The leaf carries no Common Name and exactly one SAN, so that SAN
+                // is the entire presented-hostname list. The bug reported the
+                // COMODO issuer CA's name here instead.
+                #expect(presentedHostnames == ["no-common-name.badssl.com"])
```
oguzkocer, line 389: It looks like we are mixing approaches between this and `presentedHostnames.contains` the PR uses in other places. The comment around the `presentedHostnames.contains` usage makes sense to me, should we convert this equality to a contains assertion as well or comment why we prefer equality here?

wp_api/src/ssl.rs
```diff
@@ -95,3 +130,113 @@ impl From<x509_cert::time::Time> for WpGmtDateTime {
...
+        );
+        assert_eq!(cert.issuer.common_name, None);
+        assert_eq!(cert.issuer.organization, Some("Test Org".to_string()));
+    }
+
+    #[test]
+    fn presented_hostnames_prepend_common_name_and_dedupe() {
+        // CN present and also repeated in the SANs: the CN leads, and the
+        // duplicate SAN is dropped.
+        let cert = parse_certificate(der(CERT_WITH_CN)).unwrap();
```
oguzkocer: Nitpick: I don't feel strongly about this, but we have been preferring `expect` over `unwrap`, per your suggestion. I think some PRs have broken that already, but figured I'd bring it to your attention.

There is a similar issue a few lines below.

oguzkocer (approved):
I have left a few nitpicks, but even if you skip all, I don't think it'd be a big loss. :shipit:

native/swift/Tests/wordpress-api/LoginTests.swift
jkmassel: Addressed in https://github.com/Automattic/wordpress-rs/pull/1531/commits/97d841097c02a99057c8397e9bdccec069116cfd

native/swift/Tests/wordpress-api/LoginTests.swift
jkmassel: It's safe for the purposes of this test, but `contains` is safer in general for the app-level code.

wp_api/src/ssl.rs
jkmassel: addressed in ee132498

PR #1533, opened 2026-08-08
Document `GET /all-domains/` subtypes and correct its fixtures

oguzkocer:
## Description

`DomainSubtypeId::DefaultAddress` was documented as "Free WordPress.com address". That understates it — the server files staging and garden subdomains under the same subtype — and the gap is load-bearing, because `DefaultAddress` is the set that `no_wpcom=true` excluded from the v1.1 endpoint. v1.2 declares only `garden`, and the API drops undeclared query parameters before the endpoint runs, so a client that wants the v1.1 behaviour has to filter on this subtype. Nothing in the crate said so, which sent the WordPress-Android integration down a wrong turn.

While documenting the subtypes it turned out two fixture entries claim subtypes `GET /all-domains/` cannot return, so the tests asserted behaviour that does not exist:

- **`site_redirect`** — the endpoint queries only mapped and parked domains, and `Domain_List_Item` has no site-redirect branch, so this subtype is never emitted. It is still valid on the per-site domain endpoints.
- **`domain_mapping`** — not one of the five subtypes the server defines. It deserialized to `Other("domain_mapping")`, quietly passing. Connected domains are `domain_connection`.

## Changelog


wp_api/src/wp_com/domains.rs
```diff
@@ -345,9 +345,24 @@ pub struct DomainTransferInfo {
 }
 
 /// Optional query parameters for `GET /all-domains/` (v1.2).
+///
+/// `garden` is the only parameter this endpoint version accepts. In
+/// particular there is no equivalent of the `no_wpcom` parameter that
+/// `GET /all-domains/` (v1.1) took, and the API discards undeclared query
+/// parameters before the endpoint sees them, so sending one has no effect.
+/// To list only the domains added to a site, filter the response on
+/// [`DomainSubtypeId::DefaultAddress`].
```
jkmassel:
```suggestion
/// To list only the domains added to a site, exclude the items whose
/// subtype is [`DomainSubtypeId::DefaultAddress`] (this reproduces
/// v1.1's `no_wpcom=true`).
```

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **BREAKING:** `product_type` fields on `Product` and `WPComProduct` changed from `String` to `ProductType`. Callers that match on or construct these values will need to wrap/unwrap with `ProductType(...)`.
 - **BREAKING:** The cache now enables SQLite foreign key enforcement on every connection it prepares, and fails with `SqliteDbError::ForeignKeysUnavailable` if the setting doesn't take effect. Removing a site relies on `ON DELETE CASCADE` to clear its cached rows, so on builds where enforcement defaulted to off those rows were silently left behind.
 - **BREAKING:** `ShoppingCart.coupon` changed from `String` to `CouponCode`, and `ShoppingCartCostOverride.override_code` from `String` to `CostOverrideCode`, so the shopping cart and site plans describe these values with the same types. Callers will need to wrap/unwrap with `CouponCode(...)` / `CostOverrideCode(...)`.
+- Documented `GET /all-domains/` subtypes and parameters. `DomainSubtypeId::DefaultAddress` covers staging and garden subdomains as well as the free WordPress.com address, and is the set v1.1's `no_wpcom=true` excluded; v1.2 has no equivalent parameter, so clients filter on this subtype instead.
```
jkmassel:
```suggestion
- Documented `GET /all-domains/` subtypes and parameters. `DomainSubtypeId::DefaultAddress` covers staging and garden subdomains as well as the free WordPress.com address, and is the set v1.1's `no_wpcom=true` excluded; v1.2 has no equivalent parameter, so clients filter this subtype out instead.
```

wp_api/src/wp_com/domains.rs
```diff
@@ -427,6 +453,9 @@ pub enum DomainSubtypeId {
     /// Domain transfer in progress.
     DomainTransfer,
     /// Site redirect to another URL.
+    ///
+    /// `GET /all-domains/` does not currently return these: it lists only
+    /// mapped and parked domains, and site redirects are neither.
```
jkmassel:
```suggestion
    /// `GET /all-domains/` does not emit this subtype — its
    /// `Domain_List_Item` has no site-redirect branch.
```

jkmassel (approved):
I left some slight clarification suggestions.

Also I don't know if I understand what a garden subdomain is? 😅

oguzkocer:
@jkmassel Thanks for the review!

The garden concept seems to be limited in scope to specific clients. It's probably not something WP or JP will ever care about, so I considered removing it entirely. However, our hope, if I am not mistaken, is that our other clients might use `wordpress-rs` in the future, so I think it makes sense to include it.

Issue #1497, opened 2026-08-07
Swift executor URLSession error audit: wrong conversions, unchecked codes, and crash-instead-of-error paths


jkmassel:
An audit of the Swift `WpRequestExecutor`'s URLSession error handling, in the same vein as Automattic/wordpress-rs#1491 and Automattic/wordpress-rs#1492: errors we convert to the **wrong** Rust error, URLError codes we could classify better but don't, and adjacent paths where a failure crashes instead of classifying at all. Filed as one umbrella issue for validation; individual findings can be split out once confirmed.

All findings exclude Automattic/wordpress-rs#1491 (`.timedOut`), Automattic/wordpress-rs#1492 (Kotlin cancellation), and the platform differences already documented in the `is_site_unreachable` / `is_device_offline` rustdoc (`.cannotConnectToHost`, `.badURL`). File references are to `native/swift/Sources/wordpress-api/SafeRequestExecutor.swift` unless noted; line numbers were verified against `fix/converge-executor-error-classification`.

## Provenance

Surfaced from a multi-agent audit (🤖 six independent review lenses over the executor, findings merged and then each checked by two adversarial verification passes against the code). 44 raw findings reduced to 23 confirmed; 2 were refuted during verification and 3 late additions could not be verified — both groups are listed at the bottom so they aren't silently dropped. Severities are provisional pending human validation.

---

### 1\. Every TLS failure with a peer certificate chain becomes `CertificateNotValidForName` — expired, self-signed, and unknown-root included

**Severity: high.** `errorIsHttpsError` (`:233-246`) matches five codes — `.secureConnectionFailed`, `.serverCertificateUntrusted`, `.serverCertificateHasBadDate`, `.serverCertificateNotYetValid`, `.serverCertificateHasUnknownRoot` — but `handleHttpsError` (`:172-208`) never inspects `urlError.code`: chain present → `.certificateNotValidForName`, chain missing → `.genericSslError`. An expired or self-signed certificate issued for the **correct** hostname — the common misconfigured self-hosted case — is reported as a hostname mismatch, often with `hostname == presentedHostnames[0]`, a self-contradictory payload. ❌

reqwest reserves `CertificateNotValidForName` for rustls's actual `NotValidForNameContext` and maps every other TLS error away from it (`wp_api/src/reqwest_request_executor.rs:230-247`). ✅

[...]

PR #1534, opened 2026-08-08
fix: read the SSL peer certificate chain from the documented SecTrust key

### LINKED ISSUE #1510: Peer chain extraction relies on the undocumented `"NSErrorPeerCertificateChainKey"` string
> Split out from Automattic/wordpress-rs#1497 — *Swift executor URLSession error audit*. Section C — SSL payload and platform gaps.
>

**Severity: low.** `getPeerCertificateChain` reads a hard-coded CFNetwork-internal key (`:299`) with no public constant. The repo's own tests document that watchOS doesn't populate it, so every watchOS SSL failure already degrades to `.genericSslError`; if Apple stops setting the key elsewhere, iOS/macOS silently degrade the same way. The documented `NSURLErrorFailingURLPeerTrustErrorKey` (a `SecTrust`) + `SecTrustCopyCertificateChain` — the same API the delegate override already uses at `:354` — is the sturdier source and may restore detail on watchOS.

jkmassel:
## Description

`getPeerCertificateChain` in the Swift executor classified invalid-SSL failures by reading `NSErrorPeerCertificateChainKey` — an undocumented CFNetwork `userInfo` string with no public constant. If Apple stops setting it, the executor silently loses the presented certificate and every SSL failure collapses to `.genericSslError`.

This reads the `SecTrust` stored under Apple's documented `NSURLErrorFailingURLPeerTrustErrorKey` instead, and derives the presented certificate chain with `SecTrustCopyCertificateChain` — the same API the authentication-challenge delegate already uses to read the server's certificate.

Fixes #1510. Split out from #1497 (Swift executor URLSession error audit).

## Changes

- **`native/swift/Sources/wordpress-api/SafeRequestExecutor.swift`**: `getPeerCertificateChain` now pulls the `SecTrust` from `NSURLErrorFailingURLPeerTrustErrorKey` and maps `SecTrustCopyCertificateChain` over it. A `CFGetTypeID(...) == SecTrustGetTypeID()` check gates the force-cast to `SecTrust` (see below). The Linux path (`return []`) is unchanged.
- **`CHANGELOG.md`**: `### Fixed` entry under `[Unreleased]`.

No test changes. The existing `LoginTests` Example 17 already asserts the exact per-platform behavior this preserves — `certificateNotValidForName` with the presented hostname on iOS/macOS/tvOS, `genericSslError` on watchOS — so it doubles as the regression test for this change.

## Why not `as? SecTrust`

A conditional downcast from `Any` to a CoreFoundation type such as `SecTrust` **always succeeds without a runtime check** — the compiler rejects it outright (`conditional downcast to CoreFoundation type 'SecTrust' will always succeed`) and points at `CFGetTypeID`. `as?` would let a non-`SecTrust` value through and crash inside `SecTrustCopyCertificateChain`. The `CFGetTypeID` comparison is the checked equivalent; the force-cast that follows is safe because the type is already verified.

## watchOS: the documented key doesn't help either, and that's fine

The documented `SecTrust` is no more available on watchOS than the undocumented key was. I confirmed this on CI ([build 6091](https://buildkite.com/automattic/wordpress-rs/builds/6091)) by temporarily tightening Example 17 to require the certificate on every platform: iOS, macOS, and tvOS passed; watchOS failed with `.invalidSslError(.genericSslError)`. So watchOS behavior is unchanged, `LoginTests` keeps its `#if os(watchOS)` branch, and the benefit here is a documented source on the three platforms that do surface the certificate — not a watchOS improvement.

## Test plan

- [x] Example 17 (`testInvalidHTTPsFails`) passes on macOS via the `.certificateNotValidForName` branch — `make xcframework-only-macos` + `swift test --filter testInvalidHTTPsFails` (1 test, 0.45s). It drives a real mismatch: `https://wordpress-1315525-4803651.cloudwaysapps.com` presents a certificate for `vanilla.wpmt.co`, and the assertions confirm `hostname == "wordpress-1315525-4803651.cloudwaysapps.com"` and `presentedHostnames == ["vanilla.wpmt.co"]`.
- [x] Per-platform behavior confirmed on CI build 6091 (strict-assertion probe, since reverted): iOS/macOS/tvOS surface `.certificateNotValidForName`; watchOS returns `.genericSslError` — matching the `#if os(watchOS)` branch the test keeps.
- [x] `swift-format lint --strict` and `swiftlint` (pinned rules) clean on the changed source.

## Changelog

### DIFF
--- a/native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
+++ b/native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
@@ -290,13 +290,14 @@ public final class WpRequestExecutor: SafeRequestExecutor {
     }
 
     private func getPeerCertificateChain(_ error: Error) -> [SslCertificateInfo]? {
-        let nserror = error as NSError
-        let info = nserror.userInfo
-
         #if os(Linux) // Linux doesn't support `SecCertificate`
         return []
         #else
-        guard let certChainArray = info["NSErrorPeerCertificateChainKey"] as? [SecCertificate] else {
+        // The certificate chain the server presented during the failed TLS handshake.
+        guard
+            let trust = (error as? URLError)?.failureURLPeerTrust,
+            let certChainArray = SecTrustCopyCertificateChain(trust) as? [SecCertificate]
+        else {
             return nil
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Internal:** Documented how to pick an xcframework build for local work, and added `make help` descriptions for the `xcframework-only-*` targets. Verifying a UniFFI change needs only `make xcframework-only-macos`, not the full 11-target `make xcframework`.
 - **Internal:** Fixed `make xcframework-only-<platform>` building the per-target libraries but never assembling them into the xcframework. The `@# Help:` comments added for those `make help` descriptions became each rule's recipe and silently shadowed the shared `xcframework-only-%` pattern rule that ran the assemble step; each rule now runs the assemble step directly.
 
+### Fixed
+
+- Swift: Classify invalid-SSL failures from the failed handshake's `SecTrust` (`URLError.failureURLPeerTrust`), via `SecTrustCopyCertificateChain`, instead of reading the undocumented `NSErrorPeerCertificateChainKey` `userInfo` string that has no public constant. Behavior is unchanged on every platform: iOS/macOS/tvOS still surface the presented certificate as `certificateNotValidForName`, and watchOS — which exposes no peer trust — still degrades to `genericSslError`. ([#1510](https://github.com/Automattic/wordpress-rs/issues/1510))
+
 ### Security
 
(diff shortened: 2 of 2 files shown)


native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -296,7 +296,23 @@ public final class WpRequestExecutor: SafeRequestExecutor {
...
         #else
-        guard let certChainArray = info["NSErrorPeerCertificateChainKey"] as? [SecCertificate] else {
+        // Read the presented certificate chain from the `SecTrust` that the failed
+        // handshake stores under `NSURLErrorFailingURLPeerTrustErrorKey`.
+        //
+        // A conditional `as?` cast to a CoreFoundation type such as `SecTrust` always
+        // succeeds without a real runtime check, so verify the type with `CFGetTypeID`
+        // before force-casting.
+        guard
+            let peerTrust = info[NSURLErrorFailingURLPeerTrustErrorKey],
```
crazytonyli: Use `(error as? URLError)?.failureURLPeerTrust` instead? https://developer.apple.com/documentation/foundation/urlerror/failureurlpeertrust

crazytonyli (approved):

PR #1535, opened 2026-08-08
fix: Swift sleep(millis:) slept 1000× too short, breaking Retry-After backoff

### LINKED ISSUE #1513: `sleep(millis:)` sleeps 1000× too short, breaking every Retry-After backoff — and `try!`s
> Split out from Automattic/wordpress-rs#1497 — *Swift executor URLSession error audit*. Section D — Crash instead of classified error.
>

**Severity: medium.** `try! await Task.sleep(nanoseconds: millis * 1000)` (`:230`) — nanoseconds require `millis * 1_000_000`, so a `Retry-After: 30` sleeps 30ms. `RetryAfterMiddleware` re-sends immediately, the server keeps returning 429, and after `max_retries` the caller observes `MisconfiguredRateLimitError` where honoring the backoff would typically have succeeded. Swift is the only executor that doesn't actually wait (reqwest uses `Duration::from_millis`, Kotlin uses `delay(millis)`). ❌ The `try!` additionally converts a cancelled sleep into `fatalError` — currently latent (uniffi 0.32 never cancels the backing Task), but a loaded gun. Fix is one line: `try? await Task.sleep(nanoseconds: millis * 1_000_000)`.

jkmassel:
## Description

Every `Retry-After` backoff on Swift finished almost instantly instead of waiting. `WpRequestExecutor.sleep(millis:)` converted milliseconds to nanoseconds with the wrong factor — `millis * 1_000` instead of `millis * 1_000_000` — so it slept **1000× too short**: a `Retry-After: 30` waited 30 ms, not 30 s. `RetryAfterMiddleware` then re-sent immediately, the server kept returning 429, and after `max_retries` the caller observed `MisconfiguredRateLimitError` where honoring the backoff would usually have succeeded. Swift was the only executor that didn't actually wait — reqwest uses `Duration::from_millis`, Kotlin uses `delay(millis)`.

The same method's `try!` was a second, latent problem: a cancelled sleep would `fatalError`. It's unreachable today (uniffi 0.32 never cancels the backing task) but a loaded gun.

Split out from #1497 (Swift executor URLSession error audit). Fixes #1513.

## Changes

- **Fix the conversion** in `WpRequestExecutor.sleep(millis:)` — now `try? await Task.sleep(for: .milliseconds(millis))`. `try?` replaces `try!`, so a cancelled sleep returns instead of crashing.
- **Fix the identical copy** in the `HTTPStubs` test mock, which had its own `try! await Task.sleep(nanoseconds: millis * 1_000)`. It's now a no-op.
- **Add a regression test** (`SafeRequestExecutorTests.swift`) that calls the real `WpRequestExecutor.sleep(millis: 200)` and asserts it waited at least 150 ms.

**`Task.sleep(for:)` replaces the manual nanosecond math so the bug can't recur.** Same result as the issue's suggested `nanoseconds: millis * 1_000_000`, but the `Duration` API takes milliseconds directly, won't trap on `UInt64` multiplication overflow, and matches how reqwest and Kotlin express the same wait. It's available on every deployment target (iOS 16 / macOS 13 / tvOS 16 / watchOS 9) and already used elsewhere in the Swift test suite.

**The mock is a no-op, not a faithful sleep.** The rate-limit test asserts error classification, not timing, and routes through the mock; a faithful sleep would add ~3 s of real wall-clock wait (3 retries × the 1 s `Retry-After`) for nothing it checks. This mirrors the no-op `sleep` in the Rust middleware unit tests, and the production sleep's timing is covered by the new regression test instead.

## Test plan

- [x] Full Swift unit suite via `BUILDKITE=1 swift test` (reproduces the CI-macOS config, excluding the Docker integration target): **77 tests in 18 suites passed**, build clean with 0 warnings.
- [x] New regression test `sleep(millis:) waits for approximately the requested duration` passes after **0.218 s** (it waited ~200 ms). Reintroducing the `millis * 1_000` conversion makes it fail — `Expectation failed: (elapsed → 0.00127625 seconds) >= (.milliseconds(150) → 0.15 seconds)` — so the assertion catches the exact bug.
- [x] `Login Spec Example 15: Rate Limited that never succeeds` still classifies as `MisconfiguredRateLimitError`, and ran at its baseline (~3.5 s, in line with its sibling login-flow tests) — confirming the no-op mock added no real sleep.

## Changelog

- [x] I've added an entry to `CHANGELOG.md` under `## [Unreleased]` (`Fixed`).

### DIFF
--- a/native/swift/Tests/wordpress-api/SafeRequestExecutorTests.swift
+++ b/native/swift/Tests/wordpress-api/SafeRequestExecutorTests.swift
@@ -1,0 +1,27 @@
+import Foundation
+import Testing
+@testable import WordPressAPI
+@testable import WordPressAPIInternal
+
+@Suite("SafeRequestExecutor")
+struct SafeRequestExecutorTests {
+
+    // Regression test for #1513: `sleep(millis:)` converted milliseconds to nanoseconds with the
+    // wrong factor (`millis * 1_000` instead of `* 1_000_000`), so it slept 1000× too short and
+    // `RetryAfterMiddleware` never actually honored a `Retry-After` backoff.
+    @Test("sleep(millis:) waits for approximately the requested duration")
+    func testSleepHonorsMillisecondDuration() async {
+        let executor = WpRequestExecutor(urlSession: .shared)
+
+        let requestedMillis: UInt64 = 200
+        let clock = ContinuousClock()
+        let start = clock.now
+        await executor.sleep(millis: requestedMillis)
+        let elapsed = start.duration(to: clock.now)
+
+        // The old bug slept ~0.2 ms for a 200 ms request. `Task.sleep` waits *at least* the
+        // requested duration, so a 150 ms lower bound sits comfortably above the buggy value and
+        // below the target — catching the regression without flaking on scheduling jitter.
+        #expect(elapsed >= .milliseconds(150))
+    }
+}
--- a/native/swift/Tests/wordpress-api/SafeRequestExecutorTests.swift
+++ b/native/swift/Tests/wordpress-api/SafeRequestExecutorTests.swift
@@ -19,9 +19,11 @@ struct SafeRequestExecutorTests {
         await executor.sleep(millis: requestedMillis)
         let elapsed = start.duration(to: clock.now)
 
-        // The old bug slept ~0.2 ms for a 200 ms request. `Task.sleep` waits *at least* the
-        // requested duration, so a 150 ms lower bound sits comfortably above the buggy value and
-        // below the target — catching the regression without flaking on scheduling jitter.
+        // `Task.sleep` waits *at least* the requested duration, so bound both sides. The 150 ms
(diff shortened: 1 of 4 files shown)


native/swift/Tests/wordpress-api/SafeRequestExecutorTests.swift
```diff
@@ -0,0 +1,27 @@
...
+        let requestedMillis: UInt64 = 200
+        let clock = ContinuousClock()
+        let start = clock.now
+        await executor.sleep(millis: requestedMillis)
+        let elapsed = start.duration(to: clock.now)
+
+        // The old bug slept ~0.2 ms for a 200 ms request. `Task.sleep` waits *at least* the
+        // requested duration, so a 150 ms lower bound sits comfortably above the buggy value and
+        // below the target — catching the regression without flaking on scheduling jitter.
+        #expect(elapsed >= .milliseconds(150))
```
oguzkocer, line 26: It might be worth bounding it in both directions, so we don't accidentally start waiting too long.

oguzkocer (approved):

PR #1544, opened 2026-08-10
fix: Swift large multipart uploads leaked their temp file

### LINKED ISSUE #1540: Swift: large (>10 MB) multipart uploads leak their temp file on the success path
Found while reviewing #1537 (the multipart failed-read crash fix). Not introduced by that PR — this is pre-existing (from #715 / #1042).

## Problem

For a multipart form estimated over 10 MB (or built with `forceWriteToFile: true`), `multipartFormDataStream` serializes the body to a UUID-named temp file under `FileManager.default.temporaryDirectory` (`native/swift/Sources/wordpress-api/MultipartForm.swift:100`) and returns `.onDisk(URL)`. `SafeRequestExecutor.upload(...)` hands that URL to `session.uploadTask(with:request, fromFile:)` (`native/swift/Sources/wordpress-api/SafeRequestExecutor.swift:654`).

`URLSession` treats the `fromFile:` source as caller-owned — it reads the file (and requires it to persist until the transfer finishes) but never deletes or moves it. Nothing in `upload(...)`, `completionHandler`, or `notifyTaskResult` removes it either. So every large media upload (4K video, ProRAW, multi-file posts) leaves a temp file behind that only the OS reclaims, and not promptly.

#1537 added temp-file cleanup for the *failure* path it introduced (`MultipartForm.swift:132-134`), but the common *success* path still leaks.

## Fix

Delete the file after the upload completes, in `SafeRequestExecutor.upload(...)`. It can't be deleted inside `multipartFormDataStream` because `uploadTask(fromFile:)` reads it asynchronously — the file must outlive serialization. A `defer` keyed on `if case let .onDisk(file) = body`, after `result.get()`, is the natural spot.

## Severity

Medium. Bounded — the OS purges the temporary directory eventually — but it accumulates real disk over a session of large uploads.

jkmassel:
## Description

For a multipart form estimated over 10 MB (or built with `forceWriteToFile: true`), `multipartFormDataStream` serializes the body to a UUID-named temp file under `FileManager.default.temporaryDirectory` and returns `.onDisk(URL)`, which `upload(...)` hands to `URLSession.uploadTask(with:fromFile:)`. `URLSession` treats a `fromFile:` source as caller-owned — it reads the file for the duration of the transfer but never deletes or moves it — and nothing in `upload(...)`, the completion handler, or `notifyTaskResult` removed it either. So every large media upload (4K video, ProRAW, multi-file posts) left a temp file behind that only the OS reclaims, and not promptly — on success, failure, and cancellation alike.

Pre-existing (from #715 / #1042), not introduced by #1537. #1537 added cleanup only for the *serialization*-failure path it introduced (a half-written temp file when a stream read fails mid-body); the *upload* itself still leaked on every outcome. Found while reviewing #1537.

Fixes #1540.

## Root Cause

`URLSession.uploadTask(with:fromFile:)` requires the file to persist until the transfer finishes but never cleans it up. The temp file must therefore outlive serialization — it can't be deleted inside `multipartFormDataStream`, because the upload reads it asynchronously — so cleanup belongs to the caller that awaits completion: `upload(...)`, whose `defer` runs only after the completion handler has resumed the continuation, i.e. after `URLSession` is done reading the file, on every path.

## Changes

- **native/swift/Sources/wordpress-api/MultipartForm.swift**: Add `MultipartFormContent.removeTemporaryFileIfNeeded()` — deletes the `.onDisk` temp file (best-effort `try?`, matching the existing `SecureEnclavePasswordTransformer` cleanup precedent), a no-op for `.inMemory`. Idempotent, so it's safe to arm more than once.
- **native/swift/Sources/wordpress-api/SafeRequestExecutor.swift**: Add `defer { body.removeTemporaryFileIfNeeded() }` at the top of `upload(...)`, so the temp file is removed once the transfer finishes on any path — success, failure, or cancellation. `WpMultipartFormRequest.perform` arms the same cleanup where serialization *acquires* the file, so a future edit that throws between serialization and the `upload(...)` call can't silently reintroduce the leak (the helper is idempotent, so the two defers don't double-free). `upload(...)` is now a file-scope function (it used no `self`), which also lets a test drive it directly against a stubbed `URLSession`. Behavior is otherwise unchanged.
- **native/swift/Tests/wordpress-api/MultipartFormTests.swift**: New tests — a cross-platform suite for the cleanup helper (serializes a real on-disk body via `multipartFormDataStream(forceWriteToFile: true)` and asserts the file is created, then removed), and a macOS/iOS-only suite (`#if os(macOS) || os(iOS)`) that drives `upload(...)` through a `URLProtocol` stub across a successful, failed, and cancelled transfer. The success stub asserts the temp file **still exists while the transfer is in flight**, so the test verifies the file *outlives* the transfer — not merely that it's gone afterward.
- **CHANGELOG.md**: `Fixed` entry.

## Test plan

- [x] `BUILDKITE=1 swift test` — full unit suite passes (97 tests across 21 suites) on macOS.
- [x] Mutation check (leak): removing the `defer` makes the upload tests fail their `!fileExists` assertion — they genuinely guard the leak.
- [x] Mutation check (premature delete): deleting the temp file *before* the transfer completes fails the **success** test with `fileDoesNotExist` via the in-flight stub assertion; neutering that assertion makes it pass again — confirming the test guards the too-early-delete regression the `perform`-level defer protects against, not just leaks.
- [x] `xcrun swift-format lint --strict` clean on the changed files; SwiftLint clean under the pinned 0.57.1 config.
- [ ] CI green on all Apple platforms (iOS/macOS/tvOS/watchOS) and Linux.

## Changelog


crazytonyli:
Is this PR ready for view? I see I'm assigned as reviewer, but the status is in draft.

jkmassel:
> Is this PR ready for view? I see I'm assigned as reviewer, but the status is in draft.

Ready now @crazytonyli, thanks!!

crazytonyli (approved):

PR #1546, opened 2026-08-10
Classify multipart mid-read failures as `MediaFileUnreadable`

### LINKED ISSUE #1541: Swift: multipart mid-read failures surface as .genericError, dropping the file path and media classification
Found while reviewing Automattic/wordpress-rs#1537. Part of the Swift executor error audit (Automattic/wordpress-rs#1497).

## Problem

A media file that fails at *field construction* maps to `RequestExecutionError.MediaFileNotFound(filePath:)` (`native/swift/Sources/wordpress-api/SafeRequestExecutor.swift:628`), which carries the path and has a dedicated binding type (Kotlin `WpRequestResult.MediaFileNotFound<T>`, rendered as `"File not found: $filePath"` in the example app).

A media file that fails *mid-read* (the case Automattic/wordpress-rs#1537 fixed) throws `MultipartFormError.inaccessibleFile`, which is neither a `RequestExecutionError` nor a `URLError`, so it falls through `WpRequestExecutor.perform`'s catch to the `.genericError` catch-all (`SafeRequestExecutor.swift:117-125`). That bucket also collects `MultipartFormError.impossible` and any other unclassified failure, and it carries only the stream error's `localizedDescription` (e.g. "Input/output error") — no file path.

So the same real-world condition (an unusable media file) reaches callers two different ways, and the mid-read path can't be distinguished from an unrelated generic failure.

This is a deliberate choice in Automattic/wordpress-rs#1537: a mid-read failure means the file exists but can't be read right now, so reusing `MediaFileNotFound` would be semantically wrong. The trade-off is the signal loss above.

## Options

* A dedicated `MediaFileUnreadable { file_path }` variant on `RequestExecutionError` (`wp_api/src/api_error.rs`) — stays honest and carries the path, but is a cross-language UniFFI change (Rust enum + regenerated Swift/Kotlin bindings).
* At minimum, thread `file.filePath` through so the message names the file. `MultipartFormField` currently discards the source path (it keeps only a basename), so this needs a small change there too.

## Severity

Low–medium. A quality/consistency issue layered on a correct fix — crash → classified-error-carrying-the-real-cause is already a strict improvement.

jkmassel:
## Summary

- Classifies a media file that fails **mid-read** during multipart serialization as a dedicated `MediaFileUnreadable { file_path }` error, instead of the path-less `.genericError` that #1537 left it as. Fixes #1541.
- Distinct from `MediaFileNotFound` (the file couldn't be opened at all), and carries the file path so callers can name it.
- Stacked on #1537 → #1545 — base is `issue-1542` until they merge; the diff here is only this change.

## Changes

- **Rust** (`wp_api/src/api_error.rs`, `wordpress_org/client.rs`): new `MediaFileUnreadable { file_path }` on `RequestExecutionError` and `WpApiError`, with the `From` / `status_code` / localization wiring. New `media_file_unreadable` FTL string (en-US only; the `.pot` and other locales are regenerated by the nightly sync).
- **Swift** (`MultipartForm.swift`, `SafeRequestExecutor.swift`): `MultipartFormField` retains the full source path (it kept only the basename before); a mid-read maps to `MediaFileUnreadable` via a testable `MultipartFormError.asRequestExecutionError` helper. The path is also threaded through #1545's stream-open-error throw, so a file that can't be opened at serialization time classifies as `MediaFileUnreadable` too. Construction-time failures are unchanged (`MediaFileNotFound`).
- **Kotlin** (`WpRequestExecutor.kt`, second commit): the file part now streams through a `MediaFileRequestBody` that tags a *read* failure (via an Okio `ForwardingSource`) as `MediaFileUnreadableException`, while a *write* (socket) failure stays an ordinary `IOException`; `executeRequestSafely` maps the tagged exception to `MediaFileUnreadable`. Bindings regenerated; the four exhaustive `when`s and the example render handle the new case.

## Test plan

- [x] `cargo test --lib -p wp_api -p wp_localization` and `cargo check --workspace` — green.
- [x] `BUILDKITE=1 swift test` — 86 tests (5 new): the serialization error carries the source path for a file field and `nil` for an in-memory field; `asRequestExecutionError` maps a path-carrying failure to `MediaFileUnreadable` and a path-less one to `nil` (→ `.genericError`); and `fileAtPath:` retains the full path.
- [x] Kotlin `:api:kotlin:compileKotlin :api:kotlin:test :api:kotlin:detektMain :api:kotlin:detektTest`, plus `compileIntegrationTestKotlin` and the example `compileKotlinDesktop`. New `MediaFileRequestBodyTest` (4 tests): a source read failure is tagged with the file path, a successful read passes through, `writeTo` streams a real file unchanged, and a file that's gone by upload time surfaces as `MediaFileUnreadable` with its path.
- No `MediaFileUnreadable` integration test on either platform: a genuine mid-read is non-deterministic against a live server (a missing file reports `hasBytesAvailable == false` / opens fine then can't fail on demand), so both sides unit-test the read-failure mapping via an injected failing stream/source rather than a live upload.

## Related issues

- Fixes #1541. Stacked on #1537 → #1545 (base `issue-1542`). Part of the Swift executor error audit (#1497).

## Changelog

- [x] Added under `## [Unreleased]` → `### Changed`, prefixed `**BREAKING:**`.


crazytonyli:
Just want to double check, is this PR ready for review?

jkmassel:
> Just want to double check, is this PR ready for review?

It is now, thanks!

crazytonyli (approved):

PR #1549, opened 2026-08-10
Expose `ParsedUrl.by_appending_query_pairs` for query-aware REST URL resolution

### LINKED ISSUE #1543: Expose query-aware REST URL resolution (ParsedUrl.by_appending_query_pairs)
**Repo:** `Automattic/wordpress-rs` (Rust + UniFFI → Swift & Kotlin bindings). **Not** GutenbergKit.
**Tracking:** GutenbergKit#579. **Prior art:** wordpress-rs#1366 (the `rest_route` discovery fix, shipped in 0.6.0).

## Why

GutenbergKit#579 consolidates six hand-rolled `rest_route` URL joiners onto wprs's `WpOrgSiteApiUrlResolver`. The resolver already does the rest_route-aware **path** join and is exported. The one gap: consumers can't attach endpoint **query parameters** (`context=edit`, `status=active`, `exclude=core,gutenberg`) to a resolved URL without re-implementing the `?`→`&` merge — because `resolve()` takes no query and `ParsedUrl`'s query methods aren't exported across FFI. Closing this gap unblocks GutenbergKit adoption and deletes the Swift + Kotlin copies of the merge.

## Goal

Expose, across the UniFFI boundary (Swift + Kotlin), a rest_route-aware way to attach arbitrary query pairs to a resolved REST URL. **Additive, non-breaking.**

## Where the code is

- `wp_api/src/parsed_url.rs` — `ParsedUrl { inner: Url }`. The rest_route join `by_extending_rest_api_path` (non-exported `impl`, ~L46; note it uses `query_pairs_mut().append_pair`, which percent-encodes). Exported block `#[uniffi::export] impl ParsedUrl` (~L127: `parse` / `url` / `pretty_url`).
- `wp_api/src/request/endpoint.rs` — `ApiUrlResolver` trait (`#[uniffi::export(with_foreign)]`, ~L136: `resolve(namespace, segments)`, `route_path(namespace, path)`); `WpOrgSiteApiUrlResolver` (~L147); `resolve` delegates to `by_extending_rest_api_path`.
- `wp_api/src/url_query.rs` — existing `pub(crate)` `AppendUrlQueryPairs` trait + `QueryPairs` wrapper (the internal typed-request query mechanism). Reuse this machinery for consistency where practical.

## Recommended approach

Add an exported method on `ParsedUrl` (a general, reusable URL primitive — composes with `resolve()`):

```rust
#[uniffi::export]
impl ParsedUrl {
    /// Appends query parameters to this URL, preserving any existing query.
    /// Works uniformly on path roots and query-based (`?rest_route=`) roots:
    /// `append_pair` adds `&k=v` after the existing query (the rest_route value)
    /// or `?k=v` if there is none.
    pub fn by_appending_query_pairs(&self, pairs: Vec<QueryPair>) -> Arc<ParsedUrl> {
        let mut url = self.inner.clone();
        for p in &pairs {
            url.query_pairs_mut().append_pair(&p.name, &p.value);
        }
        Arc::new(ParsedUrl::new(url))
    }
}
```

Consumer flow (host or GutenbergKit): `resolver.resolve(ns, segments).byAppendingQueryPairs([...])`. Correct on **both** root forms because `append_pair` handles the `?`/`&` bookkeeping and encoding — the same mechanism `by_extending_rest_api_path` already uses, so behavior stays consistent.

Keep `resolve()`, `route_path`, and the `ApiUrlResolver` trait **unchanged** (the trait is `with_foreign`; adding a required method breaks Swift/Kotlin implementors).

### Decisions to make (flagged, with evidence)

- **Param type across FFI.** No `#[uniffi::export]` signature in the repo takes `Vec<(String, String)>` — tuples appear only in internal code — and the convention is `#[derive(uniffi::Record)]`. So introduce a small record:
  ```rust
  #[derive(Debug, Clone, uniffi::Record)]
  pub struct QueryPair { pub name: String, pub value: String }
  ```
  First check `url_query.rs` for an existing Record-friendly pair type to reuse. If the pinned UniFFI version supports exported tuples and the team prefers them, `Vec<(String, String)>` is acceptable — confirm before choosing.
- **Optional resolver convenience.** Optionally add `resolve_with_query(namespace, segments, pairs)` so callers resolve-and-attach in one call. Not required; the `ParsedUrl` method is the must-have. If added, keep it additive (new method, not a changed signature).

## Behavior spec — golden cases (must hold)

Appending `[{context, edit}, {status, active}]` to a resolved base:

| Base (from `resolve`) | Result |
|---|---|
| `https://example.com/wp-json/wp/v2/themes` (path root) | `…/wp/v2/themes?context=edit&status=active` |
| `https://example.com/index.php?rest_route=%2Fwp%2Fv2%2Fthemes` (query root) | `…?rest_route=%2Fwp%2Fv2%2Fthemes&context=edit&status=active` |
| query root already carrying `&debug=1` | new pairs appended after; `&debug=1` preserved |

Plus:
- **Empty `pairs`** → URL returned unchanged.
- **Reserved chars in a value**, e.g. `exclude=core,gutenberg` → serialized `exclude=core%2Cgutenberg` (form-urlencoded; WordPress decodes it). Assert the encoded form **deliberately** — it's byte-different from GutenbergKit's current literal-comma output but functionally equivalent, and it matches `by_extending_rest_api_path`'s existing encoding.
- **Order preserved; duplicate keys allowed** (append, do not dedupe).

## Tests

- Add an rstest table in `parsed_url.rs`'s test module, mirroring the existing `#[case::rest_route_query_form(...)]` style, covering every golden case above.
- If the repo has a Swift/Kotlin binding smoke-test harness, add one call through the generated bindings to prove the FFI export + the `QueryPair` record round-trip.

## Bindings / build verification

- This must cross the UniFFI boundary. Regenerate and verify the Swift **and** Kotlin bindings expose `by_appending_query_pairs` and `QueryPair`.
- Per the repo's own CHANGELOG note, `make xcframework-only-macos` is the fast way to verify a UniFFI change (not the full 11-target `make xcframework`). Confirm the Kotlin bindings regenerate cleanly too.

## CHANGELOG

Add an `### Added` entry under `## [Unreleased]` in `CHANGELOG.md`, e.g.:

> REST URL resolution can now attach endpoint query parameters via `ParsedUrl.by_appending_query_pairs` (Swift/Kotlin), so consumers building `?rest_route=` URLs no longer re-implement the `?`→`&` merge. Complements `WpOrgSiteApiUrlResolver.resolve`.

## Out of scope

- No GutenbergKit or host-app changes; no `EditorConfiguration` API change (that's #579 downstream).
- Don't change `resolve()` / `route_path` signatures or the `ApiUrlResolver` trait shape.
- Don't touch preload-key formatting (a GutenbergKit concern).
- Cutting a wprs release is a separate step after merge.

## Acceptance criteria

- [ ] `ParsedUrl.by_appending_query_pairs` (+ `QueryPair` if used) exported and callable from Swift and Kotlin.
- [ ] Correct on path roots and `?rest_route=` roots (trailing slash and none), preserves existing query params, order-stable, empty-safe, encoding matches `by_extending_rest_api_path`.
- [ ] rstest coverage for all golden cases; binding smoke test if a harness exists.
- [ ] `resolve()` / `route_path` / trait unchanged; change is purely additive.
- [ ] `## [Unreleased] → Added` CHANGELOG entry.
- [ ] `make xcframework-only-macos` (or equivalent) confirms the Swift surface; Kotlin bindings regenerate cleanly.

## References

- GutenbergKit#579 (consolidation), wordpress-rs#1366 (rest_route fix, in 0.6.0)
- `parsed_url.rs`: https://github.com/Automattic/wordpress-rs/blob/835be4335b13e01094b0096d0395882f37fac24f/wp_api/src/parsed_url.rs#L46-L101
- `endpoint.rs` (resolver): https://github.com/Automattic/wordpress-rs/blob/835be4335b13e01094b0096d0395882f37fac24f/wp_api/src/request/endpoint.rs#L136-L172
- api-fetch reference behavior: https://github.com/WordPress/gutenberg/blob/816fdbb14f353498aed164cde410274d741057ec/packages/api-fetch/src/middlewares/root-url.ts#L11-L40

jkmassel:
## Description

Closes #1543. GutenbergKit#579 consolidates six hand-rolled `rest_route` URL joiners onto wprs's `WpOrgSiteApiUrlResolver`. The resolver already does the rest_route-aware **path** join, but consumers couldn't attach endpoint **query parameters** (`context=edit`, `status=active`, `exclude=core,gutenberg`) to a resolved URL without re-implementing the `?`→`&` merge — `resolve()` takes no query and `ParsedUrl`'s query methods weren't exported across the UniFFI boundary.

This adds a general, composable URL primitive: `ParsedUrl.by_appending_query_pairs`. The consumer flow becomes `resolver.resolve(ns, segments).byAppendingQueryPairs([...])`, correct on **both** API-root forms. The change is additive and non-breaking — `resolve()`, `route_path`, and the `ApiUrlResolver` trait are **unchanged** (the trait is `with_foreign`, so adding a required method would break Swift/Kotlin implementors).

## Changes

- **`wp_api/src/parsed_url.rs`**: New `QueryPair { name, value }` `uniffi::Record` and an exported `ParsedUrl.by_appending_query_pairs(pairs:)`. It appends form-urlencoded pairs while preserving any existing query, working uniformly on path roots (`…/wp/v2/themes` → `?context=edit&status=active`) and `?rest_route=` query roots (keeps the existing `rest_route` value, appends `&context=edit&status=active`). Pairs are order-stable and duplicate keys are kept. Empty `pairs` returns the URL unchanged via an early return that sidesteps the `url` crate's `query_pairs_mut()` pushing a `?` on its first call.
- **`native/swift/Sources/wordpress-api/Exports.swift`**: Re-export `QueryPair` from the public `WordPressAPI` module — the generated type lives in `WordPressAPIInternal`, so without the typealias Swift consumers can't name or construct it.
- **`wp_api/src/request/endpoint/plain_permalinks_url_tests.rs`** (new): A golden URL table that drives all 45 self-hosted endpoint URL builders through a `?rest_route=`-seeded `WpOrgSiteApiUrlResolver` and asserts the exact plain-permalinks URL each one produces — the counterpart to the per-endpoint `wp-json` goldens in each `*_endpoint.rs`. It exercises the `rest_route` join across every endpoint shape (embedded-slash paths like `plugins`/`block-renderer`, all three self-hosted namespaces, and the API root). This is complementary coverage: it locks the shipped `resolve()` path (#1366) on plain permalinks, not the new `by_appending_query_pairs` surface.
- **`CHANGELOG.md`**: `Unreleased → Added` entry, plus an `Internal` note for the endpoint golden table.

## Test plan

Encoding deliberately matches `by_extending_rest_api_path`: `exclude=core,gutenberg` serializes to `exclude=core%2Cgutenberg` (form-urlencoded) — byte-different from GutenbergKit's current literal-comma output but functionally equivalent, and WordPress decodes it.

- [x] `cargo test -p wp_api --lib` — 30+ new `rstest` cases in `parsed_url.rs` (both root forms, trailing-slash and none, `&debug=1` preservation, reserved/space/unicode/`+`/`/` encoding, empty-name and empty-value, order-stable, duplicate keys, fragment preservation, non-mutation, chainability, resolve→append composition) plus a resolver-level composition test in `endpoint.rs`.
- [x] `cargo clippy --tests --all-features -- -D warnings` clean; `cargo fmt --all -- --check` clean.
- [x] Kotlin: UniFFI bindings regenerate; `compileKotlin` + `compileIntegrationTestKotlin` + `detekt` pass. Generated `fun byAppendingQueryPairs(pairs: List<QueryPair>): ParsedUrl` and `data class QueryPair(name, value)` confirmed; smoke test in `ParsedUrlTest.kt`.
- [x] Swift: `make xcframework-only-macos` + `swift test` — 79 tests pass, including the new `ParsedUrlTests` (path root, `?rest_route=` root, `%2C` encoding, empty-unchanged) round-tripping the `QueryPair` record through the generated bindings; `swift-format --strict` clean.
- [x] `cargo test -p wp_api --lib request::endpoint::plain_permalinks_url_tests` — 45 endpoints, one golden `?rest_route=` URL each, all green. Expected strings were authored by an independent form-urlencoder (not the resolver), so the table is a genuine regression lock rather than a tautology. Every endpoint resolves correctly on plain permalinks — no surprises.
- [x] `cargo test -p wp_api --lib request::endpoint::index_self_href_url_tests` — server-sourced full-URL oracle: for the 22 endpoints WordPress advertises as non-parameterized routes, our resolver reproduces the server's own published URL (`_links.self.href`, captured in the real-site index fixture `test-data/api-details/test-case-03.json`), byte-for-byte, across all three namespaces. 22/22 green. This guards the golden table's accuracy against an independent source rather than my own encoder.
- [x] Integration (`make test-server`): `test_login_plain_permalinks_mut` — both tests pass against a live Dockerized WordPress. The REST index publishes no URL for parameterized routes, so the new `fetch_object_by_id_on_plain_permalinks_site` fetches a real object through `/wp/v2/users/<id>` over `?rest_route=` on a plain-permalink site and asserts the id round-trips; the existing `/users/me` check was refactored onto a shared helper and still passes.

## Related issues

- Closes #1543
- Unblocks GutenbergKit#579
- Builds on #1366 (the `rest_route` discovery fix, shipped in 0.6.0); the plain-permalinks endpoint golden table extends that fix's coverage to the whole endpoint surface

## Changelog

- [x] I've added an entry to `CHANGELOG.md` under `## [Unreleased]`, using the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) categories.


wp_api/src/request/endpoint.rs
```diff
@@ -245,4 +250,38 @@ mod tests {
...
+    /// on both API-root forms.
+    #[rstest]
+    #[case::pretty_permalinks(
+        "https://example.com/wp-json",
+        "https://example.com/wp-json/wp/v2/themes?context=edit&exclude=core%2Cgutenberg"
+    )]
+    #[case::plain_permalinks_rest_route_form(
+        "https://example.com/index.php?rest_route=/",
+        "https://example.com/index.php?rest_route=%2Fwp%2Fv2%2Fthemes&context=edit&exclude=core%2Cgutenberg"
+    )]
```
dcalhoun, line 265: Should we add a third case here for the bare `?rest_route=` root (no trailing slash)?

Trailing-slash normalization is already covered on `by_extending_rest_api_path` (`parsed_url.rs:287`), and this method doesn't touch the route value. So, the untested bit is the interaction through the composed path, which I believe is what GBK will call:

```rust
#[case::plain_permalinks_rest_route_no_trailing_slash(
    "https://example.com/index.php?rest_route=",
    "https://example.com/index.php?rest_route=%2Fwp%2Fv2%2Fthemes&context=edit&exclude=core%2Cgutenberg"
)]
```

dcalhoun (commented):
This looks good to me. I believe there is value in receiving @oguzkocer's feedback on the implementation as well. 

A couple of clarifying questions from reviewing this...

**1. Sequencing work**

> Does not block fix: build REST URLs for sites using plain permalinks #573 or WordPress-iOS#25859 — both are correct, tested, user-facing fixes. This supersedes them structurally, later.

As quoted above, https://github.com/wordpress-mobile/GutenbergKit/issues/579 describes proceeding with the original, currently unapproved implementations. Do we still want to merge those as-is or instead merge/integrate this PR? I'm fine with either approach, just mitigating miscommunication. 

**2. Preloading cache keys**

It seems GBK needs to know the canonical route key for a request, separate from the request URL. This is because `EditorPreloadList` and `EditorURLCache` both key on canonical WP paths, which is a contract with api-fetch's preloading middleware. Should constructing the canonical route key be the responsibility of wprs or the consumer?

Further details from Claude's review: 

<details><summary>Claude's review</summary>

> These changes close the gap it targets, and the URLs it produces preload correctly — I checked all three root forms through api-fetch's `normalizePath` + `rest_route` unwrap, and `%2F`/`%2C` encoding and param order are all normalized away. No change needed here for that.
> 
> One question before GBK adopts it: **does wprs intend to own canonical route keys?**
> 
> https://github.com/wordpress-mobile/GutenbergKit/issues/579 has `route_path` supplying preload keys, but it takes no query. Matching is a plain dict lookup on the normalized path, and `normalizePath` early-returns when there's no query — so a query-less key can't match a query-bearing request. GBK's keys are `/wp/v2/themes?context=edit&status=active` and `/wp/v2/types/post?context=edit`; core-data sends `context=edit` for themes (`entities.js:199`), so the query is on both sides. Verified: `route_path`-shaped key → miss; full resolved URL as key → miss (normalization doesn't strip the origin). A miss isn't an error, just a silent fallback to network on exactly the sites preloading helps most.
> 
> Two workable answers:
> 
> 1. **wprs owns keys** — a `route_path` variant taking query pairs.
> 2. **Consumers own keys** — fine, but worth a doc note, since GBK's resolver protocol then needs a `routeKey` separate from `resolve` from day one.
> 
> Either way: _`route_path` for keys, `resolve` for requests._ Preference?


</details>

oguzkocer (commented):
The approach looks good to me. I didn't see a blocker during my first review, but I need to do another pass. In the meantime, here are some notes from my session with Claude:

---

## PR #1549 review — 0ed486ff

### Worth fixing

**1. The oracle skips routes the fixture already publishes — and the docs claim it doesn't.**

`index_self_href_url_tests.rs:1` says "for **every** self-hosted endpoint that WordPress advertises as a non-parameterized route" and "this covers the 22 collection/singleton endpoints." The fixture publishes `_links.self[0].href` for these `/wp/v2` collection routes that the module never asserts:

`themes`, `plugins`, `templates`, `template-parts`, `menus`, `menu-items`, `categories`, `tags`

Those are exactly the endpoints whose `plain_permalinks_url_tests.rs` entries carry hand-written paths with no server oracle behind them — the gap the oracle exists to close. Extending it is cheap; the fixture already holds the data.

Two related claims to narrow either way:

- `CHANGELOG.md:42` — "To guard the table's accuracy against an independent source, the 22 non-parameterized routes are also checked … against WordPress's own published URL." The oracle runs against a `wp-json` **path** root (`index_self_href_url_tests.rs:28`), so it validates the route *path segments* the table builds on, not the `?rest_route=` join or encoding the table asserts.
- `plain_permalinks_url_tests.rs:1` and `CHANGELOG.md:42` — "all 45 self-hosted endpoints" is one representative call per endpoint *module*; the users endpoint alone has create/update/delete builders that aren't exercised.

### Maybe worth fixing

**2. A failing assertion leaves the shared test server on plain permalinks.**

`test_login_plain_permalinks_mut.rs:46` and `:75` — both tests restore the permalink structure as their last statement, so any panic before it (a failed `assert_response()`, a failed `assert_eq!`) skips the restore. `wp_cli::set_permalink_structure` is called nowhere else in the workspace, so nothing else puts it back: the rest of that run, and the next one against the same container, sees a mis-configured server. Pre-existing for the first test, but this PR adds a second copy — a Drop guard that restores on unwind would cover both while the code is open.

**3. The 787 KB fixture is re-read and re-parsed on every assertion.**

`index_self_href_url_tests.rs:46` — `published_self_href` does a `read_to_string` + full `serde_json` parse of a 28.5k-line file per call, 22 times per run. A `LazyLock<Value>` reads it once.

### Nits

- `index_self_href_url_tests.rs:36` — `url.split('?').next()` always yields `Some`, so `.unwrap_or(url)` is unreachable. `url.split_once('?').map_or(url, |(base, _)| base)` says it directly.
- `parsed_url.rs:566` and `endpoint.rs:266` assert the same two-root composition, one through `by_extending_rest_api_path` and one through `resolve()`. The resolver one is the public flow.
- `parsed_url.rs` — of the new rstest cases, ~11 (`space_encoded_as_plus`, `ampersand`/`equals`/`plus`/`slash`/`unicode` in value, `space_in_name`, `empty_name`, `empty_value`, `order_preserved`, `duplicate_keys_kept`) pin the `url` crate's `form_urlencoded` serializer rather than this method, and `..._is_chainable` / `..._does_not_mutate_receiver` pin what `&self` returning a new value already guarantees. The empty-pairs early return, both root forms, existing-query preservation, and `%2C` are the ones carrying weight.
- `parsed_url.rs:484, 500, 533, 542, 567` use `.unwrap()`; the new `endpoint.rs` test uses `.expect("valid url")`. `.expect()` with a message everywhere gives a useful failure line.
- `plain_permalinks_url_tests.rs` — fully-qualified `crate::users::UserId`, `crate::blocks::BlockId`, … throughout; `use` statements at the top would cut most of the noise.
- `endpoint.rs:267` — function-local `use crate::parsed_url::QueryPair;` belongs in the module's use block at line 189.

Issue #1499, opened 2026-08-07
`.networkConnectionLost` is classified as `DeviceIsOfflineError` even when the server severed the connection


jkmassel:
> Split out from Automattic/wordpress-rs#1497 — *Swift executor URLSession error audit*. Section A — Converted to the wrong Rust error.
>
> File references are to `native/swift/Sources/wordpress-api/SafeRequestExecutor.swift` unless noted; `:NNN` line numbers were verified against `fix/converge-executor-error-classification`.

**Severity: high.** Apple documents -1005 as "a client or server connection was severed in the middle of an in-progress load" — it fires for server/proxy RSTs, load-balancer idle timeouts, and the notorious iOS stale keep-alive reuse case, with the device fully online. Apps show offline banners or queue for connectivity restoration when a plain retry would succeed. Swift is the only executor reporting offline for a server-side abort: reqwest maps the equivalent io `UnexpectedEof` to `HttpError { "The server terminated the connection unexpectedly" }` and never constructs `DeviceIsOfflineError`; Kotlin only claims offline after `NetworkAvailabilityProvider` confirms the network is down.

**Fix:** keep `.notConnectedToInternet` as `DeviceIsOfflineError`. For `.networkConnectionLost`, either (a) emit `HttpError` matching reqwest, or (b) — the faithful fix, since -1005 also fires on genuine mid-request connectivity drops — gate on a connectivity signal (`NWPathMonitor`, mirroring Kotlin's provider) and emit `DeviceIsOfflineError` only when the path is actually unsatisfied.

**Caveat:** `.networkConnectionLost` is listed as a Swift offline signal in the new `is_device_offline` rustdoc, so the contract text and the variant-coverage tests from 8aa3cf20 must move with the fix. The `waitsForConnectivity` caveat from Automattic/wordpress-rs#1491 doesn't help here — -1005 is reported for established connections.

PR #1557, opened 2026-08-12
fix: classify Swift .dataNotAllowed / .internationalRoamingOff / .callIsActive as DeviceIsOfflineError

### LINKED ISSUE #1501: `.dataNotAllowed`, `.internationalRoamingOff`, `.callIsActive` fall to `GenericError` instead of `DeviceIsOfflineError`
>

**Severity: medium.** All three mean the device cannot use the network right now; the site is not implicated. `.dataNotAllowed` is the high-frequency one (Wi-Fi off with the per-app cellular toggle disabled, carrier policy). `errorIsDeviceIsOffline` (`:267-273`) only checks two codes today. Adding these stays within the documented contract ("Swift derives it from the OS-reported `URLError` codes") and matches what Android callers effectively get through the `NetworkAvailabilityProvider` gate.

jkmassel:
## Description

Three `URLError` codes that mean the device can't use the network right now — cellular data disallowed for the app or by carrier policy (`.dataNotAllowed`), international roaming turned off while abroad (`.internationalRoamingOff`), and a voice call holding the radio on a single-radio device (`.callIsActive`) — had no branch in the Swift executor's `errorIsDeviceIsOffline`, so they fell through to the catch-all `.genericError`. An app checking `isDeviceOffline` (or matching `DeviceIsOfflineError`) to show an offline state therefore saw nothing for the highest-frequency case of all: Wi-Fi off with the per-app cellular toggle disabled, which surfaces as `.dataNotAllowed`. This classifies all three as `DeviceIsOfflineError`, alongside the existing `.notConnectedToInternet` / `.networkConnectionLost`.

This stays within the documented contract — `RequestExecutionErrorReason::is_device_offline` already says Swift derives offline from the OS-reported `URLError` codes — and matches what Kotlin callers effectively get through the `NetworkAvailabilityProvider` gate. Split out from the Swift executor error audit in #1497 (section B, finding 4).

Fixes #1501

## Changes

- Classify `.dataNotAllowed`, `.internationalRoamingOff`, and `.callIsActive` as `DeviceIsOfflineError` by adding them to `errorIsDeviceIsOffline` in `SafeRequestExecutor.swift`, with a comment grouping the five codes by meaning (no connection at all vs. the OS refusing network use).
- Update the offline-predicate docs to cover the new codes: the `is_device_offline` rustdoc's enumerated Swift code list, and the Swift-facing `isDeviceOffline` doc-comments in `Extensions.swift`, which described only "the device has no network connection".
- Add a parameterized regression test that drives each new code through a real `URLSession` completion path via a `URLProtocol` that fails the request with it.
- Exclude that test and its `URLProtocol`/request stubs from watchOS with `#if !os(watchOS)`.
- Add a `Fixed` entry to `CHANGELOG.md`.

**The regression test is excluded on watchOS; nothing else on that platform changes.** The test injects each `URLError` through a `URLProtocol`. watchOS's URL loading system delivers an injected *natural* failure such as `.timedOut` faithfully, but not these cellular-radio codes — so on that leg the executor observes a different reason and the `.deviceIsOfflineError` assertion trips. `errorIsDeviceIsOffline`, the classifier under test, has no platform conditionals and is exercised identically on macOS, iOS, and tvOS, so driving the stub on watchOS added fragility without coverage. The first CI run ([build #6181](https://buildkite.com/automattic/wordpress-rs/builds/6181)) failed only on `test-swift-watchOS`, once per parameterized case; the gate is test-only and changes no shipped behavior.

## Test plan

These codes come from device state a unit test can't set (cellular policy, roaming, an in-progress call), so the test injects the `URLError` through a failing `URLProtocol` rather than provoking it. The common case is reproducible by hand — disable Wi-Fi and turn the app's Cellular Data switch off, yielding `.dataNotAllowed` — after which `isDeviceOffline` returns `true` instead of `false`.

- [x] `swift test --filter SafeRequestExecutor` on macOS — 4 tests pass. The parameterized case drives `.dataNotAllowed` (`-1020`), `.internationalRoamingOff` (`-1018`), and `.callIsActive` (`-1019`) through the executor and asserts each returns `.deviceIsOfflineError`; the pre-existing `sleep(millis:)` and timeout tests stay green.
- [x] [Build #6184](https://buildkite.com/automattic/wordpress-rs/builds/6184) green end to end, including all four Swift legs and the xcframework builds. macOS/iOS/tvOS run the new parameterized test; watchOS skips it per the gate above.
- [x] `xcrun swift format lint --strict` and `cargo fmt --check` clean.

## Not in this PR

`.networkConnectionLost` (`-1005`) keeps its existing offline classification and is untouched here. Apple also fires it when the server severs an established connection (RSTs, load-balancer idle timeouts) with the device fully online, so reclassifying it is a behavior change tracked separately in #1499. This PR is purely additive, so the two don't conflict — but when #1499 lands, the rustdoc grouping and the Swift predicate order both need to move with it.

## Changelog


native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
```diff
@@ -317,9 +317,22 @@ public final class WpRequestExecutor: SafeRequestExecutor {
...
+        // holding the radio on an older single-radio device (`.callIsActive`). This
+        // matches what Kotlin callers get through the `NetworkAvailabilityProvider`
+        // gate. See #1501.
         [
             .networkConnectionLost,
-            .notConnectedToInternet
+            .notConnectedToInternet,
+            .dataNotAllowed,
+            .internationalRoamingOff,
+            .callIsActive
```
crazytonyli, line 335: Not against adding it. But I'm curious is this still a thing, in the context of 4G and 5G?

crazytonyli (approved):

native/swift/Sources/wordpress-api/SafeRequestExecutor.swift
jkmassel: My understanding is it's pretty carrier-dependent but can theoretically still happen, even on 4G?

crazytonyli: I remember that was an problem years and years ago, where you can't connect to internet while on a phone call. I don't think that's still a thing now. But, as I mentioned, there is no issue including this error case.

PR #1560, opened 2026-08-12
Add new `WpApiParamCommentsStatus` to be used in query

crazytonyli:
## Description

At the moment, the `CommentStatus` (the enum that's used to parse the comment status response) is used to query by comment status in GET requests. However, "approved" (`CommentStatus.approved) is not an acceptable status value in the GET endpoint: it's "approve".

This PR introduce a new WpApiParamCommentsStatus to be used in the GET request query.

The Android app currently workaround this issue by [using the `custom` variant](https://github.com/wordpress-mobile/WordPress-Android/blob/6a06b107e1f9f2e46551b3687e34f3c11c09a833/WordPress/src/main/java/org/wordpress/android/ui/commentsrs/CommentsRsListTab.kt#L45-L50). @nbradbury FYI, this PR introduces breaking changes: you will need to change that code (and the `CommentStatus.custom("all")") once this PR is released.

### DIFF
--- a/wp_api_integration_tests/tests/test_comments_err.rs
+++ b/wp_api_integration_tests/tests/test_comments_err.rs
@@ -324,6 +324,8 @@ async fn list_err_forbidden_param_comment_type(
 async fn list_err_forbidden_param_status(
     #[values(
         WpApiParamCommentsStatus::Hold,
+        WpApiParamCommentsStatus::All,
+        WpApiParamCommentsStatus::Any,
         WpApiParamCommentsStatus::Spam,
         WpApiParamCommentsStatus::Trash
     )]
--- a/wp_api_integration_tests/tests/test_comments_err.rs
+++ b/wp_api_integration_tests/tests/test_comments_err.rs
@@ -2,6 +2,7 @@ use wp_api::{
     comments::{
         CommentCreateParams, CommentCreateParamsBuilder, CommentDeleteParams, CommentListParams,
         CommentRetrieveParams, CommentStatus, CommentType, CommentUpdateParams,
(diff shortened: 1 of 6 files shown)


wp_api_integration_tests/tests/test_comments_err.rs
```diff
@@ -321,7 +322,12 @@ async fn list_err_forbidden_param_comment_type(
 #[rstest]
 #[parallel]
 async fn list_err_forbidden_param_status(
-    #[values(CommentStatus::Hold, CommentStatus::Spam, CommentStatus::Trash)] status: CommentStatus,
+    #[values(
```
jkmassel, line 325: Should we also test `all`/`any`?

jkmassel (approved):
Left a testing suggestion – Claude also suggested `parent_exclude` isn't tested anywhere?

wp_api_integration_tests/tests/test_comments_immut.rs
```diff
@@ -244,10 +288,12 @@ async fn parse_extras() {
 #[case::parent(generate!(CommentListParams, (parent, vec![CommentId(1), CommentId(2)])))]
 #[case::parent_exclude(generate!(CommentListParams, (parent, vec![CommentId(1), CommentId(2)])))]
```
crazytonyli: @jkmassel This PR only touches the status query. The `parent_exclude` thing you mentioned might be an issue here, where it's supposed to add a `parent_exclude` test case using the `parent_exclude` query, but the declaration says use `parent`. Probably just a copy-paste error.

PR #1561, opened 2026-08-12
Add API to read cached post

crazytonyli:
## Description

I want to use this API to read cached post in the comments screens, to show what post the comment is posted on.

I'd like to use the function signature `read_posts_by_ids_from_db -> AnyPostWithEditContext`. The function name is already taken. So I changed the existing function to `read_post_full_entities_by_ids_from_db -> FullEntity<AnyPostWithEditContext>`.

### DIFF
--- a/wp_mobile/src/service/posts.rs
+++ b/wp_mobile/src/service/posts.rs
@@ -632,7 +632,10 @@ impl PostService {
     /// Read posts by IDs from the database cache.
     ///
     /// Returns full entity data for all requested IDs that exist in the cache.
-    /// Posts not in the cache are silently omitted from the result.
+    /// Posts not in the cache are silently omitted from the result. Each
+    /// distinct ID yields at most one entity: duplicate IDs in `ids` collapse
+    /// to a single result at the first occurrence's position. The result
+    /// follows the order of `ids`.
     ///
     /// # Arguments
     /// * `ids` - Post IDs to load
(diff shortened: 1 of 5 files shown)


wp_mobile/src/service/posts.rs
```diff
@@ -699,6 +706,29 @@ impl PostService {
...
+    /// Cache-only: no network request is made, so this is safe to call for
+    /// resolving display data (e.g. post titles for a comments list) without
+    /// triggering fetches. IDs not present in the cache are silently omitted
+    /// from the result; callers that need to distinguish missing posts must
+    /// compare the result against the requested IDs.
+    ///
+    /// Results follow the order of `post_ids`. Each distinct ID yields at most
+    /// one post: duplicate IDs collapse to a single entry at the first
+    /// occurrence's position, so this behaves as a set-style batch lookup.
+    pub fn read_posts_by_ids_from_db(
```
oguzkocer: I think this should be `async` which will avoid clients accidentally running it on the main thread, at least it does in Kotlin by making it a `suspend fun`.

It looks like we broke that pattern in a few places in the repo which we'll need to address.

oguzkocer (approved):
@crazytonyli I left a comment which I believe is important to address before landing this, but I'm still going to approve it because it's a simple enough change that you can fix and land at the same time.

wp_mobile/src/service/posts.rs
crazytonyli: I have addressed this one case in https://github.com/Automattic/wordpress-rs/pull/1561/commits/ebd1df9f5346e9eee2c821d721da0498798e4e96

PR #1564, opened 2026-08-13
Add unspam and untrash comment endpoints

crazytonyli:
## Description

Mirror wp-admin's "Not Spam" and "Restore" actions with dedicated executor methods, implemented as paramless POST variants that send the REST status param's unspam/untrash values. These restore the comment's saved pre-spam/pre-trash status and fire the unspam_comment / untrash_comment hooks, which a plain hold status write would not do.

| Not Spam | Restore |
| - | - |
|  |  |

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


jkmassel (changes requested):
I think this has some unwanted side-effects. See https://github.com/Automattic/wordpress-rs/pull/1566

PR #1565, opened 2026-08-13
Give every date a type that says what it is

oguzkocer:
## Description

Every field holding a point in time is now `WpGmtDateTime` or `WpDateString` rather than a bare `String`, and every way of reading one goes through a single parser.

The two types split on a question the API answers differently per field: can this value be resolved to an absolute instant? `WpGmtDateTime` is for values that can, because they are UTC or carry an offset — the bindings lower it to a unix timestamp, so a value that isn't genuinely an instant becomes a wrong one. `WpDateString` is for values that can't: a bare calendar date, or a datetime in the *site's* timezone, which needs the site's offset to place. Deciding between them means reading what the endpoint produces rather than what the field is called, which is why the new `CONVENTIONS.md` records the rule and the reasoning behind the awkward cases.

Along the way this fixes three defects that the typing exposed:

- A datetime WordPress never set was being read as a real instant in 1 BCE. WordPress stores `0000-00-00 00:00:00` to mean a date was never written — deliberately, for a draft whose publish date should float until it's published — and an endpoint that doesn't guard the value hands it to PHP's formatter. Four of its spellings parsed cleanly and came back as 30 November of 1 BCE; the rest failed the response. One value, two behaviours, decided by nothing but which format the endpoint used.
- A cached post or media item whose stored `modified_gmt` couldn't be read was treated as up to date and never refetched, so the stale copy stayed in the cache indefinitely.
- Blocks' create/update `date` was a `WpGmtDateTime`, but that serializes without an offset and WordPress reads an offsetless value as site-local — so passing a UTC instant published at that wall clock in the site's timezone instead.

### Types

- `WpDateString` widened to cover site-local datetimes as well as bare calendar dates, and given `AsQueryValue`.
- `/wp/v2` `date` and `modified` are site-local → `WpDateString`; their `_gmt` twins were already `WpGmtDateTime`. Covers posts, pages, media, comments, blocks, navigation, templates, template parts and every revision and autosave resource.
- WordPress.com stats: response `date`, the `date`/`start_date`/`end_date` parameters, and the post rows they carry.
- `users.registered_date`, application passwords' `created` and `last_used`, and `blog_registration_date` on the individual-subscriber stats are GMT → `WpGmtDateTime`.
- `NavigationListParams`' four date windows were left as `String` while every sibling list-params struct used `WpGmtDateTime`; they now match.
- `StatsVisitsPeriod`, `StatsSubscribersPeriod` and `StatsPostViewPeriod` for the time-series `period` values, which are not dates. The shape follows the requested unit, and each endpoint builds its labels in its own code without agreeing with the others — a yearly label is `2013-01-01` on visits and `2013` on subscribers — so they get a type each rather than one shared one.
- `WpDateString` and the three period types are `uniffi::Record`s rather than newtypes erased to `String`, so Swift and Kotlin see distinct types with somewhere to hang behaviour, instead of a bare string that invites whatever date helper is nearest to hand.
- `modified_gmt` on the cache's list-metadata row and input types, removing three stringify-and-reparse round trips.

### Parsing

- One parser, `parse_wp_date_time`, backs `FromStr`, `Deserialize`, the optional-field deserializer and the bindings' timestamp lift. They previously disagreed about what a `WpGmtDateTime` is.
- It recognises WordPress's never-set date in each spelling it arrives in and reports it as distinct from a malformed value, so a field that can legitimately be unset reads it as absent while everything else still errors. Any instant before year 1 is rejected rather than accepted.
- Fields on endpoints that format their date column *without* guarding that value — `/wp/v2` users and comments, and the WordPress.com individual-subscriber stats — are optional, so one such record no longer fails an entire list response.

### Behaviour fixes

- The post and media staleness checks refetch an item whose cached timestamp can't be read, instead of treating it as current.
- Lifting a `WpGmtDateTime` across the bindings rejects a timestamp resolving to the never-set date or to any instant before year 1, and no longer turns an out-of-range value into the unix epoch silently.

### Documentation

- `CONVENTIONS.md` (new, linked from `CLAUDE.md`) records the date rule, how to choose between the two types, and which way each endpoint behaves when a date was never set. It also records the rule the period types came out of: two values share a type only when they are the same thing, and an answer that needs a qualifier is a no.
- The trade-off that comes with recognising the never-set date is written where an error message would send someone looking — on `WpDateTimeParseError::NotSet`.

## Notes for reviewers

23ed30dd retyped the four date-window parameters to `WpDateString` and 28142580 reverts it wholesale. The premise was wrong: `WP_Date_Query::build_mysql_datetime` passes the site timezone to `date_create` only as a fallback, so an offset-bearing value keeps its own offset and is converted to site-local before comparison — `WpGmtDateTime` was always correct for those. f64e962d re-lands only the part that was genuinely broken.

## Subsequent work

Date-related follow-ups deliberately left out of this PR:

- **Helpers on `WpDateString` and the period types.** They are records now, so there is somewhere to put behaviour — parsing the calendar date, applying a site's offset to place a site-local datetime — but this PR adds no methods, only the representation that makes them possible.
- **What the list-metadata read should do when a stored `modified_gmt` won't parse.** Two call sites diverge — `ListMetadataReader::get_items` swallows the error and reports the list as never cached, while the metadata collections surface it as a `CollectionError::DatabaseError`. Both `.ok().flatten()` calls predate this PR; what changed is that a new class of value can reach them, since a database written by an older build may hold a value the parser now rejects. Whether any real database holds one is worth establishing first — it may make the question moot.
- **An integration test for the never-set date.** Core guards `/wp/v2` posts, so the reachable path is a comment with a zero `comment_date_gmt` inserted via direct SQL in `scripts/setup-test-site.sh`.
- **The plugin directory's `last_updated` and `added`.** `added` is a bare calendar date; `last_updated` is `"2024-07-29 10:47am GMT"`, a GMT instant that needs a 12-hour-clock format added to the parser. This is the wordpress.org directory API, a different surface from the rest of this PR.

## Changelog


wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -1,10 +1,152 @@
...
+    NotSet,
+    /// The value matches none of the forms WordPress sends, or resolves to an
+    /// instant before year 1, which no WordPress datetime legitimately has.
+    Invalid,
+}
+
+impl Display for WpDateTimeParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::NotSet => write!(f, "Date is WordPress's never-set value"),
```
crazytonyli: > The value is WordPress's never-set date rather than a datetime.

What is "never-set value"?

wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -1,10 +1,152 @@
...
+/// it to another first lands on a different day — `Europe/Berlin` to UTC gives
+/// `-0001-11-29 23:06:32` — which no prefix here matches, so it is rejected as
+/// malformed rather than recognised as unset. That distinction only changes
+/// whether an optional field reads `None` or errors; either way the value
+/// never reaches a caller as an instant.
+const NOT_SET_SPELLINGS: [&str; 3] = ["0000-00-00", "-001-11-30", "-0001-11-30"];
+
+/// The instant PHP derives from the zero date, as a unix timestamp. Reached by
+/// any spelling of it that one of the accepted formats happens to parse.
+const NOT_SET_TIMESTAMP: i64 = -62_169_984_000;
```
crazytonyli: Is "zero date" same as "never-set date", where the string or number value is 0?

wp_serde_helper/src/wp_serde_date.rs
```diff
@@ -1,10 +1,152 @@
...
+/// Returns [`WpDateTimeParseError::NotSet`] for WordPress's never-set date,
+/// and [`WpDateTimeParseError::Invalid`] if the value matches none of the
+/// forms above.
+pub fn parse_wp_date_time(s: &str) -> Result<DateTime<Utc>, WpDateTimeParseError> {
+    if NOT_SET_SPELLINGS
+        .iter()
+        .any(|spelling| s.starts_with(spelling))
+    {
+        return Err(WpDateTimeParseError::NotSet);
+    }
```
crazytonyli, line 87: The string comparison feels a bit too strict to me.

Just an idea: I wonder if we can relax the parsing a little bit. If the `s` parse result is around (like +- one day) the UTC time of "0000-00-00 00:00:00", we'll treat it as a NotSet error. No real user-generated content would be set in that time, I hope?

wp_serde_helper/src/wp_serde_date.rs
oguzkocer: This was sort-of explained in the PR description, but the docs didn't call it out and I had two different terms for the same thing - so thank you for calling out!

I tried to improve the docs in 54cbcc00d3b867edf935ef7b3a12ff0044a5f7d7 preferring the zero-date term over never-set one. My understanding of the WordPress source implementation is that some columns will have MySQL's zero-date `0000-00-00 00:00:00` assigned to it and some endpoints try to format it without checking for it. So, we end up receiving a value that looks like a proper date, but it should be `null` instead.

I don't know how often this happens in practice. I haven't had any previous experience with this particular issue, but Claude found several references of it. I'm hoping to add an integration test for it at some point whenever I can find some time for it.

wp_serde_helper/src/wp_serde_date.rs
oguzkocer: That's right. I've unified the two terms in 54cbcc00d3b867edf935ef7b3a12ff0044a5f7d7. Hopefully it's clear now!

wp_serde_helper/src/wp_serde_date.rs
oguzkocer: That's a good catch, thank you! I used a one-day window as you suggested in e776349b01be2c8fbad4f9f3a168e9d9570999d4.

crazytonyli: Does that mean the if block above and `ZERO_DATE_SPELLINGS` can be removed?

oguzkocer: We still need these ones because the parsers can't handle them resulting in the `WpDateTimeParseError::Invalid` error instead of `WpDateTimeParseError::NotSet`. I verified these in the unit tests.

However, the docs were a bit confusing, so I tried to improve the relevant parts in 023183d196e28b30b4cb2a1988968047335bc0c9. Let me know what you think!

crazytonyli (approved):

PR #1569, opened 2026-08-17
Remove obsolete OpenSSL workaround

iangmaia:
## Description

Removes the obsolete explicit Ruby OpenSSL dependency as is no longer needed now that we use Ruby 3.4.9 (see paaHJt-aks-p2). 

## Changes

- Removed the explicit OpenSSL dependency and its lockfile entries.
- Added an internal changelog entry.

## Changelog

- [x] Added an entry to CHANGELOG.md under Unreleased > Changed.


AliSoftware (approved):

mokagio:
Nice!

PR #1578, opened 2026-08-18
Read the WordPress.com REST v1.x Error Code from `error`

oguzkocer:
## Description

`WpError` deserializes the API error code from `code`, which is what the WordPress REST API sends. WordPress.com REST v1.x sends it as `error`, so a v1.x rejection never parsed and reached callers as `WpApiError::UnknownError` with the code and message dropped, leaving nothing to branch on.

Aliasing the field covers both shapes. OAuth2 bodies pair `error` with `error_description` rather than `message`, so they still fall through to the other error paths.

The first commit, 1a308789, adds the tests and is red on its own; the second, b0bd8fc4, makes it green.

## Changes

- Added `#[serde(alias = "error")]` to `WpError.code`.
- Added unit tests over `WpError::try_parse` covering the `code` shape, the `error` shape, and an OAuth2 body that carries neither pairing.
- Added `domains::suggestions` e2e coverage: the happy path plus the `invalid_query`, `empty_results`, and `invalid_minimum_quantity` rejections.
- Updated `transactions::redeem_empty_cart` and `post_stats::get_stats_post_homepage` to match `WpErrorCode::CustomError` rather than searching the raw response body for `empty_cart` and `invalid_blog`, which is what they did while the code never reached them.

## Changelog


wp_api/src/api_error.rs
```diff
@@ -923,6 +927,41 @@ mod tests {
...
+
+    /// OAuth2 rejections carry `error` without a `message`, so they stay
+    /// outside this shape and fall through to the other error paths.
+    #[test]
+    fn test_does_not_parse_oauth2_error_body() {
+        let body = br#"{"error":"invalid_token","error_description":"The token is invalid"}"#;
+
+        assert_eq!(WpError::try_parse(body), None);
+    }
+
```
oguzkocer, line 964: We should address this as well potentially by adding an alias for `error_description`, but I guess that's going to be in a future PR as I need this PR to land for `0.7.0` and don't have the bandwidth to do the necessary research for it.

crazytonyli (approved):

PR #1580, opened 2026-08-19
Reduce Rust test debug info in CI to prevent OOM

iangmaia:
Fixes AINFRA-2938

## Description

We've noticed some builds cause the [CI agent to be lost](https://buildkite.com/automattic/wordpress-rs/builds/6251#01a017b5-6cb0-42e0-a903-3d1fe6949bbf), who kept running for long without being terminated.
  
This PR aims to prevent OOM-related Buildkite agent loss without reducing Rust test concurrency.

## Changes

- Use line-table-only debug info for Rust library and doc tests (also made the job duration to fall from **~8m** to **~5m40s**).

## Test plan

- `make -n test-rust-lib test-rust-doc`
- [Buildkite run](https://buildkite.com/automattic/wordpress-rs/builds/6260)

## Changelog

- [x] Added an entry under `## [Unreleased]`.


AliSoftware (approved):
This feels like a worth change to try to check if it will indeed reduce OOMs significantly. And to my understanding should have limited impact to the debuggability of the compiled library, so seems worth a tradeoff to make.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Internal:** Use line-table-only debug info for Rust tests to prevent OOM-induced Buildkite agent loss while retaining file/line backtraces.
```
oguzkocer: The changelog entry should go into the `Unreleased` section.

oguzkocer, line 51: The changelog entry should have gone into the `Unreleased` section.

Makefile
```diff
@@ -224,11 +224,12 @@ test-swift-tvOS: xcframework
...
 	scripts/xcodebuild-test.sh watchOS-26-5
 
+# Line tables preserve file/line backtraces without full debug info's CI memory cost.
 test-rust-lib:
-	$(rust_docker_run) cargo test --lib -- --nocapture
+	$(rust_docker_run) env CARGO_PROFILE_TEST_DEBUG=line-tables-only cargo test --lib -- --nocapture
 
 test-rust-doc:
-	$(rust_docker_run) cargo test --doc -- --nocapture
+	$(rust_docker_run) env CARGO_PROFILE_TEST_DEBUG=line-tables-only cargo test --doc -- --nocapture
```
oguzkocer, line 232: I don't think this covers the entire CI run, so we probably needed it to be part of root `Cargo.toml` file:
```
 [profile.test]
 debug = "line-tables-only"
```

I was looking into the implications of that seeing that since we were also pinged for review, but the PR got merged already. 😕

CHANGELOG.md
oguzkocer: Fixed in https://github.com/Automattic/wordpress-rs/pull/1581.

Makefile
oguzkocer: Updated in https://github.com/Automattic/wordpress-rs/pull/1581. I'm waiting for the CI result and will request a review afterwards.

PR #1581, opened 2026-08-19
Move the Rust test debug info setting into `[profile.test]`

oguzkocer:
## Description

#1580 capped Rust test debug info by prefixing two `Makefile` targets with `env CARGO_PROFILE_TEST_DEBUG=line-tables-only`. This reverts that commit and reapplies the same setting as `[profile.test]` in the workspace `Cargo.toml`.

The environment variable reached `test-rust-lib` and `test-rust-doc` only. The four other `cargo test` invocations in the `Makefile` — one of them behind `scripts/run-rust-integration-tests.sh` — and any `cargo test` run outside `make` still built with full debug info, and every test target added later would have needed the prefix repeated. A profile entry covers all of them and cannot be forgotten. It also keeps local runs and CI on one setting, so alternating between `make test-rust-lib` and the `cargo test --lib` that `CLAUDE.md` documents no longer rebuilds the dependency graph each way.

The profile applies to every crate a `cargo test` run builds, dependencies included, rather than only the crate under test — confirmed with `cargo test -p wp_serde_helper --lib --no-run -v`, where the dependencies compile with `-C debuginfo=line-tables-only`.

Shipped artifacts are untouched: the Apple libraries build with `--profile $(CARGO_PROFILE)` (`release` on CI) under `--config profile.$(CARGO_PROFILE).debug=true`, and the Android JNI libraries with `cargo ndk ... build --release` under `RUSTFLAGS=-g`. Crash-report symbolication is unaffected on both platforms.

## Changes

- Reverted a68ebebf, dropping the `env CARGO_PROFILE_TEST_DEBUG=line-tables-only` prefix from `test-rust-lib` and `test-rust-doc`, and removing its changelog entry — which had landed inside the released `## [0.7.0]` section rather than `## [Unreleased]`.
- Added `[profile.test] debug = "line-tables-only"` to the workspace `Cargo.toml`, commented with the tradeoff (backtraces still resolve to file and line; a debugger can no longer print locals) and the `CARGO_PROFILE_TEST_DEBUG=full` override for a debugging session.
- Added the changelog entry under `## [Unreleased]`, filed as `Changed` to match the other `**Internal:**` build and CI entries.

## Changelog


Cargo.toml
```diff
@@ -24,6 +24,18 @@ members = [
...
+# memory a `cargo test` link consumes on CI. Backtraces still resolve to file
+# and line; a debugger can no longer print locals or format values. For a
+# debugging session, override per invocation:
+#
+#   CARGO_PROFILE_TEST_DEBUG=full cargo test
+#
+# This profile covers every crate a `cargo test` run builds, dependencies
+# included — not just the crate under test.
+[profile.test]
+debug = "line-tables-only"
```
iangmaia, line 37: 👍

iangmaia (approved):

iangmaia:
@oguzkocer Thanks for the changelog fix and the config improvement and apologies for rushing the previous PR -- I was too focused on getting the test tasks with the debug info setting and running it on CI.
I appreciate the `[profile.test]` also affecting other test as well as local cargo runs and potential future test targets 👍

PR #1585, opened 2026-08-20
Validate changelog entries under Unreleased

iangmaia:
## Description

Improves changelog validation so missing updates fail while misplaced entries receive an advisory warning.

## Changes

- Require `CHANGELOG.md` to be updated and retain `## [Unreleased]`.
- Warn when code changes do not change non-whitespace content under `## [Unreleased]`.
- Allow dedicated changelog corrections and release-style updates.

## Test plan

- [Advisory CI run](https://buildkite.com/automattic/wordpress-rs/builds/6321#01a03f91-6996-4cad-8210-007d316f42d5): moving this PR's entry under released version `0.7.0` posted a PR warning and a persistent Buildkite annotation without failing the step.
- [Clean changelog step](https://buildkite.com/automattic/wordpress-rs/builds/6322#01a03f92-96d4-4406-8bf5-32a0d0fd524b): restoring the entry under `## [Unreleased]` passed and removed the transient PR warning.
- Tested missing files/headings, historical-only edits, comment failures, large changelogs, and whitespace-only edits in synthetic PR repositories.

## Changelog

- [x] I've added an entry to `CHANGELOG.md` under `## [Unreleased]`, using a Keep a Changelog category.

### DIFF
--- a/.buildkite/commands/validate-changelog.sh
+++ b/.buildkite/commands/validate-changelog.sh
@@ -2,6 +2,30 @@
 
 set -euo pipefail
 
+# The top-level "- " entries under "## [Unreleased]", read from stdin and
+# sorted so comm can diff two of these sets.
+unreleased_entries() {
+  awk '
+    $0 == "## [Unreleased]" { in_unreleased = 1; next }
+    in_unreleased && /^## / { exit }
+    in_unreleased && /^- / { print }
+  ' | LC_ALL=C sort -u
+}
+
+# Pass, deleting any failure comment from a prior run so a green run leaves no
+# residue. The delete no-ops if the comment does not exist.
+pass() {
+  printf '%s\n' "$1"
+  comment_on_pr --id changelog-check --if-exist delete "" || true
+  exit 0
+}
+
+fail() {
+  printf '\n%s\n\n' "$1" >&2
+  comment_on_pr --id changelog-check "$1"
+  exit 1
+}
+
 # Only enforce on PR builds — the check is about *new* entries being added
 # alongside the change that ships them.
 if [[ "${BUILDKITE_PULL_REQUEST:-false}" == "false" ]]; then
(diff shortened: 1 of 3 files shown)


oguzkocer (commented):
I haven't fully reviewed the changes yet, but wouldn't this prevent us from fixing entries. Arguably they won't end up in the wrong place from now on, but there might still be legitimate reasons for modifying the CHANGELOG in a way that's not expected by this PR, right? For example, I had [a recent PR where I had to add `BREAKING` to an existing entry.](https://github.com/Automattic/wordpress-rs/pull/1579)

iangmaia:
> I haven't fully reviewed the changes yet, but wouldn't this prevent us from fixing entries. Arguably they won't end up in the wrong place from now on, but there might still be legitimate reasons for modifying the CHANGELOG in a way that's not expected by this PR, right? For example, I had [a recent PR where I had to add `BREAKING` to an existing entry.](https://github.com/Automattic/wordpress-rs/pull/1579)

Good point, the count-based check is probably too strict 🤔 
What if we instead just (at the cost of making the check slightly more complex, but eventually we could migrate it to Ruby / Danger to hopefully simplify it):
- Allow PRs that only change `CHANGELOG.md`.
- For other PRs, require something under `## [Unreleased]` to change (no count check).

That would still catch entries added to an old release while allowing corrections like #1579. Does that sound reasonable?

oguzkocer:
@iangmaia That sounds good to me. Another option might be to post a warning comment on the PR when we detect that the change didn't go under `Unreleased` section and let the developers be responsible for making the right call. Even when an entry goes into the wrong place, it's fairly easy to fix, so I don't think we need to be too strict about it.

iangmaia:
@oguzkocer 
> @iangmaia That sounds good to me. Another option might be to post a warning comment on the PR when we detect that the change didn't go under `Unreleased` section and let the developers be responsible for making the right call. Even when an entry goes into the wrong place, it's fairly easy to fix, so I don't think we need to be too strict about it.

Thanks! I went with the warning approach, but still failing when `CHANGELOG.md` wasn’t updated, or when the file or `Unreleased` heading is missing. If a code PR only changes an older release section, we now leave a non-blocking warning instead. It seems like a reasonable balance?
Anyway, if you think this is ok for now, we can keep it as is and iterate later moving this check to Danger where we can more easily add other deterministic validations like milestone checks and similar.

oguzkocer (approved):
Looks good to me. :shipit: 

Claude had a few suggestions if you're interested, but good to merge as is.

---

_Reviewed at `663ce5ab` (branch `iangmaia/validate-changelog-section`, base `trunk`)._

## 1. The template checkbox now describes CI's tolerance instead of the convention

`.github/PULL_REQUEST_TEMPLATE.md:15` went from "I've added an entry to `CHANGELOG.md` under `## [Unreleased]`" to "I've updated `CHANGELOG.md`, normally under `## [Unreleased]`". The template is where contributors learn the convention, and "normally" makes the checkbox honestly checkable by someone who added their entry under `## [0.7.0]` — the exact misplacement the rest of this PR exists to catch. The script's leniency is there for the rare correction PR; it doesn't need to be restated as guidance to every author.

The script's own failure message still says the strict version ("Every PR should add an entry under the `## [Unreleased]` section", `validate-changelog.sh:74`), so the two now disagree about what's expected. I'd keep the template prescriptive and let the warning handle the exceptions.

## 2. The advisory result leaves no signal in Buildkite

`warn()` (`validate-changelog.sh:33-38`) exits 0, so the step is green, and its only durable output is the PR comment — which is `|| true`'d. If `comment_on_pr` fails (missing token on a fork PR, GitHub API blip), the warning disappears entirely: green step, message buried on stderr in a passing job's log. `fail()` has the same `|| true`, but exit 1 keeps its signal; only the advisory path degrades to nothing.

`buildkite-agent annotate --style warning --context changelog-check` alongside the comment would keep it visible in the build regardless. Nothing else in `.buildkite/` uses annotations, so it's a new pattern here — but this is the one path in the script where a failed comment means no signal at all.

## 3. The changelog-only exemption no longer buys anything

`validate-changelog.sh:96-99` exempts PRs whose only changed file is `CHANGELOG.md`, for "a dedicated changelog correction". Both cases it's meant to cover already pass without it:

- A release PR (`Documentation/RELEASING.md:7` — "only edits `CHANGELOG.md`") moves the Unreleased entries into the new version section, so `CURRENT_UNRELEASED` differs from `BASE_UNRELEASED` and it passes at line 108 anyway.
- A genuine correction to a released section would fall through to `warn()`, which is non-blocking and whose text already answers itself: "If the changelog update is intentionally correcting an older release, no action is needed."

The exemption is the only remaining path where a misplaced entry gets no signal at all. Dropping it removes a branch and costs one benign comment on correction PRs — worth it if the plan is to port this to Danger later.

## 4. The `unreleased_section` comment undersells what it does

`validate-changelog.sh:5` says "Prints the non-whitespace contents of every Unreleased section", which reads as "drops blank lines". The body also collapses runs of whitespace inside each line and trims the ends (lines 11-15) — that's the reason re-indenting an entry doesn't count as a change. Worth saying, since a reader deciding whether the comparison is trustworthy needs to know indentation differences are invisible to it.

